PR #102, opened 2017-04-20
Document all the properties and methods of MSALPublicClientApplication

RPangrle:


MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
 
 @interface MSALPublicClientApplication : NSObject
 
+/*! Whether or not to validate the authority used with the application, default YES */
```
partnerinflight: Might want to expand this. To a developer fresh from the field, "authority" concept will be unclear, as is unclear what is meant by "validate."

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
 @property BOOL validateAuthority;
+
+/*! The authority used with the application object */
 @property (readonly) NSURL *authority;
```
partnerinflight: great opportunity to briefly explain what an authority is. :) We don't want to basically copy Vittorio's book into this file, but something brief may be helpful.

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+
+/*! The client ID of the application */
 @property (readonly) NSString *clientId;
```
partnerinflight: again, putting on my end-user-developer hat, would love it if this documentation included a pointer of where I'm supposed to get it from, so I don't need to go search through docs to find that.

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+
+/*! The redirect URI of the application */
 @property (readonly) NSURL *redirectUri;
 
 /*! Used in logging callbacks to identify what component in the application
     called MSAL. */
 @property NSString *component;
```
partnerinflight: may consider specifying that this is entirely for dev convenience and has no impact on msal itself.

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
 
+/*!
+    Initialize a MSALPublicClientApplication with a given clientID for the default
+    authority
```
partnerinflight: again, most devs will have no idea what "default authority" means. Either we take the concept of authority out of this function definition altogether, or we briefly summarize the meaning.

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+    @param  authority   The authority to use for the application
+    @param  error       The error that occurred creating the application object, if any, if you're
+                        not interested in the specific error pass in nil.
+ */
 - (id)initWithClientId:(NSString *)clientId
              authority:(NSString *)authority
                  error:(NSError * __autoreleasing *)error;
 
+/*!
+    Returna an array of users visible to this application
```
partnerinflight: nit: spelling

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+    @param  error   The error that occured trying to the user, if any, if you're
+                    not interested in the specific error pass in nil.
+ */
 - (MSALUser *)userForIdentifier:(NSString *)identifier
                           error:(NSError * __autoreleasing *)error;
 
 #pragma SafariViewController Support
 
+/*!
+    Returns whether or not the URL is a response to a MSAL web authentication
```
partnerinflight: might provide a bit more context. in other words,something like "should be called from the appdelegate to determine whether the passed in url...." etc

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+    session. Call this before passing the URL into -handleMSALResponse:
+ */
 + (BOOL)isMSALResponse:(NSURL *)response;
+
+/*!
+    Pass a URL response from your application delegate's openURL handler into
+    MSAL for web authnetication sessions.
+ */
 + (void)handleMSALResponse:(NSURL *)response;
```
partnerinflight: so i'm curious about this paradigm. for AT/ATS calls, we don't use the "is-then" paradigm -- we just call ATS. Why then have it here?

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+
+/*!
+    Cancel a currently running interactive web authentication session.
+ */
 + (void)cancelCurrentWebAuthSession;
```
partnerinflight: need a bit more detail here. what should happen to the WebView when you call this? What about a passed-in WebView?

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
 #pragma mark -
 #pragma mark acquireToken using Login Hint
 
+
+/*!
+    Acquire a token for a new user using interactive authentication
+ 
+    @param  scopes          Scopes to request from the server, the scopes that come back
+                            can differ from the ones in the original call
+    @param  loginHint       A loginHint (usually an email) to pass to the service at the
```
partnerinflight: may consider mentioning scenarios where it's needed

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
+/*!
+    Acquire a token for a new user using interactive authentication
+ 
+    @param  scopes                  Scopes to request from the server, the scopes that come back
+                                    can differ from the ones in the original call
+    @param  additionalScopes        Scopes for the service to ask the user to consent to, will
+                                    not be included in the token returned by the service, but
+                                    if the user consented to those scopes you will be able to
+                                    silently acquire tokens with those scopes later without requiring
+                                    further user interaction
```
partnerinflight: again, from a newbie perspective, it'd be useful to explain scenarios where you'd want to pass a scope in additionalScopes vs. just passing it in the scopes array.

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -90,16 +196,60 @@
...
+                            interactive authentication flow will be locked down to.
+    @param  completionBlock The completion block that will be called when the authentication
+                            flow completes, or encounters an error.
+ */
 - (void)acquireTokenForScopes:(NSArray<NSString *> *)scopes
                          user:(MSALUser *)user
               completionBlock:(MSALCompletionBlock)completionBlock;
 
+/*!
+    Acquire a token requiring interaction for a previously used user
```
partnerinflight: nit "previously used" is a bit weird -- consider clarifying

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: I'll need to consult the wider team to figure out how we're explaining scopes and resources in our libraries as a whole. Hopefully there will be a good doc that I can point people to.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: Would `specified` be better? I wanted to make explicit that you have to have acquired a token for the user with the library prior to using this method, to differentiate it from the versions of AT that don't take a user.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: There is no passed in webview in MSAL. There is only SFSafariViewController. If the app developer calls this we dismiss the SafariViewController (if it was up) and cancel whatever authentication request was in process.

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
 
 @interface MSALPublicClientApplication : NSObject
 
+/*! Whether or not to validate the authority used with the application, default YES */
 @property BOOL validateAuthority;
+
+/*! The authority used with the application object */
 @property (readonly) NSURL *authority;
+
+/*! The client ID of the application */
```
jmprieur: Do we want to add that the application must be registered in the application registration portal (https://identity.microsoft.com) and registered applications have a ClientID. ?

MSAL/src/public/MSALPublicClientApplication.h
```diff
@@ -34,50 +34,156 @@
...
 @property (readonly) NSString *clientId;
+
+/*! The redirect URI of the application */
```
jmprieur: Do we ant to add that this is the URI where the application will be called back by the authentication authority once the user will have signed-in?

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: That's covered by the quick start documentation and the initializer.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: That's covered by the quick start documentation and the error that comes back if the developer hasn't registered it.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: It's never needed. The only case where it'd be used is if you previously collected the user's email in some other piece of UI.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: I would be more inclined to remove that part, as the whole point of this initializer is so that the developer doesn't have to worry or think about the authority.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: The app developer already knows that they need to call AT/ATS by the time they get to them, there is no ambiguity. These methods are for developers to call in in their app delegate's openURL: method. Because that method gets all of the URLs being sent to the application there needs to be some disambiguation. We could combine the two and have handleMSALResponse return NO if it was not a MSAL response, if you prefer. These two methods just happen to match what we did in ADAL.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: There's further mention of that when you initialize it, this is a read-only property.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: That's more appropriate in the initializer documentation when the developer actually has to provide a authority, less so on the read only property.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: Added a bit more information to this.

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: @vibronet You have a good one-or-two sentence explanation you'd like to see for both authority and authority validation?

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: It's a read only property

MSAL/src/public/MSALPublicClientApplication.h
RPangrle: Filed an issue to collapse the two: https://github.com/AzureAD/microsoft-authentication-library-for-objc/issues/122

Issue #77, opened 2017-04-12
Split saveAccessAndRefreshToken into two methods


RPangrle:
saveAccessToken and saveRefreshToken are really two separate things that should be in two separate methods. Parsing token responses shouldn't really be in MSALTokenCacheAccessor

Issue #76, opened 2017-04-12
Remove MSALRequestParameters from MSALTokenCacheAccessor methods


RPangrle:
MSALTokenCacheAccessor is not a request object, so it's not really who MSALRequestParameters was intended for. MSALTokenCacheAccessor should take in the specific parameters it cares about (authority, user, client_id, scopes)

PR #117, opened 2017-04-25
Refactor MSALTokenCache

unpluggedk:
- Remove MSALRequestParameter from MSALTokenCache methods
- Split saveAccessAndRefreshToken into 2 methods

Addresses #76 and #77

### DIFF
--- a/MSAL/src/cache/MSALTokenCache.m
+++ b/MSAL/src/cache/MSALTokenCache.m
@@ -56,18 +56,25 @@
     return _dataSource;
 }
 
-- (MSALAccessTokenCacheItem *)saveAccessAndRefreshToken:(MSALRequestParameters *)requestParam
-                                               response:(MSALTokenResponse *)response
-                                                context:(nullable id<MSALRequestContext>)ctx
-                                                  error:(NSError * __autoreleasing *)error
+- (MSALAccessTokenCacheItem *)saveAccessTokenWithAuthority:(NSURL *)authority
+                                                  clientId:(NSString *)clientId
+                                                  response:(MSALTokenResponse *)response
+                                                   context:(id<MSALRequestContext>)context
+                                                     error:(NSError * __autoreleasing *)error
 {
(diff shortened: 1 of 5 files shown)


MSAL/src/cache/MSALTokenCache.m
```diff
@@ -158,11 +178,11 @@ - (MSALAccessTokenCacheItem *)findAccessToken:(MSALRequestParameters *)requestPa
     
     for (MSALAccessTokenCacheItem *tokenItem in allAccessTokens)
     {
-        if (requestParam.unvalidatedAuthority && [key matches:[tokenItem tokenCacheKey:nil]])
+        if (authority && [key matches:[tokenItem tokenCacheKey:nil]])
```
RPangrle: This line worries me. There's more to a token cache key then authority...

jasoncoolmax:
I have changed the class names:
`MSALTokenCacheAccessor`     -> `MSALTokenCache`
`MSALTokenCacheDataSource` -> `MSALTokenCacheAccessor`,
which caused confusion at that time.

@RPangrle Do we still need to address those issues (#76 and #77)?  As we are already aligned with .Net and Android for TokenCache APIs:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/74b53690c328eb6611146a0a5dfb798b340a10ab/src/Microsoft.Identity.Client/TokenCache.cs#L104 

If we want to make the changes, I am also OK :)

RPangrle:
Those issues I filed were also to try to cleanup the caching logic. Our two goals are to 1) write clean code and 2) match the object model is the C# code as much as is reasonable. If you find places where #2 seems in conflict with #1 then it's a good topic for conversation and filing issues. :)

MSAL/src/cache/MSALTokenCache.m
unpluggedk: I've moved things around, no more need for MSALAccessTokenCacheKey and [key matches:]

Things checked:
- scopes: checked as a first condition to match
- clientId: filtered when allAccessTokensForUser: is called
- authority: checked after scopes.
- user: filtered when allAccessTokensForUser: is called

PR #140, opened 2017-04-30
Log result or error at the end of every acquireToken call

### LINKED ISSUE #139: Log error/result information at the end of every acquireToken request
Having a definitive result statement at the of an acquireToken call makes it a lot easier to determine which errors are intermediate and which ones were final.

RPangrle:
Wrap the passed in completionBlock so we can easily inspect and log the result of the acquireToken flow before returning to the calling app.

Fixes #139

### DIFF
--- a/MSAL/src/MSALPublicClientApplication.m
+++ b/MSAL/src/MSALPublicClientApplication.m
@@ -368,17 +368,17 @@
 {
     if (error)
     {
-        LOG_ERROR(ctx, @"%@ returning with error: %@, %ld", operation, error.domain, (long)error.code);
-        LOG_ERROR_PII(ctx, @"%@ returning with error: %@, %ld", operation, error.domain, (long)error.code);
+        NSString *errorDescription = error.userInfo[MSALErrorDescriptionKey];
+        errorDescription = errorDescription ? errorDescription : @"";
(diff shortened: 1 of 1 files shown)


MSAL/src/MSALPublicClientApplication.m
```diff
@@ -361,6 +361,27 @@ - (void)acquireTokenSilentForScopes:(NSArray<NSString *> *)scopes
 #pragma mark -
 #pragma mark - private methods
 
++ (void)logOperation:(NSString *)operation
+              result:(MSALResult *)result
+               error:(NSError *)error
+             context:(id<MSALRequestContext>)ctx
+{
+    if (error)
```
partnerinflight: any PII considerations doing this?

MSAL/src/MSALPublicClientApplication.m
```diff
@@ -426,13 +453,13 @@ - (void)acquireTokenForScopes:(NSArray<NSString *> *)scopes
     if (!request)
     {
         [params.urlSession invalidateAndCancel];
-        completionBlock(nil, error);
+        block(nil, error);
```
partnerinflight: How are Android and .NET approaching this? Are they doing the same?

PR #142, opened 2017-04-30
Add and use a short hex string utility for logging hashed PII

### LINKED ISSUE #141: Use short hex strings in logging
The longer hex blobs are a bit harder to read in logs, and occasionally make people think we're logging actual ATs in logs (we're not, they're hashes) a 8 or 16 character hex string is less likely to be confused for the genuine article, while making logs a bit more readable.

RPangrle:
Fixes #141

### DIFF
--- a/MSAL/src/util/NSString+MSALHelperMethods.m
+++ b/MSAL/src/util/NSString+MSALHelperMethods.m
@@ -341,6 +341,21 @@ static inline void Encode3bytesTo4bytes(char* output, int b0, int b1, int b2)
     return toReturn;
 }
 
+- (NSString *)msalShortSHA256Hex
+{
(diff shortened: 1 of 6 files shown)


MSAL/src/util/NSString+MSALHelperMethods.m
```diff
@@ -341,6 +341,21 @@ - (NSString *)msalComputeSHA256Hex
     return toReturn;
 }
 
+- (NSString *)msalShortSHA256Hex
+{
+    NSData *hashData = [MSALCryptoHelper msalSHA256fromString:self];
```
partnerinflight: can the result of this be nil?

RPangrle: If it is then we end up returning a string of length 0.

Issue #157, opened 2017-05-03
Query Parameters from OIDC Authorization Endpoint need to be passed along


RPangrle:
We're currently squashing QPs that come along after OIDC discovery, which is causing B2C tenants to not work.

partnerinflight:
done

PR #170, opened 2017-05-04
Retain QPs from OIDC Discovery

RPangrle:
URLs from OIDC discovery on a tenant can include query parameters on them that we must retain, or else we break B2C.


partnerinflight:
fixes #157

PR #179, opened 2017-05-05
Readme update

brandwe:
* Updated with agreed template from LT
* Added Security reporting 
* Kept how to use the library for now, although the LT was concerned that makes READMEs too large
* We'll most likely refactor that in to a Wiki page later.
* Will update with QuickStart walkthrough once available in the portal

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,10 @@ The MSAL library preview for iOS gives your app the ability to begin using the [
 
 Note that for the preview, **only iOS is supported.** macOS support will be provided later. 
 
+## Important Note about the MSAL Preview
+
+These libraries are suitable to use in a production environment. We provide the same production level support for these libraries as we do our current production libraries. During the preview we reserve the right to make changes to the API, cache format, and other mechanisms of this library without notice which you will be required to take along with bug fixes or feature improvments. This may impact your applicaiton. For instance, a change to the cache format may impact your users, such as requiring them to sign in again and an API change may require you to update your code. When we provide our General Availability release later, we will require you to update your application to our General Availabilty version within six months to continue to get support. 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,10 +234,32 @@ Use the client ID from yout app listing when initializing your MSALPublicClientA
...
+    application.acquireTokenSilent(forScopes: kScopes, user: user) { (result, error) in
+    
+        if !(error != nil)  {
+
+          accessToken = result.accessToken!
+                        
+                        
+         } else  {
+                        
+          // Interactive auth will be required
```
partnerinflight: that's actually wrong, right? We need to specifically look for interaction_required error rather than a generic error case, otherwise we'll fall back to AT for net failures.

README.md
```diff
@@ -101,10 +234,32 @@ Use the client ID from yout app listing when initializing your MSALPublicClientA
     }
 ```
 
+### Silently Acquiring an Updated Token (Swift)
+```swift
+
+    application.acquireTokenSilent(forScopes: kScopes, user: user) { (result, error) in
+    
+        if !(error != nil)  {
```
partnerinflight: why not just "if (error == nil)" ?

README.md
```diff
@@ -70,7 +184,26 @@ Use the client ID from yout app listing when initializing your MSALPublicClientA
     }
 ```
 
-### Silently Acquiring an Updated Token
+### Acquiring Your First Token (Swift)
+
+```swift
+    application.acquireToken(forScopes: kScopes) { (result, error) in
+                
+        if !(error != nil) {
```
partnerinflight: what's with the double-negative? :)

brandwe: Fixed.

README.md
brandwe: fixed

README.md
brandwe: Fixed.

PR #187, opened 2017-05-05
Remove Static Lib target and fix xcconfigs

RPangrle:
- Remove quote marks around “@rpath” in DYLIB_INSTALL_NAME_BASE, this was causing issues when dropping a framework built from our the project into another project, including when using Carthage
- Remove the static library build
- Add a clean unit test host app, as changes in init order are now exposing test failures (and maybe were before that we just didn't see...)


MSAL/test/unit/ios/unit-test-host/AppDelegate.h
```diff
@@ -0,0 +1,36 @@
+//------------------------------------------------------------------------------
```
partnerinflight: so what does this have to do with the state goal of the review? Confused.

unpluggedk:
Can we fix the CI issue?
```
xcodebuild test  -workspace MSAL.xcworkspace -scheme "MSAL (iOS Framework)" -configuration Debug -enableCodeCoverage YES -sdk iphonesimulator CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' | xcpretty
xcodebuild: error: Failed to build workspace MSAL with scheme MSAL (iOS Framework).
	Reason: Could not determine bundle identifier for MSAL iOS Unit Tests's TEST_HOST: "/Users/travis/Library/Developer/Xcode/DerivedData/MSAL-dnckqpgjtujhvvatrivpxdcjabbv/Build/Intermediates/CodeCoverage/Products/Debug-iphonesimulator/unit-test-host.app"
iOS Framework [test] Failed
```

unpluggedk:
Woops wrong button pressed ;

MSAL/test/unit/ios/unit-test-host/AppDelegate.h
RPangrle: "Add a clean unit test host app, as changes in init order are now exposing test failures (and maybe were before that we just didn't see...)"

RPangrle: Logger in the test and automation app was causing exceptions to be thrown, it's just before when the code was being pulled in as a static lib it ran fast enough that we usually ran all of the tests and exited out before hitting the exception case, but with framework init time it slowed down just enough to hit the exception and fail.

PR #196, opened 2017-05-09
Remove default slice parameters

### LINKED ISSUE #195: Remove hardcoding of test slice in slice parameters

RPangrle:
Fixes #195


MSAL/src/MSALPublicClientApplication+Internal.h
```diff
@@ -27,8 +27,8 @@
 
 #import "MSALPublicClientApplication.h"
 
-// TODO: Remove these defaults once uid/client info is in prod
-#define DEFAULT_SLICE_PARAMS @"slice" : @"testslice", @"uid" : @"true"
+// If modifying this make sure UT_SLICE_PARAMS_QUERY matches
```
partnerinflight: Hm do we still need the defaultSliceParameters function then? Maybe get rid of all of that? Or are we expecting devs to be modifying this? If so, in what scenario, and what documentation exists for it?

MSAL/test/unit/utils/MSALTestConstants.h
```diff
@@ -39,4 +39,4 @@
 #define UNIT_TEST_DEFAULT_REDIRECT_URI      UNIT_TEST_DEFAULT_REDIRECT_SCHEME"://auth"
 
 #define UT_SLICE_PARAMS_DICT DEFAULT_SLICE_PARAMS
```
partnerinflight: so this now becomes nothing. which means tests like MSALInteractiveRequestTests:testAuthorizationUri's sliceParameters dictionary becomes empty. If so, why do we have it at all? Should we just remove that? Seems an unnecessary layer of confusion at this point?

MSAL/src/MSALPublicClientApplication+Internal.h
RPangrle: I'm hesitant to remove it, because the code was added in the first place because sometimes it's easier to just hard code MSAL to a specific test slice then modify the code using MSAL. It wouldn't surprise me if it sees some utility in our testing.

PR #224, opened 2017-09-28
Update to Xcode 9 and add recommended compiler settings and warnings.

duncan:
This small update is the result of loading the project into Xcode 9, taking the suggested compiler settings and warnings, and then fixing the resulting issues in `MSAL/test/app/ios/MSALTestAppSettingsViewController.m`


duncan:
So, Xcode 9 changes where the `Coverage.profdata` file is stored to `$OBJROOT/../[DEVICE_GUID]/Coverage.profdata` -- looking for where to pick up that file from the `build.py` script.

RPangrle:
We manage build settings are managed in xcconfig files, not in the project file, as that makes it harder to keep all the targets in sync. build.py since has been updated to pick up the proper coverage.profdata.

PR #253, opened 2017-12-28
Use MSID test utilities

antrix1989:


MSAL/test/unit/utils/MSIDTestURLSession+MSAL.m
```diff
@@ -25,62 +25,21 @@
...
+#pragma mark - DataTask creation
 
-@end
-
-@implementation MSALTestSentinel
-
-static MSALTestSentinel *s_sentinel = nil;
-
-+ (void)initialize
+- (NSURLSessionDataTask *)dataTaskWithRequest:(NSURLRequest *)request
```
RPangrle: I'm kinda surprised this isn't available in the bigger MSIDTestURLSession, maybe just move it there?

antrix1989: In MSIDTestURLSession we have similar method, but instead of block we use delegate. I suggest to replace/add it to common MSIDTestURLSession in separate PR.

PR #262, opened 2018-02-20
Remove TeamID from project settings

charlymr:
## PROBLEM:
- When adding this project as submodule, we need to recursively update (to update submodule in submodule)
- When adding this project via Submodule in a CI environment, if you build for `Device` the build failed as I don't have your certificate on the the machine.

## SUGGESTED SOLUTION IN THIS PULL REQUEST:
- Added instruction to readme:  `git submodule update --init --recursive`
- Framework are signed only when releasing. We can safely remove your TeamID and tell Xcode not to - sign the Frameworks.


jasoncoolmax:
Sorry for the late response and thank you very much for contributing to the project :)

PR #263, opened 2018-02-20
Adding README and CHANGELOG to Xcode workspace

charlymr:
## PROBLEM:
- It is not really a problem but a  convenient way to get access to the ReadME and Changelog.

## SUGGESTED SOLUTION IN THIS PULL REQUEST:

- Added the 2 files at the Top of the Xcode workspace
- Added the magic .xcodesamplecode.plist to the Xcode workspace (telling Xcode to render MD files)


jasoncoolmax:
Hi @charlymr, that looks very nice! 
But I could not find any official document from Apple about adding .xcodesamplecode.plist. If that is the case, probably we'd better not to add that changes as the side effect is unknown for now, and mostly importantly, for the future :)

charlymr:
@jasoncoolmax Sure thing. I understand and you are right. Let's close this pull request then.
Thanks, Denis

PR #357, opened 2018-09-21
Add stack protector flag

antrix1989:


antrix1989:
Didn't find any documentation where mentioned that it is enabled by default, I assume it is not, so lets keep it explicitly.

PR #386, opened 2018-10-26
Public constants class added with typedefs and fixes framework includes

unpluggedk:
Public constants class added with typedefs and fixes framework includes. 
The problem was with having MSAL.h with framework headers was being included in all files.

Also includes #385


MSAL/MSAL.xcodeproj/project.pbxproj
```diff
@@ -2671,6 +2667,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D65A6FE71E40002800C69FBA /* msal__framework__ios.xcconfig */;
 			buildSettings = {
+				CLANG_ENABLE_MODULES = NO;
```
antrix1989: should we move it out to xcconfig?

MSAL/MSAL.xcodeproj/project.pbxproj
```diff
@@ -2684,6 +2681,7 @@
...
```
antrix1989: same as above

PR #387, opened 2018-10-27
Revise MSALAccountId

unpluggedk:
remove duplicate initializers and add equality

### DIFF
--- a/MSAL/src/MSALAccountId.m
+++ b/MSAL/src/MSALAccountId.m
@@ -66,6 +66,11 @@
         return YES;
     }
     
+    if (![object isKindOfClass:MSALAccountId.class])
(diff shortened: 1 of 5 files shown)


MSAL/src/MSALAccountId.m
```diff
@@ -75,4 +59,24 @@ - (NSUInteger)hash
     return hash;
 }
 
+- (BOOL)isEqual:(id)object
+{
+    if (self == object)
+    {
+        return YES;
+    }
+
```
antrix1989: I think we also need to check object type here:
```
if (![object isKindOfClass:MSALAccountId.class])
{
    return NO;
}
```

PR #479, opened 2019-03-04
Broker url scheme backup plan

jasoncoolmax:


jasoncoolmax:
Close the PR as it turns out we don't have to go for the backup plan :)

PR #499, opened 2019-03-27
Rename uiBehavior to promptType

antrix1989:


antrix1989:
@Om83 yes, it is partially related to that PR.

PR #541, opened 2019-04-29
Initial commit for osx cache.

rohitnarula7176:
*Work in progress.


d-col:
Couple of comments (I know it's a work-in-progress...) 

It seems like MSIDAccountCacheItem should set isShared to YES by default. Are accounts not always shared? If it’s the default of NO, it looks like we’ll write accounts to the keychain with the wrong (non-shared) account attribute.

It seems less error-prone to have a single primaryAttributesForKey:key method which looks at key.isShared and sets kSecAttrAccount accordingly, rather than two almost-identical methods and having all the callers need to if-check the isShared flag.

PR #572, opened 2019-05-24
update common msalobjc submodule

d-col:
Add last modification info for macOS cache (#445)
Use a dispatch queue on macOS for keychain API calls (#459)


d-col:
No longer needed, changes included by another update

Issue #923, opened 2020-05-05
Unable To Clear Accounts: "Multiple accounts found in cache."


AnthonyUccello:
Hello

I have an app where 2 users logged in (on subsequent app launches). Now when I invoke `application.getCurrentAccount` I get the error:

**Couldn't query current account with error: Err domain:MSALErrorDomain code:-50000 userInfo:{ MSALErrorDescriptionKey = "Multiple accounts found in cache."; MSALInternalErrorCodeKey = "-42102"; }**

I figured the issue then was to clear the cache however [there is no api on the cache config](https://azuread.github.io/microsoft-authentication-library-for-objc/Classes/MSALCacheConfig.html). My next idea was to get all accounts and sign out the users one by one. However, when I get the accounts object, it crashes when I try to inspect the array or access it in any fashion. 

```
    func signOutAllAccounts()
    {
                guard let applicationContext = self.applicationContext else { return }
        
                let msalParameters = MSALAccountEnumerationParameters()
                msalParameters.completionBlockQueue = DispatchQueue.main
                    
                // TODO implement for multi
                applicationContext.accountsFromDevice(for: msalParameters)
                    { (accounts, error) in
        
                    if let error = error {
                        print("Couldn't query current account with error: \(error)")
                        return
                    }
                    
                    var acts:[MSALAccount] = accounts ?? [] 
                     // acts has 2 objects in it
                    var first = acts.first // crash on access
                }
    }
```

I tried to iterate the accounts but it says it fails to conform

<img width="966" alt="Screen Shot 2020-05-05 at 1 16 13 PM" src="https://user-images.githubusercontent.com/3809669/81095185-aaea9c80-8ed2-11ea-928e-edf16e443502.png">

Even just printing the returned object fails:

<img width="997" alt="Screen Shot 2020-05-05 at 1 27 40 PM" src="https://user-images.githubusercontent.com/3809669/81096363-3dd80680-8ed4-11ea-926f-f480a98c8ceb.png">

Can you advise what else I can do to either log out the accounts or clear the cache? I am unable to getCurrentAccount anymore so login always has to be done manually. (Also note I have cleared my safari cache and that didnt work either)

rohitnarula7176:
[...]

PR #1142, opened 2020-10-26
Update README.md for multiple accounts in the cache scenario

NerevarineRule:
## Proposed changes

Added sample ObjC & Swift codes and provided description for querying a specific account when multiple accounts are present in the cache, since customers don't have sufficient information to handle such a scenario in the user guide. 

Related customer issue:
https://github.com/AzureAD/microsoft-authentication-library-for-objc/issues/923

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,7 +408,7 @@ let parameters = MSALAccountEnumerationParameters(identifier:accountIdentifier)
 
 var scopeArr = ["https://graph.microsoft.com/.default"]
 
-if #available(iOS 13.0, *)
+if #available(iOS 13.0, macOS 10.15, *)
 {
 	 application.accountsFromDeviceForParameters(with: parameters, completionBlock:{(accounts, error) in
          if let error = error 
@@ -451,7 +451,7 @@ if #available(iOS 13.0, *)
 
     NSArray<NSString *> *scopeArr = [[NSArray alloc] initWithObjects: @"https://graph.microsoft.com/.default",nil]; //define scope
 
-    if (@available(iOS 13.0, *)) //Currently, this public API requires iOS version 13 or greater.
+    if (@available(iOS 13.0, macOS 10.15, *)) //Currently, this public API requires iOS version 13 or greater.
     {
         [application accountsFromDeviceForParameters:parameters
                                      completionBlock:^(NSArray<MSALAccount *> * _Nullable accounts, __unused NSError * _Nullable error)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -379,6 +379,112 @@ parameters.completionBlockQueue = dispatch_get_main_queue();
...
+                                           	redirectUri:redirectUri
+                                            	authority:authority)
+guard let application = MSALPublicClientApplication(configuration: config) else { return }
+
+let accountIdentifier = "9f4880d8-80ba-4c40-97bc-f7a23c703084.f645ad92-e38d-4d1a-b510-d1b09a74a8ca"
+let parameters = MSALAccountEnumerationParameters(identifier:accountIdentifier)
+
+var scopeArr = ["https://graph.microsoft.com/.default"]
+
+if #available(iOS 13.0, *)
```
kaisong1990, line 411: Is this API only available for iOS?

README.md
```diff
@@ -379,6 +379,112 @@ parameters.completionBlockQueue = dispatch_get_main_queue();
...
+{
+	 application.accountsFromDeviceForParameters(with: parameters, completionBlock:{(accounts, error) in
+         if let error = error 
+         {
+            //Handle error
+         }
+         
+         guard let accountObjs = accounts else {return}
+         
+         let tokenParameters = MSALSilentTokenParameters(scopes:scopeArr, account: accountObjs[0]);
```
kaisong1990, line 421: Are we telling users to use the first account always? Or leave for them to handle multi-account case?

kaisong1990 (approved):
Approved with questions

README.md
NerevarineRule: This is just a sample code :) Just to illustrate that once you pass in the desired account identifier as the search criterion, it will return an array containing accounts that the user can then process, even if there are multiple accounts stored in the cache with different account identifiers. 
I am just picking the first account with the user-provided identifier as an example for this particular scenario; the user can choose any account in the array as they see fit :)

README.md
NerevarineRule: Good catch - this API is also available on MacOS 10.15. I overlooked it since the customer that initially reported this issue had an iOS set up. Thank you so much for catching it! :)

PR #1153, opened 2020-11-11
Cross cloud support

jasoncoolmax:
## Proposed changes

The main change is to make sure we only compare `MSALAccount` based on home account id regardless of authority. Other logic to fill up the tenant profile is already in place.

Common core changes: https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/892

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/MSALAccount.m
+++ b/MSAL/src/MSALAccount.m
@@ -161,7 +161,7 @@
 - (NSUInteger)hash
 {
     NSUInteger hash = 0;
-    // Equality of MSALAccount is depending on either one of homeAccountId or username
+    // Equality of MSALAccount is depending on equality of homeAccountId or username
     // So we are not able to calculate a precise hash
     return hash;
 }
--- a/MSAL/src/MSALAccount.m
+++ b/MSAL/src/MSALAccount.m
@@ -161,7 +161,8 @@
 - (NSUInteger)hash
 {
     NSUInteger hash = 0;
-    hash = hash * 31 + self.environment.hash;
+    // Equality of MSALAccount is depending on either one of homeAccountId or username
+    // So we are not able to calculate a precise hash
     return hash;
 }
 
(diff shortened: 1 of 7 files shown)


MSAL/IdentityCore
```diff
@@ -1 +1 @@
-Subproject commit 525cfb35eb7ea82ca6b9e1cfc1af2bf403bbf39f
```
kaisong1990, line 1: This pull request does not update CHANGELOG.md.

kaisong1990 (approved):

MSAL/src/MSALAccount.m
```diff
@@ -161,7 +161,8 @@ - (BOOL)isEqual:(id)object
 - (NSUInteger)hash
 {
     NSUInteger hash = 0;
-    hash = hash * 31 + self.environment.hash;
+    // Equality of MSALAccount is depending on equality of homeAccountId or username
+    // So we are not able to calculate a precise hash
     return hash;
 }
```
hieunguyenmsft, line 167: Should we remove the hash function here? Fix the hash value to 0 seems incorrect in general.

jasoncoolmax: `hash` is required if we want to define our own `isEqual` as indicated [here](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418795-isequal).
And it also requires [here](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418859-hash):
> If two objects are equal (as determined by the isEqual: method), they must have the same hash value.

So I think we may need to keep it anyway even though it doesn't help with performance :P

hieunguyenmsft: Thank for the info, I see we implement the own logic in `isEqual` so we don't need the hash function other than for the procedure :D

PR #1154, opened 2020-11-11
Add username check and return nil when username is nil or empty

kaisong1990:
## Proposed changes

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/MSALPublicClientApplication.m
+++ b/MSAL/src/MSALPublicClientApplication.m
@@ -440,6 +440,11 @@
                               error:(NSError * __autoreleasing *)error
 {
     MSID_LOG_WITH_CTX_PII(MSIDLogLevelInfo, nil, @"Querying MSAL account for username %@", MSID_PII_LOG_EMAIL(username));
+    if ([NSString msidIsStringNilOrBlank:username])
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelWarning, nil, @"username is nil or empty which is unexpected");
+        return nil;
+    }
     
     MSALAccountsProvider *request = [[MSALAccountsProvider alloc] initWithTokenCache:self.tokenCache
                                                                 accountMetadataCache:self.accountMetadataCache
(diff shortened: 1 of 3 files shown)


MSAL/src/MSALPublicClientApplication.m
```diff
@@ -440,6 +440,11 @@ - (MSALAccount *)accountForUsername:(NSString *)username
                               error:(NSError * __autoreleasing *)error
```
kaisong1990, line 440: This pull request does not update CHANGELOG.md.

kaisong1990: Updated

MSAL/src/MSALPublicClientApplication.m
```diff
@@ -440,6 +440,11 @@ - (MSALAccount *)accountForUsername:(NSString *)username
...
 {
     MSID_LOG_WITH_CTX_PII(MSIDLogLevelInfo, nil, @"Querying MSAL account for username %@", MSID_PII_LOG_EMAIL(username));
+    if ([NSString msidIsStringNilOrBlank:username])
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelWarning, nil, @"username is nil or empty which is unexpected");
```
hieunguyenmsft, line 445: Should we throw an error so the client able to handle it? And the log level should it be error level? Just my opinion.

kaisong1990: I was little bit struggle here. If putting an error lvl log, I assume we will throw an error. However, we will need to modify the public API which might not be necessary. That's why I only logged a warning instead of error

hieunguyenmsft (approved):

MSAL/src/MSALPublicClientApplication.m
kaisong1990: Please ignore above comments, I have updated with the error approach and Utest

PR #1373, opened 2021-09-08
Update py scripts to python3.

antrix1989:
## Proposed changes

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/device_guids.py
+++ b/device_guids.py
@@ -24,7 +24,7 @@ def get_guid_i(device) :
 	version_regex = re.compile("([0-9]+)\\.([0-9]+)(?:\\.([0-9]+))?")
 	
 	command = "instruments -s devices"
-	print "##[group]Devices"
+	print("##[group]Devices")
 	p = subprocess.Popen(command, stdout = subprocess.PIPE, stderr = subprocess.PIPE, shell = True)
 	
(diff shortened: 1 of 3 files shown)


device_guids.py
```diff
@@ -35,11 +35,13 @@ def get_guid_i(device) :
 	latest_os_version = None
 	
 	for line in p.stdout :
-		sys.stdout.write(line)
-		if (dev_name_regex.match(line) == None) :
+		sys.stdout.buffer.write(line)
```
NerevarineRule, line 38: you need  line.decode('utf-8')  here.  since strLine comes after this line. instead of creating strLine, also better to just do line.decode('utf-8') on other instances. that should fix the script and let it pass CI/CD check.

NerevarineRule (changes requested):
please address the comments. I think that should get the script to pass the pipeline.

device_guids.py
antrix1989: We don't need to convert bytes to string here since we can write bytes directly to stdout.

oldalton (approved):

NerevarineRule (approved):

PR #1879, opened 2023-09-21
Native Auth base classes that interact with MSAL

borgesmb:
## Proposed changes

This PR shows the SDK that the CIAM team had developed. It is based on the Native Auth merge into MSAL plan here: https://microsofteur.sharepoint.com/:w:/t/DevExDublin/EdfT2AmlFZFIsTtr-EaoSBIBPqcOlRo1MJ1USbEsfjfEcQ

To comply with the design changes made by our API team, we have created a more dynamic SDK. This means that we make different API requests for each authentication flow. There is also some logic involved in each of the responses we get from each request.

We have created a state machine for each flow (check this [Figma](https://www.figma.com/file/2I4zEwwLREDKQOluQ8tgQs/Native-Auth-State-Models?type=design&node-id=0%3A1&t=VUyK4gIoyUChTXWF-1) to see all the possible states), and we are making it external to developers by using delegates instead of completion blocks, as @Serhii Demchenko suggested. In this way we can guide the developer through the authentication process offering a discovering interface. We think that this approach will improve the overall developer experience.

**This PR is focused in the Native Auth base classes that interact with MSAL Objective-C, and their unit tests for:**
**- cache**
**- configuration**
**- controllers**
**- validation**
**- telemetry**
**- logger**
 **It is not compilable, so tests are not going to run. To build and test this code, please use the `ciam-master-snapshot` branch.**

The IdentityCore changes can be found in this PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/1294

**For context, this is the Native Auth technical overview:** https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?path=/%5BiOS%5D%20Native%20authentication/technical_overview.md&version=GBdanilo/native-authentication&_a=preview

State Machine:
https://www.figma.com/file/2I4zEwwLREDKQOluQ8tgQs/Native-Auth-State-Models?type=design&node-id=0-1&mode=design&t=80VUxa1RvxMn4rnl-0

## Native Auth general high level overview

**MSALNativeAuthPublicClientApplication.swift**
The public interface is the entry point of the SDK. Here is where all the supported authentication flows begin.
Regarding the sign-up flow, developers can call either `signUp(...)` or `signUpUsingPassword(...)`.
Notice that in those methods they must pass a class that conforms to `SignUpStartDelegate`. or `SignUpPasswordStartDelegate`, respectively.

**MSALNativeAuthSignUpController.swift**
The controller handles most of the logic. Each of the methods inside the MARK: - Internal can be called from the public interfaces, that are:
    - MSALNativeAuthPublicClientApplication
    - One of the States (located in the file SignUpDelegates.swift). These States are returned through the delegates, as we will see.

For each of these methods, the controller:
    - Creates a request (`MSIDHttpRequest`) using its requestProvider (`MSALNativeAuthSignUpRequestProviding`).
    - Executes the request.
    - Passes the result to its responseValidator (`MSALNativeAuthSignUpResponseValidating`)
    - Handles the validated result and returns it to the public interface, which uses the delegate to communicate back to the developer.
    - Creates and handle the local telemetry events.

The controller decides what to do with the ValidatedResponse given by the ResponseValidator.

Every validated response usually has:
**A success:** it means that the SDK moves on to the next state, which usually involves making another request or returning back to the user via the delegate.
Redirect: when the backend detects an error and wants the SDK to fallback to the WebView-based flow.
**A known error:** in some cases, these errors put the user in a recoverable state (for example, an error that requires the user to provide more attributes). In other cases, the error means the end of the flow.
**An unexpected error:** these errors are likely bugs, they happen when we there are missing attributes from the response, etc.
(Please keep in mind that not every ValidatedResponse falls into the same categories, this is just an approximation)

**MSALNativeAuthRequestConfigurator.swift**
Each of the requestProviders from the Controllers uses the RequestConfigurator to create and configure the requests. Although there are different RequestProviders (one per flow), there's only one RequestConfigurator.

The RequestConfigurator injects a custom ErrorHandler (`MSALNativeAuthResponseErrorHandler`) to every request. This is needed in order to decode the errors from the API.
These errors follow a structure that you can see in `MSALNativeAuth(SignUp/SignIn/PasswordReset)StartResponseError` for example. In the directory native_auth/network/errors/ you can see all the files used.

**MSALNativeAuth(SignUp/SignIn/PasswordReset)ResponseValidator.swift**
The ResponseValidator is in charge of analysing the api response and returning to the Controller a validated response (for instance, for the response of the (SignUp/SignIn/PasswordReset)/start endpoint, it returns a `MSALNativeAuth(SignUp/SignIn/PasswordReset)StartValidatedResponse`).

**SignUp/SignIn/PasswordReset)States.swift**
In this file there are the different States that the Controller can return to the developer through the delegate.

This is how the developer will use them. For more examples of how developers can use the SDK you can check the Sample App (located in /Samples/ios-native-auth-simple in the branch `ciam-master-snapshot`). We will include it in a separate PR:

extension EmailAndPasswordViewController: SignUpPasswordStartDelegate {

    func onSignUpCodeRequired(newState: MSAL.SignUpCodeRequiredState,
                                                      sentTo _: String,
                                                      channelTargetType _: MSAL.MSALNativeAuthChannelType,
                                                      codeLength _: Int) {

           // show UI to get the user's email OOB code, and wait for user's input
           // then, use the newState to send it to the SDK.
           newState.submitCode(code: code, delegate: self)
    }
}

Each of these states then calls one of the Controller's internal functions, and it follows the same process that we just described in the previous steps.

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
+++ b/MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
@@ -1,0 +1,181 @@
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+@_implementationOnly import MSAL_Private
+
+class MSALNativeAuthCacheAccessor: MSALNativeAuthCacheInterface {
+
+    private let tokenCacheAccessor: MSIDDefaultTokenCacheAccessor = {
+        let dataSource = MSIDKeychainTokenCache()
+        return MSIDDefaultTokenCacheAccessor(dataSource: dataSource, otherCacheAccessors: [])
+    }()
+
+    private let accountMetadataCache: MSIDAccountMetadataCacheAccessor = MSIDAccountMetadataCacheAccessor(dataSource: MSIDKeychainTokenCache())
+    private let externalAccountProvider: MSALExternalAccountHandler = MSALExternalAccountHandler()
+    private let validator = MSIDTokenResponseValidator()
+
+    func getTokens(
+        account: MSALAccount,
+        configuration: MSIDConfiguration,
+        context: MSIDRequestContext) throws -> MSALNativeAuthTokens {
+            let accountConfiguration = try getAccountConfiguration(configuration: configuration, account: account)
+            let idToken = try tokenCacheAccessor.getIDToken(
+                forAccount: account.lookupAccountIdentifier,
+                configuration: accountConfiguration,
+                idTokenType: MSIDCredentialType.MSIDIDTokenType,
+                context: context)
+            let refreshToken = try tokenCacheAccessor.getRefreshToken(
+                withAccount: account.lookupAccountIdentifier,
+                familyId: nil,
+                configuration: accountConfiguration,
+                context: context)
+            let accessToken = try tokenCacheAccessor.getAccessToken(
+                forAccount: account.lookupAccountIdentifier,
+                configuration: accountConfiguration,
+                context: context)
+            return MSALNativeAuthTokens(accessToken: accessToken, refreshToken: refreshToken, rawIdToken: idToken.rawIdToken)
+        }
+
+    func getAccount(
+        accountIdentifier: MSIDAccountIdentifier,
+        authority: MSIDAuthority,
+        context: MSIDRequestContext) throws -> MSIDAccount? {
+            return try tokenCacheAccessor.getAccountFor(
+                accountIdentifier,
+                authority: authority,
+                realmHint: nil,
+                context: context)
+        }
+
+    func getAllAccounts(configuration: MSIDConfiguration) throws -> [MSALAccount] {
+        let request = MSALAccountsProvider(tokenCache: tokenCacheAccessor,
+                                           accountMetadataCache: accountMetadataCache,
+                                           clientId: configuration.clientId,
+                                           externalAccountProvider: externalAccountProvider)
+        return try request?.allAccounts() ?? []
+    }
+
+    func validateAndSaveTokensAndAccount(
+        tokenResponse: MSIDTokenResponse,
+        configuration: MSIDConfiguration,
+        context: MSIDRequestContext) throws -> MSIDTokenResult? {
+            let ciamOauth2Provider = getCIAMOauth2Provider(clientId: configuration.clientId)
+            return try? validator.validateAndSave(tokenResponse,
+                                                  oauthFactory: ciamOauth2Provider.msidOauth2Factory,
+                                                  tokenCache: tokenCacheAccessor,
+                                                  accountMetadataCache: accountMetadataCache,
+                                                  requestParameters: getRequestParameters(tokenResponse: tokenResponse,
+                                                                                          configuration: configuration,
+                                                                                          context: context),
+                                                  saveSSOStateOnly: false)
+        }
+
+    // Here we create the MSIDRequestParameters required by the validateAndSave method
+    private func getRequestParameters(
+        tokenResponse: MSIDTokenResponse,
+        configuration: MSIDConfiguration,
+        context: MSIDRequestContext
+    ) -> MSIDRequestParameters {
+
+        // We are creating the default MSIDRequestParameters to prevent unintended functionality changes.
+        // If the method `validateAndSaveTokenResponse` from `MSIDTokenResponseValidator` changes
+        // the implementation here also needs to change to match the properties needed by the method
+        // Currently only the required and used parameters are set
+        let parameters = MSIDRequestParameters()
+        // MSIDRequestParameters has to follow MSIDRequestContext protocol
+        parameters.correlationId = context.correlationId()
+        parameters.logComponent = context.logComponent()
+        parameters.telemetryRequestId = context.telemetryRequestId()
+        parameters.appRequestMetadata = context.appRequestMetadata()
+
+        parameters.msidConfiguration = configuration
+        parameters.clientId = configuration.clientId
+
+        let displayableId = tokenResponse.idTokenObj?.username()
+        let homeAccountId = tokenResponse.idTokenObj?.userId
+
+        let  accountIdentifier = MSIDAccountIdentifier(displayableId: displayableId, homeAccountId: homeAccountId)
+        parameters.accountIdentifier = accountIdentifier
+        parameters.authority = configuration.authority
+        parameters.instanceAware = false
+        let defaultOIDCScopesArray = MSALPublicClientApplication.defaultOIDCScopes().array as? [String]
+        parameters.oidcScope = defaultOIDCScopesArray?.joinScopes()
+        return parameters
+    }
+
+    func removeTokens(
+        accountIdentifier: MSIDAccountIdentifier,
+        authority: MSIDAuthority,
+        clientId: String,
+        context: MSIDRequestContext) throws {
+            try tokenCacheAccessor.clearCache(
+                forAccount: accountIdentifier,
+                authority: authority,
+                clientId: clientId,
+                familyId: nil,
+                clearAccounts: false,
+                context: context)
+        }
+
+    func clearCache(
+        accountIdentifier: MSIDAccountIdentifier,
+        authority: MSIDAuthority,
+        clientId: String,
+        context: MSIDRequestContext) throws {
+            try tokenCacheAccessor.clearCache(
+                forAccount: accountIdentifier,
+                authority: authority,
+                clientId: clientId,
+                familyId: nil,
+                clearAccounts: true,
+                context: context)
+        }
+
+    private func getCIAMOauth2Provider(clientId: String) -> MSALCIAMOauth2Provider {
+        return MSALCIAMOauth2Provider(clientId: clientId,
+                               tokenCache: tokenCacheAccessor,
+                               accountMetadataCache: accountMetadataCache)
+
+    }
+
+    private func getAccountConfiguration(configuration: MSIDConfiguration,
+                                         account: MSALAccount) throws -> MSIDConfiguration? {
+        // When retrieving tokens from the cache, we first have to get the
+        // Tenant Id from the AccountMetadataCache. Because in NativeAuth
+        // We use only CIAM authorities, we retrieve using its provider
+        let ciamOauth2Provider = getCIAMOauth2Provider(clientId: configuration.clientId)
+        let accountConfiguration = configuration.copy() as? MSIDConfiguration
+        let errorPointer: NSErrorPointer = nil
+        let requestAuthority = ciamOauth2Provider.issuerAuthority(with: account,
+                                                                      request: configuration.authority,
+                                                                      instanceAware: false,
+                                                                      error: errorPointer)
+        if let errorPointer = errorPointer, let error = errorPointer.pointee {
+            throw error
+        }
+        accountConfiguration?.authority = requestAuthority
+        return accountConfiguration
+    }
+}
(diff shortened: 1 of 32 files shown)


MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
```diff
@@ -0,0 +1,181 @@
...
+                configuration: accountConfiguration,
+                context: context)
+            let accessToken = try tokenCacheAccessor.getAccessToken(
+                forAccount: account.lookupAccountIdentifier,
+                configuration: accountConfiguration,
+                context: context)
+            return MSALNativeAuthTokens(accessToken: accessToken, refreshToken: refreshToken, rawIdToken: idToken.rawIdToken)
+        }
+
+    func getAccount(
```
antrix1989, line 61: I noticed that MSID* and MSAL* classes are mixed in this class:
- getAccount method returns MSIDAccount
- getAllAccounts returns [MSALAccount]

Should we separate layers and don't mix them? MSAL classes are just public wrappers around MSID classes, so I think usually for your private logic you should use MSID layer only.

MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
```diff
@@ -0,0 +1,181 @@
...
+                                                  tokenCache: tokenCacheAccessor,
+                                                  accountMetadataCache: accountMetadataCache,
+                                                  requestParameters: getRequestParameters(tokenResponse: tokenResponse,
+                                                                                          configuration: configuration,
+                                                                                          context: context),
+                                                  saveSSOStateOnly: false)
+        }
+
+    // Here we create the MSIDRequestParameters required by the validateAndSave method
+    private func getRequestParameters(
```
antrix1989, line 96: why do you need this method? MSIDRequestParameters should be created in the top layer api (when dev calls our public api) and should contain all params provided from the dev. MSIDConfiguration is also part of MSIDRequestParameters.

MSAL/src/native_auth/cache/MSALNativeAuthTokens.swift
```diff
@@ -0,0 +1,37 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@_implementationOnly import MSAL_Private
+
+class MSALNativeAuthTokens {
```
antrix1989, line 27: will we return this class in public api or it is for private logic only? If it is only for private, why don't we use MSIDTokenResult?

MSAL/src/native_auth/configuration/MSALNativeAuthConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@_implementationOnly import MSAL_Private
+
+struct MSALNativeAuthConfiguration {
```
antrix1989, line 27: I see a mix of MSID* and MSAL* classes here as well.

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
```diff
@@ -0,0 +1,537 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@_implementationOnly import MSAL_Private
+
+// swiftlint:disable file_length
```
antrix1989, line 27: swiftlint is open sourced lib, should we include it's license?

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
```diff
@@ -0,0 +1,537 @@
...
+        }
+
+        MSALLogger.log(
+            level: .info,
+            context: context,
+            format: "resetpassword: waiting for \(pollInterval) seconds before retrying"
+        )
+
+        do {
+            try await Task.sleep(nanoseconds: 1_000_000_000 * UInt64(pollInterval))
```
antrix1989, line 516: do we always wait on BG thread, or this can be invoked on the main thread too?

MSAL/test/unit/native_auth/cache/MSALNativeAuthCacheAccessorTest.swift
```diff
@@ -0,0 +1,329 @@
...
+        XCTAssertNoThrow(try cacheAccessor.validateAndSaveTokensAndAccount(tokenResponse: tokenResponse, configuration: parameters.msidConfiguration, context: contextStub))
+
+        XCTAssertNoThrow(tokens = try cacheAccessor.getTokens(account: parameters.account, configuration: parameters.msidConfiguration, context: contextStub))
+        XCTAssertEqual(tokens?.accessToken?.accessToken, newAccessToken)
+        XCTAssertEqual(tokens?.refreshToken?.refreshToken, newRefreshToken)
+        XCTAssertEqual(tokens?.rawIdToken, newIdToken)
+    }
+
+    func testGetAllAccounts_whenAllInfoPresent_shouldRetrieveDataCorrectly() {
+        clearCache()
```
antrix1989, line 77: should clearCache be called in setUp?

MSAL/test/unit/native_auth/cache/MSALNativeAuthCacheAccessorTest.swift
```diff
@@ -0,0 +1,329 @@
...
+        XCTAssertNoThrow(account = try cacheAccessor.getAllAccounts(configuration: parameters.msidConfiguration).first)
+        XCTAssertEqual(account?.username, parameters.accountIdentifier.displayableId)
+        XCTAssertEqual(account?.identifier, parameters.accountIdentifier.homeAccountId)
+        XCTAssertEqual(account?.environment, "contoso.com")
+        XCTAssertNil(account?.accountClaims)
+        XCTAssertNoThrow(tokens = try cacheAccessor.getTokens(account: account!, configuration: parameters.msidConfiguration, context: contextStub))
+        XCTAssertEqual(tokens?.accessToken?.accessToken, tokenResponse.accessToken)
+        XCTAssertEqual(tokens?.refreshToken?.refreshToken, tokenResponse.refreshToken)
+        XCTAssertEqual(tokens?.rawIdToken, tokenResponse.idToken)
+        clearCache()
```
antrix1989, line 92: should clearCache be called in the tearDown?

MSAL/test/unit/native_auth/controllers/MSALNativeAuthSignUpControllerTests.swift
```diff
@@ -0,0 +1,1948 @@
...
+
+        await fulfillment(of: [exp], timeout: 1)
+        XCTAssertTrue(helper.onSignUpInvalidAttributesCalled)
+        XCTAssertEqual(helper.newState?.flowToken, "signUpToken 2")
+        XCTAssertEqual(helper.invalidAttributes, ["attribute"])
+
+        checkTelemetryEventResult(id: .telemetryApiIdSignUpSubmitAttributes, isSuccessful: false)
+    }
+
+    // MARK: - Sign-in with SLT
```
antrix1989, line 1787: What the full name of "SLT" acronym?

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
borgesmb: Always BG thread. This is called inside a `Task` from `MSALNativeAuthPublicClientApplication`, which is not a `MainActor`.

MSAL/test/unit/native_auth/controllers/MSALNativeAuthSignUpControllerTests.swift
borgesmb: Added full name (Short-Lived Token)

MSAL/test/unit/native_auth/cache/MSALNativeAuthCacheAccessorTest.swift
borgesmb: Correct. Cache is now boing cleared in 'setup' and 'teardown' only.

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
nilo-ms: Thanks Sergei, what your team usually do in this case? Where the license should be included? Thanks

MSAL/src/native_auth/cache/MSALNativeAuthTokens.swift
borgesmb: This object is not going to be returned externally, but the idea here was to reduce its scope. We just needed these 3 properties (`accessToken`, `refreshToken`, `rawIdToken`); using `MSIDTokenResult` we would also need `account`, `authority` and `correlationId` since they are not nullable.

MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
spetrescu84: On our end, the only case where we needed a `MSIDRequestParameters` instance was when we accessed the cache. The reason why we don't use an instance of `MSIDRequestParameters` from the top layer is because for us a lot of the parameters needed by `MSIDRequestParameters` are not present. For example, one of the main ones is the `redirectURL` which we don't need as we make all calls directly to the server, without the use of the UIWebView use by Auth UX.
Another thing is that we don't need the scopes all the time. For example when we initiate the process of sign up of an user we don't pass in the scopes:
```
public func signUpUsingPassword(
        username: String,
        password: String,
        attributes: [String: Any]? = nil,
        correlationId: UUID? = nil,
        delegate: SignUpPasswordStartDelegate
    )
```
We only pass those when we need to call the token endpoint do do a Sign In operation. One other case where we don't use scopes is the Reset Password operation as they are not needed.
Also because of how CIAM/Microsoft Entra is designed we are missing stuff like `providedAuthority` or `cloudAuthority`
As for the function itself, why it's needed it's used to interact with the `MSIDTokenResponseValidator` class (specifically the `validateAndSaveTokenResponse` function) which doesn't necessarily require the "network" side of the `MSIDRequestParameters` but rather the telemetry and properties that are used to interact with the cache and save the token.
Up to this point, we don't need to create an instance of `MSIDRequestParameters` and this is the only place in our code where we use it and we build it based on the data we pass around - `MSIDConfiguration` and `MSIDRequestContext`

MSAL/src/native_auth/cache/MSALNativeAuthTokens.swift
spetrescu84: It's for public. That `@_implementationOnly import MSAL_Private` is a system we use inside the framework to access Objective C code without exposing it to the Public.
Because `MSALNativeAuthTokens` is part of `MSALNativeAuthUserAccountResult` which is public, we didn't want `MSIDTokenResult` to also be public so we only took parts of it, the 3 tokens mainly.
Let me know if that clears it up

MSAL/src/native_auth/configuration/MSALNativeAuthConfiguration.swift
nilo-ms: That's intended because the MSALCIAMAutority is available publicly to the user (in the MSALNativeAuthPublicClientApplication init), while the MSIDCIAMAuthority is used inside in the SDK. 
We also created this config class, so we don't need to add the "@_implementationOnly import MSAL_Private" import in the MSALNativeAuthPublicClientApplication class. We tried to avoid to have that kind of import in the public classes

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
antrix1989: Emailed the guidance.

antrix1989:
This PR looks good to me, feel free to merge.

antrix1989 (approved):

MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
nilo-ms: Thanks Sergei, we removed the unused "getAccount " method on ADO. We will reflect this change also on GitHub

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
nilo-ms: Thanks Sergei for the email, we will investigate internally. It seems that the component detection task is automatically added as part of every ADO pipeline. I'll better investigate and report it back

PR #1880, opened 2023-09-21
Native Auth network classes that interact with MSAL

borgesmb:
## Proposed changes

This PR shows the SDK that the CIAM team had developed. It is based on the Native Auth merge into MSAL plan here: https://microsofteur.sharepoint.com/:w:/t/DevExDublin/EdfT2AmlFZFIsTtr-EaoSBIBPqcOlRo1MJ1USbEsfjfEcQ

To comply with the design changes made by our API team, we have created a more dynamic SDK. This means that we make different API requests for each authentication flow. There is also some logic involved in each of the responses we get from each request.

We have created a state machine for each flow (check this [Figma](https://www.figma.com/file/2I4zEwwLREDKQOluQ8tgQs/Native-Auth-State-Models?type=design&node-id=0%3A1&t=VUyK4gIoyUChTXWF-1) to see all the possible states), and we are making it external to developers by using delegates instead of completion blocks, as @Serhii Demchenko suggested. In this way we can guide the developer through the authentication process offering a discovering interface. We think that this approach will improve the overall developer experience.

**This PR is focused in the Native Auth base classes that interact with MSAL Objective-C, and their unit tests for:**
**- network**
 **It is not compilable, so tests are not going to run. To build and test this code, please use the `ciam-master-snapshot` branch.**

The IdentityCore changes can be found in this PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/1294

**For context, this is the Native Auth technical overview:** https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?path=/%5BiOS%5D%20Native%20authentication/technical_overview.md&version=GBdanilo/native-authentication&_a=preview

State Machine:
https://www.figma.com/file/2I4zEwwLREDKQOluQ8tgQs/Native-Auth-State-Models?type=design&node-id=0-1&mode=design&t=80VUxa1RvxMn4rnl-0

## Native Auth general high level overview

**MSALNativeAuthPublicClientApplication.swift**
The public interface is the entry point of the SDK. Here is where all the supported authentication flows begin.
Regarding the sign-up flow, developers can call either `signUp(...)` or `signUpUsingPassword(...)`.
Notice that in those methods they must pass a class that conforms to `SignUpStartDelegate`. or `SignUpPasswordStartDelegate`, respectively.

**MSALNativeAuthSignUpController.swift**
The controller handles most of the logic. Each of the methods inside the MARK: - Internal can be called from the public interfaces, that are:
    - MSALNativeAuthPublicClientApplication
    - One of the States (located in the file SignUpDelegates.swift). These States are returned through the delegates, as we will see.

For each of these methods, the controller:
    - Creates a request (`MSIDHttpRequest`) using its requestProvider (`MSALNativeAuthSignUpRequestProviding`).
    - Executes the request.
    - Passes the result to its responseValidator (`MSALNativeAuthSignUpResponseValidating`)
    - Handles the validated result and returns it to the public interface, which uses the delegate to communicate back to the developer.
    - Creates and handle the local telemetry events.

The controller decides what to do with the ValidatedResponse given by the ResponseValidator.

Every validated response usually has:
**A success:** it means that the SDK moves on to the next state, which usually involves making another request or returning back to the user via the delegate.
Redirect: when the backend detects an error and wants the SDK to fallback to the WebView-based flow.
**A known error:** in some cases, these errors put the user in a recoverable state (for example, an error that requires the user to provide more attributes). In other cases, the error means the end of the flow.
**An unexpected error:** these errors are likely bugs, they happen when we there are missing attributes from the response, etc.
(Please keep in mind that not every ValidatedResponse falls into the same categories, this is just an approximation)

**MSALNativeAuthRequestConfigurator.swift**
Each of the requestProviders from the Controllers uses the RequestConfigurator to create and configure the requests. Although there are different RequestProviders (one per flow), there's only one RequestConfigurator.

The RequestConfigurator injects a custom ErrorHandler (`MSALNativeAuthResponseErrorHandler`) to every request. This is needed in order to decode the errors from the API.
These errors follow a structure that you can see in `MSALNativeAuth(SignUp/SignIn/PasswordReset)StartResponseError` for example. In the directory native_auth/network/errors/ you can see all the files used.

**MSALNativeAuth(SignUp/SignIn/PasswordReset)ResponseValidator.swift**
The ResponseValidator is in charge of analysing the api response and returning to the Controller a validated response (for instance, for the response of the (SignUp/SignIn/PasswordReset)/start endpoint, it returns a `MSALNativeAuth(SignUp/SignIn/PasswordReset)StartValidatedResponse`).

**SignUp/SignIn/PasswordReset)States.swift**
In this file there are the different States that the Controller can return to the developer through the delegate.

This is how the developer will use them. For more examples of how developers can use the SDK you can check the Sample App (located in /Samples/ios-native-auth-simple in the branch `ciam-master-snapshot`). We will include it in a separate PR:

extension EmailAndPasswordViewController: SignUpPasswordStartDelegate {

    func onSignUpCodeRequired(newState: MSAL.SignUpCodeRequiredState,
                                                      sentTo _: String,
                                                      channelTargetType _: MSAL.MSALNativeAuthChannelType,
                                                      codeLength _: Int) {

           // show UI to get the user's email OOB code, and wait for user's input
           // then, use the newState to send it to the SDK.
           newState.submitCode(code: code, delegate: self)
    }
}

Each of these states then calls one of the Controller's internal functions, and it follows the same process that we just described in the previous steps.

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/network/MSALNativeAuthResponseErrorHandler.swift
+++ b/MSAL/src/native_auth/network/MSALNativeAuthResponseErrorHandler.swift
@@ -1,0 +1,193 @@
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+@_implementationOnly import MSAL_Private
+
+final class MSALNativeAuthResponseErrorHandler<T: Decodable & Error>: NSObject, MSIDHttpRequestErrorHandling {
+    private var customError: T?
+
+    // swiftlint:disable:next function_parameter_count
+    func handleError(
+        _ error: Error?,
+        httpResponse: HTTPURLResponse?,
+        data: Data?,
+        httpRequest: MSIDHttpRequestProtocol?,
+        responseSerializer: MSIDResponseSerialization?,
+        externalSSOContext ssoContext: MSIDExternalSSOContext?,
+        context: MSIDRequestContext?,
+        completionBlock: MSIDHttpRequestDidCompleteBlock?
+    ) {
+        guard let httpResponse = httpResponse else {
+            completionBlock?(nil, error)
+            return
+        }
+
+        if shouldRetry(httpResponse: httpResponse, httpRequest: httpRequest) {
+            retryRequest(httpRequest: httpRequest,
+                         context: context,
+                         completionBlock: completionBlock)
+            return
+        }
+
+        if httpResponse.statusCode == 400 || httpResponse.statusCode == 401 {
+            // PKeyAuth challenge
+            if let authValue = wwwAuthValue(httpResponse: httpResponse) {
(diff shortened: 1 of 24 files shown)


MSAL/src/native_auth/network/MSALNativeAuthResponseErrorHandler.swift
```diff
@@ -0,0 +1,193 @@
...
+                        context: context,
+                        completionBlock: completionBlock)
+    }
+
+    private func shouldRetry(httpResponse: HTTPURLResponse,
+                             httpRequest: MSIDHttpRequestProtocol?) -> Bool {
+        guard let httpRequest = httpRequest, httpRequest.retryCounter > 0 else {
+            return false
+        }
+        return httpResponse.statusCode >= 500 && httpResponse.statusCode <= 599
```
antrix1989, line 80: I see some logic is common with MSIDAADRequestErrorHandler, should we refactor them to remove copy/paste?

antrix1989: Thanks Danilo, it is a blocker.

antrix1989 (approved):

MSAL/src/native_auth/network/MSALNativeAuthResponseErrorHandler.swift
nilo-ms: Thanks Sergei, we will work on it as soon as possible and we will add you as reviewer of the PR

PR #1914, opened 2023-11-16
Make delegates optional

diegojerezba:
## Proposed changes

This PR makes all delegates optional. The only delegates that are mandatory now are those that return errors.

With extensibility in mind, a new set of classes called "DelegateDispatchers" (name can be changed) has been introduced. The role of these dispatchers is to handle the different versions of delegates that the developer implements as we introduce new possible changes in the public interface, and to return in the delegate that has been implemented. 

If the developer hasn't implemented a delegate necessary to continue the flow, an error is returned.

Example:

```
if let onSignUpCompleted = delegate.onSignUpCompletedNew {
  await onSignUpCompleted(newState, date)
} else if let onSignUpCompleted = delegate.onSignUpCompletedOld {
  await onSignUpCompleted(newState)
} else {
  await delegate.onSignUpError(error: error)
}
```

I also considered the approach of moving the entire response handling to a different class, but I think in this case we would have a class that needs to handle the response + decide which delegate it should use to return to the developer. As our public methods grow, it can become difficult to manage all that code in one single place.

This PR also adds a `newState` parameter to the `SignUpResendCodeDelegate` method:
`@MainActor func onSignUpResendCodeError(error: ResendCodeError, newState: SignUpCodeRequiredState?)`

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
+++ b/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
@@ -115,6 +115,29 @@ class MSALNativeAuthBaseController {
         MSIDTelemetry.sharedInstance().flush(context.telemetryRequestId())
     }
 
+    /// Stops a telemetry event.
+    /// - Parameters:
+    ///   - event: The local event to be stopped.
+    ///   - context: The context object.
+    ///   - delegateDispatcherResult: The result sent by the ``DelegateDispatcher`` that contains whether the developer
+    ///                               has implemented the optional delegate or not.
+    ///   - controllerError: Optional error set by the Controller when handles the response from API.
+    ///                      (ex: SignUpController gets an .attributeValidationFailed. The controller will generate and error and send it here).
+    func stopTelemetryEvent(
+        _ event: MSIDTelemetryAPIEvent?,
+        context: MSIDRequestContext,
+        delegateDispatcherResult: Result<Void, MSALNativeAuthError>,
+        controllerError: MSALNativeAuthError? = nil
+    ) {
+        switch delegateDispatcherResult {
+        case .success:
+            stopTelemetryEvent(event, context: context, error: controllerError)
+        case .failure(let error):
+            MSALLogger.log(level: .error, context: context, format: "Error \(error.errorDescription ?? "No error description")")
+            stopTelemetryEvent(event, context: context, error: error)
+        }
+    }
+
     func complete<T>(
         _ telemetryEvent: MSIDTelemetryAPIEvent?,
         response: T? = nil,
--- a/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
+++ b/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
@@ -83,7 +83,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
         return nil
     }
 
-    func refreshToken(context: MSALNativeAuthRequestContext, authTokens: MSALNativeAuthTokens) async -> Result<String, RetrieveAccessTokenError> {
+    func refreshToken(context: MSALNativeAuthRequestContext, authTokens: MSALNativeAuthTokens) async -> RefreshTokenCredentialControllerResponse {
         MSALLogger.log(level: .verbose, context: context, format: "Refresh started")
         let telemetryEvent = makeAndStartTelemetryEvent(id: .telemetryApiIdRefreshToken, context: context)
         let scopes = authTokens.accessToken?.scopes.array as? [String] ?? []
@@ -93,7 +93,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
             context: context
         ) else {
             stopTelemetryEvent(telemetryEvent, context: context, error: MSALNativeAuthInternalError.invalidRequest)
-            return .failure(RetrieveAccessTokenError(type: .generalError))
+            return .init(.failure(RetrieveAccessTokenError(type: .generalError)))
         }
         let config = factory.makeMSIDConfiguration(scopes: scopes)
         let response = await performAndValidateTokenRequest(request, config: config, context: context)
@@ -145,7 +145,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
         scopes: [String],
         context: MSALNativeAuthRequestContext,
         telemetryEvent: MSIDTelemetryAPIEvent?
-    ) -> Result<String, RetrieveAccessTokenError> {
+    ) -> RefreshTokenCredentialControllerResponse {
         let config = factory.makeMSIDConfiguration(scopes: scopes)
         switch response {
         case .success(let tokenResponse):
@@ -162,7 +162,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
                 context: context,
                 format: "Refresh Token completed with error: \(error.errorDescription ?? "No error description")")
             stopTelemetryEvent(telemetryEvent, context: context, error: error)
-            return .failure(error)
+            return .init(.failure(error))
         }
     }
 
@@ -171,22 +171,25 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
         telemetryEvent: MSIDTelemetryAPIEvent?,
         context: MSALNativeAuthRequestContext,
         config: MSIDConfiguration
-    ) -> Result<String, RetrieveAccessTokenError> {
+    ) -> RefreshTokenCredentialControllerResponse {
         do {
             let tokenResult = try cacheTokenResponse(tokenResponse, context: context, msidConfiguration: config)
-            telemetryEvent?.setUserInformation(tokenResult.account)
-            stopTelemetryEvent(telemetryEvent, context: context)
             MSALLogger.log(
                 level: .verbose,
                 context: context,
                 format: "Refresh Token completed successfully")
-            return .success(tokenResult.accessToken.accessToken)
+            return .init(.success(tokenResult.accessToken.accessToken), telemetryUpdate: { [weak self] result in
+                telemetryEvent?.setUserInformation(tokenResult.account)
+                self?.stopTelemetryEvent(telemetryEvent, context: context, delegateDispatcherResult: result)
+            })
         } catch {
+            let error = RetrieveAccessTokenError(type: .generalError)
             MSALLogger.log(
                 level: .error,
                 context: context,
                 format: "Token Result was not created properly error - \(error)")
-            return .failure(RetrieveAccessTokenError(type: .generalError))
+            stopTelemetryEvent(telemetryEvent, context: context, error: error)
+            return .init(.failure(error))
         }
     }
 }
--- a/MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
+++ b/MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
@@ -56,19 +56,19 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
 
     // MARK: - Internal interface methods
 
-    func resetPassword(parameters: MSALNativeAuthResetPasswordStartRequestProviderParameters) async -> ResetPasswordStartResult {
+    func resetPassword(parameters: MSALNativeAuthResetPasswordStartRequestProviderParameters) async -> ResetPasswordStartControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdResetPasswordStart, context: parameters.context)
         let response = await performStartRequest(parameters: parameters)
         return await handleStartResponse(response, event: event, context: parameters.context)
     }
 
-    func resendCode(passwordResetToken: String, context: MSIDRequestContext) async -> ResetPasswordResendCodeResult {
+    func resendCode(passwordResetToken: String, context: MSIDRequestContext) async -> ResetPasswordResendCodeControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdResetPasswordResendCode, context: context)
         let response = await performChallengeRequest(passwordResetToken: passwordResetToken, context: context)
         return await handleResendCodeChallengeResponse(response, event: event, context: context)
     }
 
-    func submitCode(code: String, passwordResetToken: String, context: MSIDRequestContext) async -> ResetPasswordVerifyCodeResult {
+    func submitCode(code: String, passwordResetToken: String, context: MSIDRequestContext) async -> ResetPasswordSubmitCodeControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdResetPasswordSubmitCode, context: context)
 
         let params = MSALNativeAuthResetPasswordContinueRequestParameters(
@@ -86,7 +86,7 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
         password: String,
         passwordSubmitToken: String,
         context: MSIDRequestContext
-    ) async -> ResetPasswordRequiredResult {
+    ) async -> ResetPasswordSubmitPasswordControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdResetPasswordSubmit, context: context)
 
         let params = MSALNativeAuthResetPasswordSubmitRequestParameters(
@@ -120,7 +120,7 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
 
     private func handleStartResponse(_ response: MSALNativeAuthResetPasswordStartValidatedResponse,
                                      event: MSIDTelemetryAPIEvent?,
-                                     context: MSIDRequestContext) async -> ResetPasswordStartResult {
+                                     context: MSIDRequestContext) async -> ResetPasswordStartControllerResponse {
 
         MSALLogger.log(level: .verbose, context: context, format: "Finished resetpassword/start request")
 
@@ -134,21 +134,21 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
             MSALLogger.log(level: .error,
                            context: context,
                            format: "redirect error in resetpassword/start request \(error.errorDescription ?? "No error description")")
-            return .error(error)
+            return .init(.error(error))
         case .error(let apiError):
             let error = apiError.toResetPasswordStartPublicError()
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in resetpassword/start request \(error.errorDescription ?? "No error description")")
-            return .error(error)
+            return .init(.error(error))
         case .unexpectedError:
             let error = ResetPasswordStartError(type: .generalError)
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Unexpected error in resetpassword/start request \(error.errorDescription ?? "No error description")")
-            return .error(error)
+            return .init(.error(error))
         }
     }
 
@@ -177,39 +177,40 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
         _ response: MSALNativeAuthResetPasswordChallengeValidatedResponse,
         event: MSIDTelemetryAPIEvent?,
         context: MSIDRequestContext
-    ) async -> ResetPasswordStartResult {
+    ) async -> ResetPasswordStartControllerResponse {
         switch response {
         case .success(let sentTo, let channelTargetType, let codeLength, let challengeToken):
             MSALLogger.log(level: .info, context: context, format: "Successful resetpassword/challenge request")
-            stopTelemetryEvent(event, context: context)
 
-            return .codeRequired(
+            return .init(.codeRequired(
                 newState: ResetPasswordCodeRequiredState(controller: self, flowToken: challengeToken, correlationId: context.correlationId()),
                 sentTo: sentTo,
                 channelTargetType: channelTargetType,
                 codeLength: codeLength
-            )
+            ), telemetryUpdate: { [weak self] result in
+                self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result)
+            })
         case .error(let apiError):
             let error = apiError.toResetPasswordStartPublicError()
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in resetpassword/challenge request \(error.errorDescription ?? "No error description")")
-            return .error(error)
+            return .init(.error(error))
         case .redirect:
             let error = ResetPasswordStartError(type: .browserRequired, message: MSALNativeAuthErrorMessage.browserRequired)
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Redirect error in resetpassword/challenge request \(error.errorDescription ?? "No error description")")
-            return .error(error)
+            return .init(.error(error))
         case .unexpectedError:
             let error = ResetPasswordStartError(type: .generalError)
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Unexpected error in resetpassword/challenge request \(error.errorDescription ?? "No error description")")
-            return .error(error)
+            return .init(.error(error))
         }
     }
 
@@ -217,24 +218,25 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
         _ response: MSALNativeAuthResetPasswordChallengeValidatedResponse,
         event: MSIDTelemetryAPIEvent?,
         context: MSIDRequestContext
-    ) async -> ResetPasswordResendCodeResult {
+    ) async -> ResetPasswordResendCodeControllerResponse {
         switch response {
         case .success(let sentTo, let channelTargetType, let codeLength, let challengeToken):
-            stopTelemetryEvent(event, context: context)
             MSALLogger.log(level: .info, context: context, format: "Successful resetpassword/challenge (resend code) request")
-            return .codeRequired(
+            return .init(.codeRequired(
                 newState: ResetPasswordCodeRequiredState(controller: self, flowToken: challengeToken, correlationId: context.correlationId()),
                 sentTo: sentTo,
                 channelTargetType: channelTargetType,
                 codeLength: codeLength
-            )
+            ), telemetryUpdate: { [weak self] result in
+                self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result)
+            })
         case .error(let apiError):
             let error = apiError.toResendCodePublicError()
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in resetpassword/challenge request (resend code) \(error.errorDescription ?? "No error description")")
-            return .error(error: error, newState: nil)
+            return .init(.error(error: error, newState: nil))
         case .redirect,
                 .unexpectedError:
             let error = ResendCodeError()
@@ -242,7 +244,7 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in resetpassword/challenge request (resend code) \(error.errorDescription ?? "No error description")")
-            return .error(error: error, newState: nil)
+            return .init(.error(error: error, newState: nil))
         }
     }
 
@@ -271,21 +273,22 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
         passwordResetToken: String,
         event: MSIDTelemetryAPIEvent?,
         context: MSIDRequestContext
-    ) async -> ResetPasswordVerifyCodeResult {
+    ) async -> ResetPasswordSubmitCodeControllerResponse {
         switch response {
         case .success(let passwordSubmitToken):
-            stopTelemetryEvent(event, context: context)
             MSALLogger.log(level: .info, context: context, format: "Successful resetpassword/continue request")
-            return .passwordRequired(newState: ResetPasswordRequiredState(controller: self,
-                                                                          flowToken: passwordSubmitToken,
-                                                                          correlationId: context.correlationId()))
+            let newState = ResetPasswordRequiredState(controller: self, flowToken: passwordSubmitToken, correlationId: context.correlationId())
+            return .init(.passwordRequired(newState: newState),
+                         telemetryUpdate: { [weak self] result in
+                self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result)
+            })
         case .error(let apiError):
             let error = apiError.toVerifyCodePublicError()
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in resetpassword/continue request \(error.errorDescription ?? "No error description")")
-            return .error(error: error, newState: nil)
+            return .init(.error(error: error, newState: nil))
         case .unexpectedError:
             let error = VerifyCodeError(type: .generalError)
             self.stopTelemetryEvent(event, context: context, error: error)
@@ -294,7 +297,7 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
                            context: context,
                            format: "Error calling resetpassword/continue \(error.errorDescription ?? "No error description")")
 
-            return .error(error: error, newState: nil)
+            return .init(.error(error: error, newState: nil))
         case .invalidOOB:
             let error = VerifyCodeError(type: .invalidCode)
             self.stopTelemetryEvent(event, context: context, error: error)
@@ -304,7 +307,7 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
                            format: "Invalid code error calling resetpassword/continue \(error.errorDescription ?? "No error description")")
 
             let state = ResetPasswordCodeRequiredState(controller: self, flowToken: passwordResetToken, correlationId: context.correlationId())
-            return .error(error: error, newState: state)
+            return .init(.error(error: error, newState: state))
         }
     }
 
@@ -333,7 +336,7 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
         passwordSubmitToken: String,
         event: MSIDTelemetryAPIEvent?,
         context: MSIDRequestContext
-    ) async -> ResetPasswordRequiredResult {
+    ) async -> ResetPasswordSubmitPasswordControllerResponse {
         MSALLogger.log(level: .info, context: context, format: "Finished resetpassword/submit request")
 
         switch response {
@@ -352,10 +355,8 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Password error calling resetpassword/submit \(error.errorDescription ?? "No error description")")
-
-            return .error(error: error, newState: ResetPasswordRequiredState(controller: self,
-                                                                             flowToken: passwordSubmitToken,
-                                                                             correlationId: context.correlationId()))
+            let newState = ResetPasswordRequiredState(controller: self, flowToken: passwordSubmitToken, correlationId: context.correlationId())
+            return .init(.error(error: error, newState: newState))
         case .error(let apiError):
             let error = apiError.toPasswordRequiredPublicError()
             self.stopTelemetryEvent(event, context: context, error: error)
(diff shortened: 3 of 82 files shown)


MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -115,6 +115,21 @@ class MSALNativeAuthBaseController {
         MSIDTelemetry.sharedInstance().flush(context.telemetryRequestId())
     }
 
+    func updateTelemetryEvent(
```
nilo-ms, line 118: Three comments about this method:
- The name doesn't clarify that the telemetry event would be always stopped. I'd include the word "stop" or "complete" to this method name
- Given that we're using a generic error message, this can make the debug a little bit more difficult
- Why we have 2 MSALNativeAuthError? The first one is in the "failure" result (line 121), while the second one is a default parameter (line 122). At line 126 why we use the variable error even if the result is successful?

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
```diff
@@ -37,6 +37,10 @@ enum MSALNativeAuthErrorMessage {
     static let passwordRequiredNotImplemented = "Implementation of onSignInPasswordRequired required"
     static let codeRequiredNotImplemented = "Implementation of onSignInCodeRequired required"
     static let codeRequiredForPasswordUserLog = "This user does not have a password associated with their account. SDK will call `delegate.onSignInCodeRequired()` and the entered password will be ignored"
+
+    static func requiredDelegateMethod(_ method: String) -> String {
+        return "Implementation of \(method) required to continue this flow"
```
nilo-ms, line 42: Is there a specific reason you didn't use the "%@" as we did for `attributeValidationFailed` string? 
Given that we have this generic string/method can you please unify `passwordRequiredNotImplemented` and `codeRequiredNotImplemented`?
I'd also remove `to continue this flow` to this message

MSAL/src/native_auth/public/state_machine/delegate/ResetPasswordDelegates.swift
```diff
@@ -88,5 +88,5 @@ public protocol ResetPasswordRequiredDelegate {
     @MainActor func onResetPasswordRequiredError(error: PasswordRequiredError, newState: ResetPasswordRequiredState?)
 
     /// Notifies the delegate that the reset password operation completed successfully.
-    @MainActor func onResetPasswordCompleted()
+    @MainActor @objc optional func onResetPasswordCompleted()
```
nilo-ms, line 91: can you please add a note in the comment that says something like (we already have these kinds of notes):
`    /// - Note: If a flow requires this optional method and it is not implemented, then ``onResetPasswordRequiredError(error)`` will be called. ` 
I'd say to use a generic note whenever possible

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordControlling.swift
```diff
@@ -26,15 +26,20 @@
 
 protocol MSALNativeAuthResetPasswordControlling: AnyObject {
 
-    func resetPassword(parameters: MSALNativeAuthResetPasswordStartRequestProviderParameters) async -> ResetPasswordStartResult
+    typealias ResetPasswordStartControllerResponse = MSALNativeAuthControllerTelemetryWrapper<ResetPasswordStartResult>
+    typealias ResetPasswordResendCodeControllerResponse = MSALNativeAuthControllerTelemetryWrapper<ResetPasswordResendCodeResult>
+    typealias ResetPasswordSubmitCodeControllerResponse = MSALNativeAuthControllerTelemetryWrapper<ResetPasswordVerifyCodeResult>
```
rodhan-ms, line 31: I think `ResetPasswordVerifyCodeResult` should be renamed to `ResetPasswordSubmitCodeResult`. I know this is from a previous PR but it stands out as wrong in this PR, especially in mapping lines like this.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -171,7 +171,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
         telemetryEvent: MSIDTelemetryAPIEvent?,
         context: MSALNativeAuthRequestContext,
         config: MSIDConfiguration
-    ) -> Result<String, RetrieveAccessTokenError> {
+    ) -> RefreshTokenCredentialControllerResponse {
         do {
             let tokenResult = try cacheTokenResponse(tokenResponse, context: context, msidConfiguration: config)
             telemetryEvent?.setUserInformation(tokenResult.account)
```
rodhan-ms, line 177: From my understanding, the telemetryEvent should no longer be stopped on the following line because the operation might yet be unsuccessful, and we are already calling the new `updateTelemetryEvent` later in the flow (line 184). This is the same in a few places in the new code.

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
```diff
@@ -453,13 +460,15 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
             case .succeeded:
                 stopTelemetryEvent(event, context: context)
```
rodhan-ms, line 461: Telemetry event should not be stopped here as it will be at line 146

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
```diff
@@ -0,0 +1,201 @@
...
+
+final class SignUpPasswordStartDelegateDispatcher: DelegateDispatcher<SignUpPasswordStartDelegate> {
+
+    func dispatchSignUpCodeRequired(
+        newState: SignUpCodeRequiredState,
+        sentTo: String,
+        channelTargetType: MSALNativeAuthChannelType,
+        codeLength: Int
+    ) async {
+        if let signUpCodeRequired = delegate.onSignUpCodeRequired {
```
rodhan-ms, line 35: For the sign in dispatcher you used a different pattern for this variable name, e.g. this variable would be called `onSignUpCodeRequired` instead of `signUpCodeRequired`. I'm already a bit worried about having to keep three things in sync - the delegate method name, the variable referring to the delegate method and the string for the error  message (line 39) - but I don't see any easy clean way that would improve things. At least I think we should use the exact same name for all three uses.

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
```diff
@@ -0,0 +1,201 @@
...
+            await signUpCodeRequired(newState, sentTo, channelTargetType, codeLength)
+        } else {
+            let error = SignUpPasswordStartError(type: .generalError, message: errorMessage(for: "onSignUpCodeRequired"))
+            telemetryUpdate?(.failure(error))
+            await delegate.onSignUpPasswordError(error: error)
+        }
+    }
+
+    func dispatchSignUpAttributesInvalid(attributeNames: [String]) async {
+        if let signUpAttributes = delegate.onSignUpAttributesInvalid {
```
rodhan-ms, line 46: This is a worse mismatch of the variable name <> function  mapping that I mentioned in another comment. I think this variable should be called `onSignUpAttributesInvalid` for consistency.

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
```diff
@@ -0,0 +1,201 @@
...
+
+final class SignUpStartDelegateDispatcher: DelegateDispatcher<SignUpStartDelegate> {
+
+    func dispatchSignUpCodeRequired(
+        newState: SignUpCodeRequiredState,
+        sentTo: String,
+        channelTargetType: MSALNativeAuthChannelType,
+        codeLength: Int
+    ) async {
+        if let signUpCodeRequired = delegate.onSignUpCodeRequired {
```
rodhan-ms, line 65: Curious to know if you considered `guard let` instead of `if let` for all the dispatcher methods?  I think it would be clearer/more swifty.

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordControlling.swift
```diff
@@ -26,15 +26,20 @@
...
+    typealias ResetPasswordSubmitPasswordControllerResponse = MSALNativeAuthControllerTelemetryWrapper<ResetPasswordRequiredResult>
```
spetrescu84, line 32: Should this be `MSALNativeAuthControllerTelemetryWrapper<ResetPasswordSubmitPasswordResult>` to keep in sync with the rest?

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInControlling.swift
```diff
@@ -27,6 +27,10 @@ import Foundation
 protocol MSALNativeAuthSignInControlling {
     typealias SignInPasswordControllerResponse = MSALNativeAuthControllerTelemetryWrapper<SignInPasswordStartResult>
     typealias SignInCodeControllerResponse = MSALNativeAuthControllerTelemetryWrapper<SignInStartResult>
+    typealias SignInSltControllerResponse = MSALNativeAuthControllerTelemetryWrapper<Result<MSALNativeAuthUserAccountResult, SignInAfterSignUpError>>
```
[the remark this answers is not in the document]
diegojerezba: Yes, I think you're right. Changed. 

One question, what is your opinion about moving all the `typealias` outside of the protocol? I think it is too long for the internal functions to return:

`func foo() -> MSALNativeAuthSignInControlling.SignInAfterSignUpControllerResponse`
`func foo() -> MSALNativeAuthResetPasswordControlling.ResetPasswordSubmitPasswordControllerResponse`
`func foo() -> MSALNativeAuthSignUpControlling.SignUpResendCodeControllerResponse`

etc.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: 1. I agree that this is not a great name. I've renamed it as `stopTelemetryEvent` as an overload of the others that also stop telemetry, and changed the parameter names to make it clearer.
2. `error.errorDescription` contains the name of the method that hasn't been implemented. The log would be like this: `"Error MSALNativeAuth has called the delegate method onSignUpCompleted that has not been implemented"`
3. I've added documentation for this. In some cases the controller creates an error (for example the SignUpController in the `.attributeValidationFailed`. That error goes here. The `failure` result is related to the developer not having implemented the optional delegate.

```swift
func stopTelemetryEvent(
  _ event: MSIDTelemetryAPIEvent?,
  delegateDispatcherResult: Result<Void, MSALNativeAuthError>,
  context: MSIDRequestContext,
  controllerError: MSALNativeAuthError? = nil
)
```

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
diegojerezba: Done. Sorry I had forgotten about the %@, I just remember that @rodhan-ms told me this in a previous PR!

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordControlling.swift
diegojerezba: Done

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
diegojerezba: That's a good catch, thank you. I've reviewed every controller, and found that it was happening in other places as well.

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
diegojerezba: Done (same as [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1914#discussion_r1400963221))

MSAL/src/native_auth/public/state_machine/delegate/DelegateDispatcher.swift
```diff
@@ -0,0 +1,39 @@
...
+class DelegateDispatcher<T> {
+    let delegate: T
+    let telemetryUpdate: ((Result<Void, MSALNativeAuthError>) -> Void)?
+
+    init(delegate: T, telemetryUpdate: ((Result<Void, MSALNativeAuthError>) -> Void)?) {
+        self.delegate = delegate
+        self.telemetryUpdate = telemetryUpdate
+    }
+
+    func errorMessage(for method: String) -> String {
```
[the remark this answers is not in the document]
diegojerezba: Done

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
diegojerezba: I've reviewed all of them and renamed `onXXX` as you suggest. Thanks for the feedback.

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
diegojerezba: Done (same as [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1914#discussion_r1400964541))

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
diegojerezba: I considered it, but if in the future we have more delegates, it will have to be changed to an `if/else if/else`. 
```swift
if let onSignUpCompletedNew = delegate.onSignUpCompletedNew {
  await onSignUpCompletedNew(newState, date)
} else if let onSignUpCompletedOld = delegate.onSignUpCompletedOld {
  await onSignUpCompletedOld(newState)
} else {
  await delegate.onSignUpError(error: error)
}
```

By using if/let now, we have the whole class written in the same style, even if we have newer delegates in the future.

MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordControlling.swift
diegojerezba: Yes, it makes sense. Updated 👍

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegateDispatchers.swift
rodhan-ms: Thanks that makes a lot of sense. I knew you would have a good reason 😀

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -625,51 +606,57 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
...
-
-            return .attributesRequired(attributes: attributes, state: state)
+            let state = SignUpAttributesRequiredState(
+                controller: self,
+                username: username,
+                flowToken: signUpToken,
+                correlationId: context.correlationId()
+            )
+            return .init(.attributesRequired(attributes: attributes, state: state), telemetryUpdate: { [weak self] result in
+                self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result, controllerError: error)
```
nilo-ms, line 635: Hey Diego, can you help me understand this piece of code? We stop the telemetry using the delegateDispatcherResult that can be an error in case the external developer doesn't implements an optional delegate method. My question is about the controllerError. I suppose `error` is created at line 624, and it always be not nil, so this telemetry event will be always marked as a "failure" event, even if the external developer correctly implemented the optional delegate method. Is this behaviour expected?

diegojerezba: Yes, it is expected behavior. In this case, `.attributesRequired` is happening after the user has already sent some attributes before (the piece of code belongs to the function `handleSubmitAttributesResult()`.

 `.attributesRequired` can also happen after the user submits the code and submits the password. In those cases it is not considered an error, but a part of the flow. This was part of the original design.

So, now with optional delegates, this is what happens:

1. User sends some attributes, and receives that more attributes are needed (we consider this an error)
2. We check if the user has implemented `onSignUpAttributesRequired(attributes:newState)`
  a. If they have, we return in that method, and we finish the event with an error about attributes are needed.
  b. If they don't have this method, we return in `onSignUpAttributesRequiredError(error:)` saying that the user has to implement the other delegate. We also finish telemetry with the same error that says that the user needs to implement that method (in this case, the error about not having implemented the delegate is more important than not having sent some attributes)

nilo-ms: Ok, thanks for the clarification. Can you please add a comment that says that the telemetry event in this case will always be a failure because the user didn't submit all the required attributes. Thanks

diegojerezba: Added in those places where this happens.

MSAL/src/native_auth/controllers/responses/SignUpResults.swift
```diff
@@ -72,18 +72,9 @@ enum SignUpVerifyCodeResult {
...
-    /// - newState: An object representing the new state of the flow with follow on methods.
-    /// - sentTo: The email/phone number that the code was sent to.
-    /// - channelTargetType: The channel (email/phone) the code was sent through.
-    /// - codeLength: The length of the code required.
-    case codeRequired(newState: SignUpCodeRequiredState, sentTo: String, channelTargetType: MSALNativeAuthChannelType, codeLength: Int)
-
-    /// An error object indicating why the operation failed.
-    case error(ResendCodeError)
-}
+/// Result type that contains information about the code sent, the next state of the reset password process and possible errors.
```
nilo-ms, line 75: nit: these comments are here because we were planning to make these result public with the async/await introduction. Given that async/await isn't in the plan anymore, should we still keep updating this internal comments?

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignInDelegateDispatchers.swift
```diff
@@ -0,0 +1,132 @@
...
+
+final class SignInStartDelegateDispatcher: DelegateDispatcher<SignInStartDelegate> {
+
+    func dispatchSignInCodeRequired(
+        newState: SignInCodeRequiredState,
+        sentTo: String,
+        channelTargetType: MSALNativeAuthChannelType,
+        codeLength: Int
+    ) async {
+        if let onSignInCodeRequired = delegate.onSignInCodeRequired {
```
nilo-ms, line 65: it seems to see a pattern here:
`if delegate.expectedMethod is implemented {
    telemetryUpdate?(.success(()))
    call expectedMethod
 } else {
      create error
      telemetryUpdate?(.failure(error))
      call delegate.error
} `
In the controller classes we used to use function that accepted a generic success and error delegate methods. Do you think is it possible to applies something similar here? We could save a lot of duplicate code.

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInControlling.swift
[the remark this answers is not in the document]
spetrescu84: The downside of moving them outside is that if in the future we need to add one that has the same name to another protocol we can't. Technically they should be unique.
I'd prefer to keep them inside even if they are longer to write, at least the belonging is clear

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignInDelegateDispatchers.swift
diegojerezba: I did it but I don't like the result, it's too complex for the small logic we have in this case. In the controllers I agree, that it's a great choice because it saves us a lot of duplication, but here I'm not sure. This is how it looks like after having created two generic functions in the parent class `DelegateDispatcher`:

```swift
    func dispatchSignInCodeRequired(
        newState: SignInCodeRequiredState,
        sentTo: String,
        channelTargetType: MSALNativeAuthChannelType,
        codeLength: Int
    ) async {
        if let onSignInCodeRequired = delegate.onSignInCodeRequired {
            await completeDelegate { await onSignInCodeRequired(newState, sentTo, channelTargetType, codeLength) }
        } else {
            let error = SignInPasswordStartError(type: .generalError, message: requiredErrorMessage(for: "onSignInCodeRequired"))
            await completeDelegateError(error) { [weak self] in
                await self?.delegate.onSignInPasswordError(error: error)
            }
        }
    }
```

If you have a different thing in mind, let me know.

MSAL/src/native_auth/controllers/responses/SignUpResults.swift
diegojerezba: Good point, I've removed them.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -171,22 +171,25 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
...
+                self?.stopTelemetryEvent(telemetryEvent, context: context, delegateDispatcherResult: result)
+            })
         } catch {
+            let error = RetrieveAccessTokenError(type: .generalError)
             MSALLogger.log(
                 level: .error,
                 context: context,
                 format: "Token Result was not created properly error - \(error)")
-            return .failure(RetrieveAccessTokenError(type: .generalError))
+            stopTelemetryEvent(telemetryEvent, context: context, error: error)
```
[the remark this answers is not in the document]
diegojerezba: When I was reviewing this class, I saw that the `stopTelemetryEvent` was missing from here. This is the `catch` statement after caching the token.

MSAL/test/unit/native_auth/mock/MSALNativeAuthResetPasswordControllerMock.swift
```diff
@@ -28,24 +28,24 @@ import XCTest
 
 class MSALNativeAuthResetPasswordControllerMock: MSALNativeAuthResetPasswordControlling {
 
-    var resetPasswordResult: ResetPasswordStartResult!
-    var resendCodeResult: ResetPasswordResendCodeResult!
-    var submitCodeResult: ResetPasswordVerifyCodeResult!
-    var submitPasswordResult: ResetPasswordRequiredResult!
+    var resetPasswordResult: ResetPasswordStartControllerResponse!
```
spetrescu84, line 31: Should we rename all of these to Response as they are now responses?

MSAL/test/unit/native_auth/public/delegate/sign_in/SignInStartDelegateDispatcherTests.swift
```diff
@@ -0,0 +1,151 @@
...
+
+        await fulfillment(of: [telemetryExp, delegateExp])
+
+        XCTAssertEqual(delegate.expectedState, expectedState)
+        XCTAssertEqual(delegate.expectedSentTo, expectedSentTo)
+        XCTAssertEqual(delegate.expectedChannelTargetType, expectedChannelTargetType)
+        XCTAssertEqual(delegate.expectedCodeLength, expectedCodeLength)
+    }
+
+    func test_dispatchSignUpCodeRequired_whenDelegateOptionalMethodsNotImplemented() async {
```
spetrescu84, line 71: This should be `test_dispatchSignInCodeRequired_whenDelegateOptionalMethodsNotImplemented` not `test_dispatchSignUpCodeRequired_whenDelegateOptionalMethodsNotImplemented` SignIn vs SignUp

MSAL/test/unit/native_auth/public/delegate/sign_up/SignUpPasswordStartDelegateDispatcherTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            }
+            self.telemetryExp.fulfill()
+        })
+
+        let expectedState = SignUpCodeRequiredState(controller: controllerFactoryMock.signUpController, username: "", flowToken: "flowToken", correlationId: correlationId)
+        let expectedSentTo = "user@contoso.com"
+        let expectedChannelTargetType = MSALNativeAuthChannelType.email
+        let expectedCodeLength = 4
+
+        await sut.dispatchSignUpCodeRequired(
```
spetrescu84, line 57: Is this call correctly being made when the tests are for PasswordStart?
Also the unit test name `test_dispatchSignUpCodeRequired_whenDelegateMethodsAreImplemented` is it correct?

MSAL/test/unit/native_auth/public/delegate/sign_up/SignUpPasswordStartDelegateDispatcherTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            checkError(customError)
+            self.telemetryExp.fulfill()
+        })
+
+        let expectedState = SignUpCodeRequiredState(controller: controllerFactoryMock.signUpController, username: "", flowToken: "flowToken", correlationId: correlationId)
+        let expectedSentTo = "user@contoso.com"
+        let expectedChannelTargetType = MSALNativeAuthChannelType.email
+        let expectedCodeLength = 4
+
+        await sut.dispatchSignUpCodeRequired(
```
spetrescu84, line 90: Is this call correctly being made when the tests are for PasswordStart?
Also the unit test name test_dispatchSignUpCodeRequired_whenDelegateOptionalMethodsNotImplemented is it correct?

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignInDelegateDispatchers.swift
nilo-ms: I thought we could use something similar as we used in the SignInController, where we were passing a code block as method parameter (as Silviu is showing in his comment). But if it would result in an addition of too much complexity I'm happy to leave it as it is

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -407,28 +395,37 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
...
-            return .codeRequired(
-                newState: SignUpCodeRequiredState(controller: self,
-                                                  username: username,
-                                                  flowToken: signUpToken,
-                                                  correlationId: context.correlationId()),
+            return .init(.codeRequired(
+                newState: SignUpCodeRequiredState(
+                    controller: self,
+                    username: username,
+                    flowToken: signUpToken, correlationId: context.correlationId()
```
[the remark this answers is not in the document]
diegojerezba: Done

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
```diff
@@ -25,7 +25,7 @@
 // swiftlint:disable line_length
 enum MSALNativeAuthErrorMessage {
     static let invalidScope = "Invalid scope"
-    static let delegateNotImplemented = "MSALNativeAuth has called an optional delegate method that has not been implemented"
+    static let requiredDelegateMethod = "MSALNativeAuth has called the delegate method %@ that has not been implemented"
```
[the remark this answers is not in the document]
diegojerezba: You're right, I shouldn't have changed it.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
```diff
@@ -117,9 +117,9 @@ extension MSALNativeAuthPublicClientApplication {
         return await controller.signIn(params: params)
     }
 
-    func resetPasswordInternal(username: String, correlationId: UUID?) async -> ResetPasswordStartResult {
+    func resetPasswordInternal(username: String, correlationId: UUID?) async -> MSALNativeAuthResetPasswordControlling.ResetPasswordStartControllerResponse {
```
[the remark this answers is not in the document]
diegojerezba: Thanks. Done.

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInControlling.swift
```diff
@@ -27,6 +27,10 @@ import Foundation
 protocol MSALNativeAuthSignInControlling {
     typealias SignInPasswordControllerResponse = MSALNativeAuthControllerTelemetryWrapper<SignInPasswordStartResult>
     typealias SignInCodeControllerResponse = MSALNativeAuthControllerTelemetryWrapper<SignInStartResult>
+    typealias SignInAfterSignUpControllerResponse = MSALNativeAuthControllerTelemetryWrapper<Result<MSALNativeAuthUserAccountResult, SignInAfterSignUpError>>
```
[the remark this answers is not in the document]
diegojerezba: This was commented [in the PR](https://identitydivision.visualstudio.com/DefaultCollection/Engineering/_git/devexdub-microsoft-authentication-library-for-objc/pullRequest/9361#1694532685) where we introduced it, and we decided to name them `Response`, so I'd rather keep them.

But thanks for catching the warning, I've updated the line.

MSAL/test/unit/native_auth/mock/MSALNativeAuthResetPasswordControllerMock.swift
diegojerezba: Yes, nice catch. Done.

MSAL/test/unit/native_auth/public/delegate/sign_in/SignInStartDelegateDispatcherTests.swift
diegojerezba: Thanks, changed.

MSAL/test/unit/native_auth/public/delegate/sign_up/SignUpPasswordStartDelegateDispatcherTests.swift
diegojerezba: yes, the call is correct, but I think I see where the confusion comes from. I've renamed the method in the `SignUpPasswordStartDelegateDispatcher` to `dispatchSignUpPasswordCodeRequired` to make it clearer.

Same for the test names.

MSAL/test/unit/native_auth/public/delegate/sign_up/SignUpPasswordStartDelegateDispatcherTests.swift
diegojerezba: Changed (same as [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1914#discussion_r1404549624))

diegojerezba:
FYI: I've made the changes to the delegate errors mentioned in the error PR [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1923#issuecomment-1825875757) to avoid conflicts in the delegateDispatchers.

MSAL/src/native_auth/public/state_machine/delegate/SignInDelegates.swift
```diff
@@ -44,28 +44,30 @@ public protocol SignInPasswordStartDelegate {
 
     /// Notifies the delegate that the sign in operation completed successfully.
     /// - Parameter result: An object representing the signed in user account.
-    @MainActor func onSignInCompleted(result: MSALNativeAuthUserAccountResult)
+    /// - Note: If a flow requires a code but this optional method is not implemented, then ``onSignInPasswordError(error:)`` will be called.
```
rodhan-ms, line 47: I think this docc comment is wrong because it shouldn't mention requiring a code, should it? I think it is better to use the generic "Note: If a flow requires this optional method and it is not implemented..." text that you used in other cases.

MSAL/src/native_auth/public/state_machine/delegate/SignInDelegates.swift
```diff
@@ -110,6 +116,7 @@ public protocol SignInVerifyCodeDelegate {
     @MainActor func onSignInVerifyCodeError(error: VerifyCodeError, newState: SignInCodeRequiredState?)
 
     /// Notifies the delegate that the sign in operation completed successfully.
+    /// - Note: If a flow requires a code but this optional method is not implemented, then ``onSignInVerifyCodeError(error:newState:)`` will be called.
```
rodhan-ms, line 119: Same here and in other places too, the doc comment is incorrect because it talks about a code being required.

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
```diff
@@ -28,21 +28,22 @@ import Foundation
 public protocol SignUpPasswordStartDelegate {
     /// Notifies the delegate that the operation resulted in an error.
     /// - Parameter error: An error object indicating why the operation failed.
-    @MainActor func onSignUpPasswordError(error: SignUpPasswordStartError)
+    @MainActor func onSignUpPasswordStartError(error: SignUpPasswordStartError)
 
     /// Notifies the delegate that a verification code is required from the user to continue.
+    /// - Note: If a flow requires this optional method and it is not implemented, then ``onSignUpPasswordError(error:)`` will be called.
```
rodhan-ms, line 34: This is now incorrect because you renamed the error method! Should say "...onSignUpPasswordStartError(error:)..."

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
```diff
@@ -142,18 +149,21 @@ public protocol SignUpAttributesRequiredDelegate {
...
     /// Notifies the delegate that there are some required attributes to be sent.
+    /// - Note: If a flow requires attributes but this optional method is not implemented, then ``onSignUpAttributesRequiredError(error:)`` will be called.
     /// - Parameters:
     ///     - attributes:  List of required attributes.
     ///     - newState: An object representing the new state of the flow with follow on methods.
-    @MainActor func onSignUpAttributesRequired(attributes: [MSALNativeAuthRequiredAttributes], newState: SignUpAttributesRequiredState)
+    @MainActor @objc optional func onSignUpAttributesRequired(attributes: [MSALNativeAuthRequiredAttributes], newState: SignUpAttributesRequiredState)
 
     /// Notifies the delegate that invalid attributes were sent.
+    /// - Note: If a flow requires attributes but this optional method is not implemented, then ``onSignUpAttributesRequiredError(error:)`` will be called.
```
rodhan-ms, line 159: The comment says is attributes are required, but the method is for attributes invalid. As said before, prefer to use the generic comment about the optional method being required.

MSAL/src/native_auth/public/state_machine/delegate/SignInDelegates.swift
diegojerezba: Good catch, updated in every place it was happening

nilo-ms (approved):

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
```diff
@@ -52,7 +52,7 @@ public protocol SignUpPasswordStartDelegate {
 public protocol SignUpStartDelegate {
     /// Notifies the delegate that the operation resulted in an error.
     /// - Parameter error: An error object indicating why the operation failed.
-    @MainActor func onSignUpError(error: SignUpStartError)
+    @MainActor func onSignUpStartError(error: SignUpStartError)
 
     /// Notifies the delegate that a verification code is required from the user to continue.
     /// - Note: If a flow requires this optional method and it is not implemented, then ``onSignUpError(error:)`` will be called.
```
spetrescu84: Same here, `Start` word missing in function name in comment

MSAL/src/native_auth/public/state_machine/delegate/SignInDelegates.swift
spetrescu84: This is now solved

MSAL/src/native_auth/public/state_machine/delegate/SignInDelegates.swift
spetrescu84: This is now fixed

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
spetrescu84: This is now fixed

spetrescu84 (approved):

MSAL/test/unit/native_auth/public/delegate/DispatchAccessTokenRetrieveCompletedTests.swift
```diff
@@ -0,0 +1,83 @@
...
+        sut = .init(delegate: delegate, telemetryUpdate: { result in
+            guard case let .failure(error) = result, let customError = error as? RetrieveAccessTokenError else {
+                return XCTFail("wrong result")
+            }
+
+            checkError(customError)
+            self.telemetryExp.fulfill()
+        })
+
+        let expectedResult = MSALNativeAuthUserAccountResultStub.result
```
spetrescu84, line 71: This is not used

spetrescu84:
Updated End To End delegate spies to use the latest Delegates

PR #1923, opened 2023-11-20
Make errors extensible

diegojerezba:
## Proposed changes

This PR makes errors extensible by transforming them from `enum` (that are frozen in Swift) into classes.

Now the developer can do for every error:

- Check for utility variables (they are different for every error, but all of them start with "error.isXXX" (ex: `error.isBrowserRequired`)
- Check for `error.errorDescription`, which returns a `String?`. This is part of the `LocalizedError` protocol that all of our errors conform to.

### Swift developers can:

```swift
func onSignUpError(error: MSAL.SignUpStartError) {

  print(error.errorDescription)

  if error.isBrowserRequired {
    ...
  } else if error.isInvalidUsername {
    ...
  } else {
    ...
  }
}
```
### Objective-C developers can:

```objective-c
- (void)onSignInPasswordErrorWithError:(SignInPasswordStartError * _Nonnull)error {

  NSLog(@"%@", error.errorDescription);

  if (error.isBrowserRequired) {
      ...
  } else if (error.isUserNotFound) {
      ...
  }
}
```
## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
+++ b/MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
@@ -24,15 +24,14 @@
 
 import Foundation
 
-@objc
+@objcMembers
 public class MSALNativeAuthError: NSObject, LocalizedError {
+    /// Describes why an error occurred and provides more information about the error.
+    public var errorDescription: String? { message }
+
     private let message: String?
 
     init(message: String? = nil) {
         self.message = message
-    }
-
-    public var errorDescription: String? {
-        message
     }
 }
--- a/MSAL/src/native_auth/public/state_machine/error/PasswordRequiredError.swift
+++ b/MSAL/src/native_auth/public/state_machine/error/PasswordRequiredError.swift
@@ -24,12 +24,17 @@
 
 import Foundation
 
-@objc
+@objcMembers
 public class PasswordRequiredError: MSALNativeAuthError {
-    /// An error type indicating the type of error that occurred
-    @objc public let type: PasswordRequiredErrorType
+    enum ErrorType: CaseIterable {
+        case browserRequired
+        case invalidPassword
+        case generalError
+    }
 
-    init(type: PasswordRequiredErrorType, message: String? = nil) {
+    let type: ErrorType
+
+    init(type: ErrorType, message: String? = nil) {
         self.type = type
         super.init(message: message)
     }
@@ -38,7 +43,7 @@ public class PasswordRequiredError: MSALNativeAuthError {
         switch signInPasswordError.type {
         case .browserRequired:
             self.type = .browserRequired
-        case .invalidPassword:
+        case .invalidCredentials:
             self.type = .invalidPassword
         default:
             self.type = .generalError
@@ -46,6 +51,7 @@ public class PasswordRequiredError: MSALNativeAuthError {
         super.init(message: signInPasswordError.errorDescription)
     }
 
+    /// Describes why an error occurred and provides more information about the error.
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
@@ -53,18 +59,21 @@ public class PasswordRequiredError: MSALNativeAuthError {
 
         switch type {
         case .browserRequired:
-            return "Browser required"
+            return MSALNativeAuthErrorMessage.browserRequired
         case .invalidPassword:
-            return "Invalid password"
+            return MSALNativeAuthErrorMessage.invalidPassword
         case .generalError:
-            return "General error"
+            return MSALNativeAuthErrorMessage.generalError
         }
     }
-}
 
-@objc
-public enum PasswordRequiredErrorType: Int {
-    case browserRequired
-    case generalError
-    case invalidPassword
+    /// Returns `true` if a browser is required to continue the operation.
+    public var isBrowserRequired: Bool {
+        return type == .browserRequired
+    }
+
+    /// Returns `true` when the password is not valid.
+    public var isInvalidPassword: Bool {
+        return type == .invalidPassword
+    }
 }
--- a/MSAL/src/native_auth/public/state_machine/error/ResendCodeError.swift
+++ b/MSAL/src/native_auth/public/state_machine/error/ResendCodeError.swift
@@ -26,11 +26,12 @@ import Foundation
 
 @objc
 public class ResendCodeError: MSALNativeAuthError {
+    /// Describes why an error occurred and provides more information about the error.
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
         }
 
-        return "General error"
+        return MSALNativeAuthErrorMessage.generalError
     }
 }
--- a/MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
+++ b/MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
@@ -24,16 +24,24 @@
 
 import Foundation
 
-@objc
+@objcMembers
 public class ResetPasswordStartError: MSALNativeAuthError {
-    /// An error type indicating the type of error that occurred
-    @objc public let type: ResetPasswordStartErrorType
+    enum ErrorType: CaseIterable {
+        case browserRequired
+        case userDoesNotHavePassword
+        case userNotFound
+        case invalidUsername
+        case generalError
+    }
 
-    init(type: ResetPasswordStartErrorType, message: String? = nil) {
+    let type: ErrorType
+
+    init(type: ErrorType, message: String? = nil) {
         self.type = type
         super.init(message: message)
     }
 
+    /// Describes why an error occurred and provides more information about the error.
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
@@ -41,24 +49,35 @@ public class ResetPasswordStartError: MSALNativeAuthError {
 
         switch type {
         case .browserRequired:
-            return "Browser required"
-        case .generalError:
-            return "General error"
+            return MSALNativeAuthErrorMessage.browserRequired
         case .userDoesNotHavePassword:
-            return "User does not have a password"
+            return MSALNativeAuthErrorMessage.userDoesNotHavePassword
         case .userNotFound:
-            return "User not found"
+            return MSALNativeAuthErrorMessage.userNotFound
         case .invalidUsername:
-            return "Invalid username"
+            return MSALNativeAuthErrorMessage.invalidUsername
+        case .generalError:
+            return MSALNativeAuthErrorMessage.generalError
         }
     }
-}
 
-@objc
-public enum ResetPasswordStartErrorType: Int {
-    case browserRequired
-    case generalError
-    case userDoesNotHavePassword
-    case userNotFound
-    case invalidUsername
+    /// Returns `true` if a browser is required to continue the operation.
+    public var isBrowserRequired: Bool {
+        return type == .browserRequired
+    }
+
+    /// Returns `true` if the user does not have a password.
+    public var isUserDoesNotHavePassword: Bool {
+        return type == .userDoesNotHavePassword
+    }
+
+    /// Returns `true` if the user that is trying to reset their password cannot be found.
+    public var isUserNotFound: Bool {
+        return type == .userNotFound
+    }
+
+    /// Returns `true` when the username is not valid.
+    public var isInvalidUsername: Bool {
+        return type == .invalidUsername
+    }
 }
--- a/MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
+++ b/MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
@@ -24,38 +24,52 @@
 
 import Foundation
 
-@objc
+@objcMembers
 public class RetrieveAccessTokenError: MSALNativeAuthError {
-    /// An error type indicating the type of error that occurred
-    @objc public let type: RetrieveAccessTokenErrorType
+    enum ErrorType: CaseIterable {
+        case browserRequired
+        case refreshTokenExpired
+        case tokenNotFound
+        case generalError
+    }
 
-    init(type: RetrieveAccessTokenErrorType, message: String? = nil) {
+    let type: ErrorType
+
+    init(type: ErrorType, message: String? = nil) {
         self.type = type
         super.init(message: message)
     }
 
+    /// Describes why an error occurred and provides more information about the error.
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
         }
 
         switch type {
-        case .generalError:
-            return "General error"
-        case .refreshTokenExpired:
-            return "Refresh token expired"
-        case .tokenNotFound:
-            return "Token not found"
         case .browserRequired:
-            return "Browser required"
+            return MSALNativeAuthErrorMessage.browserRequired
+        case .refreshTokenExpired:
+            return MSALNativeAuthErrorMessage.refreshTokenExpired
+        case .tokenNotFound:
+            return MSALNativeAuthErrorMessage.tokenNotFound
+        case .generalError:
+            return MSALNativeAuthErrorMessage.generalError
         }
     }
-}
 
-@objc
-public enum RetrieveAccessTokenErrorType: Int {
-    case generalError
-    case refreshTokenExpired
-    case tokenNotFound
-    case browserRequired
+    /// Returns `true` if a browser is required to continue the operation.
+    public var isBrowserRequired: Bool {
+        return type == .browserRequired
+    }
+
+    /// Returns `true` if the refresh token has expired.
+    public var isRefreshTokenExpired: Bool {
+        return type == .refreshTokenExpired
+    }
+
+    /// Returns `true` if the existing token cannot be found.
+    public var isTokenNotFound: Bool {
+        return type == .tokenNotFound
+    }
 }
--- a/MSAL/src/native_auth/public/state_machine/error/SignInAfterSignUpError.swift
+++ b/MSAL/src/native_auth/public/state_machine/error/SignInAfterSignUpError.swift
@@ -26,11 +26,12 @@ import Foundation
 
 @objc
 public class SignInAfterSignUpError: MSALNativeAuthError {
+    /// Describes why an error occurred and provides more information about the error.
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
         }
 
-        return "General error"
+        return MSALNativeAuthErrorMessage.generalError
     }
 }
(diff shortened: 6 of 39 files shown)


MSAL/src/native_auth/public/state_machine/error/AttributesRequiredError.swift
```diff
@@ -26,11 +26,16 @@ import Foundation
 
 @objc
 public class AttributesRequiredError: MSALNativeAuthError {
+    /// Describes an error that provides messages describing why an error occurred and provides more information about the error.
```
rodhan-ms, line 29: This string doesn't read write. Maybe it should be just `/// Describes why an error occurred and provides more information about the error.`?

rodhan-ms: Same comment applies to many places because of the duplication.

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
```diff
@@ -26,13 +26,13 @@ import Foundation
 
 @objc
 public class MSALNativeAuthError: NSObject, LocalizedError {
+    public let identifier: String
```
rodhan-ms, line 29: Shouldn't there be a doc comment as this is a public member?

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
```diff
@@ -24,38 +24,39 @@
...
-        case .generalError:
-            return "General error"
-        case .refreshTokenExpired:
-            return "Refresh token expired"
-        case .tokenNotFound:
-            return "Token not found"
-        case .browserRequired:
-            return "Browser required"
-        }
+        return super.errorDescription ?? type.rawValue
```
rodhan-ms, line 38: I don't think we should be mixing errorDescription, which is intended to be a human-friendly brief description of the error, with the rawValue of the type, which should be a more computer-friendly, short, fixed string identifier for the type, e.g. `token_not_found` (or even possibly `tokenNotFound`).

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
```diff
@@ -24,38 +24,39 @@
...
 }
 
-@objc
-public enum RetrieveAccessTokenErrorType: Int {
-    case generalError
-    case refreshTokenExpired
-    case tokenNotFound
-    case browserRequired
+public enum RetrieveAccessTokenErrorType: String, CaseIterable {
+    case browserRequired = "Browser required"
```
rodhan-ms, line 58: Related to my previous comment, I think these string raw values for each case of the type should be a more fixed short computer-friendly string, e.g. `browser_required` that we know never ever to change. Having these strings as human-friendly short sentences is not as suitable for string comparison, using in `switch case` statements, etc.

MSAL/src/native_auth/public/state_machine/error/SignInAfterSignUpError.swift
```diff
@@ -26,11 +26,16 @@ import Foundation
...
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
         }
 
         return "General error"
     }
+
+    init(message: String? = nil) {
+        super.init(identifier: "Sign in after sign up", message: message)
```
rodhan-ms, line 39: I understand there is no need for a `type` member for some of the error subclasses like this one because there is only a single type, but for the sake of consistency I think it would be good to follow the same pattern for all error types.

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
rodhan-ms, line 29: Why not expose this to Objective-C? I understand as you mention in the PR description that you can only use it in a `switch case` in Swift but it might still be useful in Objective-C code.

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
spetrescu84: Indeed, we should have a distinction between `type` and `identifier`. Whichever is used for the switch should be computer friendly strings. Maybe have 2 enums one for type one for description/identifiers

MSAL/src/native_auth/public/state_machine/error/PasswordRequiredError.swift
```diff
@@ -43,28 +42,27 @@ public class PasswordRequiredError: MSALNativeAuthError {
...
-            return "Invalid password"
-        case .generalError:
-            return "General error"
-        }
+    /// Returns `true` if the error requires to use a browser.
+    public var isBrowserRequired: Bool {
+        return type == .browserRequired
+    }
+
+    /// Returns `true` when the password introduced is not valid.
```
nilo-ms, line 58: nit: what if you remove the word "introduced". It seems to make the comment more readable

MSAL/src/native_auth/public/state_machine/error/ResendCodeError.swift
```diff
@@ -26,11 +26,16 @@ import Foundation
...
     public override var errorDescription: String? {
         if let description = super.errorDescription {
             return description
         }
 
         return "General error"
     }
+
+    init(message: String? = nil) {
+        super.init(identifier: "Resend code", message: message)
```
nilo-ms, line 39: Shouldn't this string be "General error" because is the only type of error that this error class can has?
Anyway, I agree with Rodhan when he says that we should have a type in each class, even if they have one single type. So, we can be more consistent

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
```diff
@@ -24,41 +24,44 @@
...
-        }
+        return super.errorDescription ?? type.rawValue
+    }
+
+    /// Returns `true` if the error requires to use a browser.
+    public var isBrowserRequired: Bool {
+        return type == .browserRequired
+    }
+
+    public var isUserDoesNotHavePassword: Bool {
```
nilo-ms, line 46: did you leave this attribute without a comment on purpose?

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
```diff
@@ -24,41 +24,44 @@
...
 }
 
-@objc
-public enum ResetPasswordStartErrorType: Int {
-    case browserRequired
-    case generalError
-    case userDoesNotHavePassword
-    case userNotFound
-    case invalidUsername
+public enum ResetPasswordStartErrorType: String, CaseIterable {
```
nilo-ms, line 61: I suppose you didn't make this enum available to objective C, because we need to use Int as rawValue, right? We could use an additional attribute in that case, so we can use it from Objective-C and Swift

MSAL/test/unit/native_auth/public/error/SignInStartErrorTests.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    func test_defaultErrorDescription() {
+        sut = .init(type: .generalError)
+        XCTAssertEqual(sut.errorDescription, SignInStartErrorType.generalError.rawValue)
+    }
+
+
+    func test_isBrowserRequired() {
+        sut = .init(type: .browserRequired)
+
```
spetrescu84, line 65: nit: spacing, same for below tests, also order should be kept for where the `XCTAssertTrue` is like in the other tests classes such as `SignInPasswordStartErrorTests`

MSAL/test/unit/native_auth/public/error/SignUpPasswordStartErrorTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    func test_defaultErrorDescription() {
+        sut = .init(type: .generalError)
+        XCTAssertEqual(sut.errorDescription, SignUpPasswordStartErrorType.generalError.rawValue)
+    }
+
+    func test_isBrowserRequired() {
+        sut = .init(type: .browserRequired)
+
```
spetrescu84, line 66: nit: spacing, same for below tests, also order should be kept for where the `XCTAssertTrue` is like in the other tests classes such as `SignInPasswordStartErrorTests`

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
```diff
@@ -24,41 +24,44 @@
...
+    /// Returns `true` if the error requires to use a browser.
+    public var isBrowserRequired: Bool {
+        return type == .browserRequired
+    }
+
+    public var isUserDoesNotHavePassword: Bool {
+        return type == .userDoesNotHavePassword
+    }
+
+    /// Returns `true` if the user that is trying to sign in cannot be found.
```
nilo-ms, line 50: in this case the user is not trying to sign in

spetrescu84:
With this implementation we lose the `.generalError` case which is supposed to be more of a catch all rather than `unknownError`. What I'm saying is that the user doesn't check `error.isGeneralError` because if it's neither of the other ones it means it's a general one or an `unknown` one (as in not treated). 
Technically the last `else` or the `default:` case are either `.generalError` or unknown/untreated error. 
Making sure that's the intention of this design.

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
diegojerezba: For objective-c developers, they can do:

```objective-c
- (void)onSignInPasswordErrorWithError:(SignInPasswordStartError * _Nonnull)error {
  error.identifier
  error.errorDescription
  error.isBrowserRequired
  error.is...
}
```
Right now `identifier` is a String, so they can't use it to compare with `SignInPasswordStartErrorType`.

If we change `identifier` to be an Int, I think it would work. We can also rename it as `errorCode` if you like. But in this case it would be better to assign proper error codes like the ones present in `MSALError.h`

diegojerezba:
> With this implementation we lose the `.generalError` case which is supposed to be more of a catch all rather than `unknownError`. What I'm saying is that the user doesn't check `error.isGeneralError` because if it's neither of the other ones it means it's a general one or an `unknown` one (as in not treated). Technically the last `else` or the `default:` case are either `.generalError` or unknown/untreated error. Making sure that's the intention of this design.

Yes, that's the intention. Android is doing the same.

MSAL/src/native_auth/public/state_machine/error/AttributesRequiredError.swift
diegojerezba: Yeah, it was weird. Thanks for the feedback. Updated 👍.

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
diegojerezba: Yes, done.

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
diegojerezba: I think you are right, I've updated it. Check [this reply](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1923#discussion_r1401085666) to make it work with objc. I haven't tested it yet, but I think it could work.

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
diegojerezba: Agreed (as [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1923#discussion_r1401117647))

MSAL/src/native_auth/public/state_machine/error/SignInAfterSignUpError.swift
diegojerezba: Ok, added.

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
diegojerezba: I see all of you folks agree on this, I will expose it.

MSAL/src/native_auth/public/state_machine/error/PasswordRequiredError.swift
diegojerezba: Done. I've removed it from every comment, and I've done the same with the comments "Returns true when the username is not valid."

MSAL/src/native_auth/public/state_machine/error/ResendCodeError.swift
diegojerezba: Done (I've used `general_error` as identifier)

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
diegojerezba: I forgot to add it. Thanks.

MSAL/test/unit/native_auth/public/error/SignInStartErrorTests.swift
diegojerezba: Done.

MSAL/test/unit/native_auth/public/error/SignUpPasswordStartErrorTests.swift
diegojerezba: Done.

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
diegojerezba: Updated.

MSAL/src/native_auth/public/state_machine/error/ResetPasswordStartError.swift
nilo-ms: I agree on making the "identifier" variable an integer, so we could change all the error type enum (example: SignUpStartErrorType) rawValue to integer, so they can be exposed to ObjectiveC. I don't agree to rename the identifier variable to something else.

diegojerezba: Updated as discussed. I've updated the PR description to reflect these changes, too.

nilo-ms (approved):

MSAL/test/unit/native_auth/controllers/MSALNativeAuthSignInControllerTests.swift
```diff
@@ -393,7 +393,7 @@ final class MSALNativeAuthSignInControllerTests: MSALNativeAuthTestCase {
         helper.expectedCodeLength = expectedCodeLength
 
         let result = await sut.signIn(params: MSALNativeAuthSignInWithPasswordParameters(username: expectedUsername, password: expectedPassword, context: expectedContext, scopes: nil))
-        result.telemetryUpdate?(.failure(.init(message: "error")))
+        result.telemetryUpdate?(.failure(.init(identifier: 1, message: "error")))
```
spetrescu84, line 396: Should we use `SignInPasswordStartErrorType.browserRequired.rawValue` here to clarify what the identifier is? Same for the rest. Should we assign values like 0/1/2 to enums? [Apparently](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum) it should start with 0

spetrescu84:
As discussed we should add the word `Start` for the `StartDelegates`:
`SignUpPasswordStartDelegate` - `onSignUpPasswordError` should change to `onSignUpPasswordStartError`
`SignUpStartDelegate` - `onSignUpError` should change to `onSignUpStartError`
`SignInPasswordStartDelegate` - `onSignInPasswordError` should change to `onSignInPasswordStartError`
`SignInStartDelegate` - `onSignInError` should change to `onSignInStartError`
`ResetPasswordStartDelegate` - `onResetPasswordError` should change to `onResetPasswordStartError`

MSAL/test/unit/native_auth/controllers/MSALNativeAuthSignInControllerTests.swift
diegojerezba: Ok, done. This error is just a stub and the number doesn't matter here. But I agree that using the `rawValue` of the enum type is clearer.

diegojerezba:
> As discussed we should add the word `Start` for the `StartDelegates`: `SignUpPasswordStartDelegate` - `onSignUpPasswordError` should change to `onSignUpPasswordStartError` `SignUpStartDelegate` - `onSignUpError` should change to `onSignUpStartError` `SignInPasswordStartDelegate` - `onSignInPasswordError` should change to `onSignInPasswordStartError` `SignInStartDelegate` - `onSignInError` should change to `onSignInStartError` `ResetPasswordStartDelegate` - `onResetPasswordError` should change to `onResetPasswordStartError`

Thank you for putting this together, Silviu.

I'm thinking about doing this change in the [optional delegates PR](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1914), as it will avoid a few conflicts when we merge both PRs to dev.

MSAL/src/native_auth/public/state_machine/error/PasswordRequiredError.swift
```diff
@@ -24,47 +24,57 @@
...
         case .invalidPassword:
-            return "Invalid password"
+            return MSALNativeAuthErrorMessage.invalidPassword
         case .generalError:
-            return "General error"
+            return MSALNativeAuthErrorMessage.generalError
         }
     }
+
+    /// Returns `true` if the error requires to use a browser.
```
rodhan-ms, line 64: I think this should be more like "Returns `true` if a browser is required to continue the operation." or something

spetrescu84 (approved):

MSAL/src/native_auth/public/state_machine/error/SignUpPasswordStartError.swift
```diff
@@ -24,42 +24,60 @@
...
+    }
 
-@objc
-public enum SignUpPasswordStartErrorType: Int {
-    case browserRequired
-    case userAlreadyExists
-    case invalidPassword
-    case invalidUsername
-    case generalError
+    /// Returns `true` when the user is trying to register an existing email.
```
nilo-ms, line 69: should we use "username"  instead of "email"  here?

diegojerezba: Absolutely, updated now. Thank you

spetrescu84:
Integration tests need to be changed to use the new `error.is...` as they now use the `error.type`

spetrescu84:
> Integration tests need to be changed to use the new `error.is...` as they now use the `error.type`

Fixed integration tests

PR #1937, opened 2023-12-11
Don't send empty attribute list in SignUp/start

nilo-ms:
## Proposed changes

Before this PR we were using an empty attribute list as default value when the user didn't specify any attribute. This resulted in adding an empty list in the signUp/start HTTP call

## Type of change

- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information
I considered other methods for resolving this problem as well. I could just modify the `MSALNativeAuthSignUpRequestProvider` class and add the attributes field in the HTTP request only when the list was not empty. I didn't think this was the right way to solve the issue, because the RequestProvider class should provide the request based on the input parameters, not based to internal logic. So, I used a nil value instead of empty list as default attributes value in the PublicClientApplication class.

### DIFF
--- a/MSAL/src/native_auth/network/sign_up/MSALNativeAuthSignUpRequestProvider.swift
+++ b/MSAL/src/native_auth/network/sign_up/MSALNativeAuthSignUpRequestProvider.swift
@@ -42,14 +42,11 @@ final class MSALNativeAuthSignUpRequestProvider: MSALNativeAuthSignUpRequestProv
     }
 
     func start(parameters: MSALNativeAuthSignUpStartRequestProviderParameters) throws -> MSIDHttpRequest {
-        guard let attributes = try formatAttributes(parameters.attributes) else {
-            throw MSALNativeAuthInternalError.invalidAttributes
-        }
-
+        let formattedAttributes = try formatAttributes(parameters.attributes)
         let params = MSALNativeAuthSignUpStartRequestParameters(
             username: parameters.username,
             password: parameters.password,
-            attributes: attributes,
+            attributes: formattedAttributes,
             context: parameters.context
         )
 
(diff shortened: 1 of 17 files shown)


MSAL/src/native_auth/network/sign_up/MSALNativeAuthSignUpRequestProvider.swift
```diff
@@ -92,7 +89,10 @@ final class MSALNativeAuthSignUpRequestProvider: MSALNativeAuthSignUpRequestProv
         return request
     }
 
-    private func formatAttributes(_ attributes: [String: Any]) throws -> String? {
+    private func formatAttributes(_ attributes: [String: Any]?) throws -> String? {
+        guard let attributes = attributes else {
```
spetrescu84, line 93: Should we strip them also if the developer passes an empty dictionary? As in calls public client interface like this?
`signUpUsingPassword(username: "Test", password: "Pass", attributes: [:], delegate: self)` 
That would mean doing `guard let attributes = attributes, !attributes.isEmpty else {` here

MSAL/test/unit/native_auth/public/MSALNativeAuthPublicClientApplicationTest.swift
```diff
@@ -76,6 +76,7 @@ final class MSALNativeAuthPublicClientApplicationTest: XCTestCase {
         let delegate = SignUpPasswordStartDelegateSpy(expectation: exp)
         sut.signUpUsingPassword(username: "", password: "", delegate: delegate)
         wait(for: [exp], timeout: 1)
+        XCTAssertNil(controllerFactoryMock.signUpController.signUpStartRequestParameters?.attributes)
```
spetrescu84, line 79: Do you think these checks are needed for these specific tests? The test names are not describing anything around this. They don't hurt to be there but I think they are a bit overkill. We could have a test where we check that if we don't set the attributes they are not sent in this class but I think it should be separate.

MSAL/src/native_auth/network/sign_up/MSALNativeAuthSignUpRequestProvider.swift
nilo-ms: Hey Silviu, I took in consideration this approach. Please check the PR description to know more

MSAL/test/unit/native_auth/public/MSALNativeAuthPublicClientApplicationTest.swift
nilo-ms: Yeah, actually make more sense to have a dedicated test for this, so we can highlight this behaviour. Thanks I'll update this test class

PR #1940, opened 2023-12-13
Don't send empty attribute list in SignUp/start

nilo-ms:
## Proposed changes

Before this PR we were using an empty attribute list as default value when the user didn't specify any attribute. This resulted in adding an empty list in the signUp/start HTTP call

## Type of change

- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

I considered other methods for resolving this problem as well. I could just modify the `MSALNativeAuthSignUpRequestProvider` class and add the attributes field in the HTTP request only when the list was not empty. I didn't think this was the right way to solve the issue, because the `RequestProvider` class should provide the request based on the input parameters, not based to internal logic. So, I used a nil value instead of empty list as default attributes value in the `PublicClientApplication` class.

### DIFF
--- a/MSAL/test/unit/native_auth/public/MSALNativeAuthPublicClientApplicationTest.swift
+++ b/MSAL/test/unit/native_auth/public/MSALNativeAuthPublicClientApplicationTest.swift
@@ -95,6 +95,25 @@ final class MSALNativeAuthPublicClientApplicationTest: XCTestCase {
         XCTAssertEqual(delegate.channelTargetType, .email)
         XCTAssertEqual(delegate.codeLength, 1)
     }
+    
+    func testSignUpPassword_whenNoAttributesAreSpecified_AttributesShouldBeNil() {
+        let exp = expectation(description: "sign-up public interface")
+        let delegate = SignUpPasswordStartDelegateSpy(expectation: exp)
+
+        let expectedResult: SignUpPasswordStartResult = .codeRequired(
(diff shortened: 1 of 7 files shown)


MSAL/test/unit/native_auth/public/MSALNativeAuthPublicClientApplicationTest.swift
```diff
@@ -95,6 +95,25 @@ final class MSALNativeAuthPublicClientApplicationTest: XCTestCase {
...
+            sentTo: "sentTo",
+            channelTargetType: .email,
+            codeLength: 1
+        )
+        controllerFactoryMock.signUpController.startPasswordResult = .init(expectedResult)
+
+        sut.signUpUsingPassword(username: "correct", password: "correct", delegate: delegate)
+
+        wait(for: [exp])
+
```
spetrescu84, line 114: Maybe add a `XCTAssertNotNil(controllerFactoryMock.signUpController.signUpStartRequestParameters)` check so we make sure the it's not the `signUpStartRequestParameters` being nil (as in forgot to be set in the test which would be a false positive) ?
Same for below

nilo-ms: Yeah, make sense. Done ;)

spetrescu84 (approved):

PR #1924, opened 2023-11-24
[Replaced] Reuse MSAL initialised cache for Native Auth flows

rodhan-ms:
NOTE: This PR has been replaced by https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1958 as there were too many complex conflicts to resolve due to changes across the project.

## Proposed changes

MSALPublicClientApplication initialises the cache and keeps references to `tokenCache` and `accountMetadataCache` for the lifetime of the application. This PR updates the SDK to reuse this correctly configured cache for all Native Auth flows by modifying `MSALNativeAuthCacheAccessor` to no longer initialise the caches each time it needs them, and by modifying all controllers that need access to the cache to take a reference to the cacheAccessor during initialisation.

## Type of change

- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High – Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)
- [X] Medium – Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)
- [ ] Small – No issues are expected. (Example: Very small bug fixes, string changes, or configuration settings changes)

## Additional information

To verify this fix you need to use a non-default value for the `keychainSharingGroup`. This can by achived by modifying the Native Auth sample app to initialise NativeAuth using a MSALPublicApplicationConfig with non-default cache configuration, e.g.

```swift
            let ciamAuthority = try MSALCIAMAuthority(url: url)

            let configuration = MSALPublicClientApplicationConfig(
                clientId: Configuration.clientId,
                redirectUri: nil,
                authority: ciamAuthority
            )

            configuration.cacheConfig.keychainSharingGroup = Bundle.main.bundleIdentifier!

            nativeAuth = try MSALNativeAuthPublicClientApplication(
                configuration: configuration,
                challengeTypes: [.OOB])
```

PR #1958, opened 2024-01-05
Reuse MSAL initialised cache for Native Auth flows

rodhan-ms:
## Proposed changes

MSALPublicClientApplication initialises the cache and keeps references to `tokenCache` and `accountMetadataCache` for the lifetime of the application. This PR updates the SDK to reuse this correctly configured cache for all Native Auth flows by modifying `MSALNativeAuthCacheAccessor` to no longer initialise the caches each time it needs them, and by modifying all controllers that need access to the cache to take a reference to the cacheAccessor during initialisation.

## Type of change

- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

To verify this fix you need to use a non-default value for the `keychainSharingGroup`. This can by achived by modifying the Native Auth sample app to initialise NativeAuth using a MSALPublicApplicationConfig with non-default cache configuration, e.g.

```swift
            let ciamAuthority = try MSALCIAMAuthority(url: url)

            let configuration = MSALPublicClientApplicationConfig(
                clientId: Configuration.clientId,
                redirectUri: nil,
                authority: ciamAuthority
            )

            configuration.cacheConfig.keychainSharingGroup = Bundle.main.bundleIdentifier!

            nativeAuth = try MSALNativeAuthPublicClientApplication(
                configuration: configuration,
                challengeTypes: [.OOB])
```

This PR replaces the original https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1924 as there were too many complex conflicts to resolve due to changes across the project.  This PR also addresses the comments in the previous PR @diegojerezba.

### DIFF
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -108,14 +108,14 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
 
     init(
         controllerFactory: MSALNativeAuthControllerBuildable,
+        cacheAccessorFactory: MSALNativeAuthCacheAccessorBuildable,
         inputValidator: MSALNativeAuthInputValidating,
-        internalChallengeTypes: [MSALNativeAuthInternalChallengeType],
-        cacheAccessorFactory: MSALNativeAuthCacheAccessorBuildable
+        internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
     ) {
         self.controllerFactory = controllerFactory
+        self.cacheAccessorFactory = cacheAccessorFactory
         self.inputValidator = inputValidator
         self.internalChallengeTypes = internalChallengeTypes
-        self.cacheAccessorFactory = cacheAccessorFactory
 
         super.init()
     }
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -30,6 +30,9 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
     let controllerFactory: MSALNativeAuthControllerBuildable
     let inputValidator: MSALNativeAuthInputValidating
     private let internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
+    
+    lazy var cacheAccessor =
+        MSALNativeAuthCacheAccessor(tokenCache: self.tokenCache, accountMetadataCache: self.accountMetadataCache)
 
     /// Initialize a MSALNativePublicClientApplication with a given configuration and challenge types
     /// - Parameters:
@@ -226,7 +229,7 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
     /// - Parameter correlationId: Optional. UUID to correlate this request with the server for debugging.
     /// - Returns: An object representing the account information if present in the local cache.
     public func getNativeAuthUserAccount(correlationId: UUID? = nil) -> MSALNativeAuthUserAccountResult? {
-        let controller = controllerFactory.makeCredentialsController()
+        let controller = controllerFactory.makeCredentialsController(cacheAccessor: cacheAccessor)
         let context = MSALNativeAuthRequestContext(correlationId: correlationId)
 
         return controller.retrieveUserAccountResult(context: context)
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -30,9 +30,11 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
     let controllerFactory: MSALNativeAuthControllerBuildable
     let inputValidator: MSALNativeAuthInputValidating
     private let internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
-    
-    lazy var cacheAccessor =
-        MSALNativeAuthCacheAccessor(tokenCache: self.tokenCache, accountMetadataCache: self.accountMetadataCache)
+
+    private var cacheAccessorFactory: MSALNativeAuthCacheAccessorBuildable
+    lazy var cacheAccessor: MSALNativeAuthCacheAccessor = {
+        return cacheAccessorFactory.makeCacheAccessor(tokenCache: tokenCache, accountMetadataCache: accountMetadataCache)
+    }()
 
     /// Initialize a MSALNativePublicClientApplication with a given configuration and challenge types
     /// - Parameters:
@@ -57,6 +59,7 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
         nativeConfiguration.sliceConfig = config.sliceConfig
 
         self.controllerFactory = MSALNativeAuthControllerFactory(config: nativeConfiguration)
+        self.cacheAccessorFactory = MSALNativeAuthCacheAccessorFactory()
         self.inputValidator = MSALNativeAuthInputValidator()
 
         try super.init(configuration: config)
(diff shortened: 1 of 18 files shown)


MSAL/test/unit/native_auth/cache/MSALNativeAuthCacheAccessorTest.swift
```diff
@@ -27,7 +27,15 @@ import XCTest
 @_implementationOnly import MSAL_Private
 
 final class MSALNativeAuthCacheAccessorTest: XCTestCase {
-    private let cacheAccessor = MSALNativeAuthCacheAccessor()
+    private let tokenCache: MSIDDefaultTokenCacheAccessor = {
```
spetrescu84, line 30: Maybe add a test case on the `MSALNativeAuthPublicClientApplicationTest` class to test that the cache accessor is the same (in terms of memory pointer so same object) in the user account result as the one that was sent on SignUp.
I'm talking here about simulating a SignUp and then when the userAccount result comes back the cacheAccessor should be the same. 
It should be like the `func testSignUpPassword_correlationId_whenSetOnStart_itCascadesToAll() {` test just that at the end you could do a check like this:
`XCTAssertTrue(ObjectIdentifier(delegateSignInAfterSignUp.expectedUserAccountResult!.cacheAccessor as! MSALNativeAuthCacheAccessor) == ObjectIdentifier(cacheAccessor as! MSALNativeAuthCacheAccessor))`
Where `cacheAccessor` would be passed to the controllerFactoryMock.
Maybe there is a easier way to test this but this would cover the scenario

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -30,6 +30,9 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
     let controllerFactory: MSALNativeAuthControllerBuildable
     let inputValidator: MSALNativeAuthInputValidating
     private let internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
+    
+    lazy var cacheAccessor =
```
nilo-ms, line 34: If we want that only one instance of `MSALNativeAuthCacheAccessor` is created and used, why we don't use the Singleton pattern for this class? This can simplify the injection of the dependency

rodhan-ms: There could be multiple instances of the cache accessor at runtime if there are multiple instances of the `MSALNativeAuthPublicClientApplication` class and that is fine, so I don't see the use for a singleton here.
The reason we use lazy initialisation (unlike how `inputValidator` is initialised for example) is because the `tokenCache` and `accountMetadataCache` members are not valid until the `MSALPublicClientApplication` initialiser has been called (i.e. `super.init(...)`) so we can't do it in the regular initialisers. I'm happy to discuss more if that doesn't answer your question or you have an alternative.

nilo-ms: I'm wondering if makes sense to have more than one instance of this class, even if we have multiple instances of the `MSALNativeAuthPublicClientApplication` class. Given that multiple instantiations of this class created the issue that we're trying to solve in this PR, I'm asking if it is ok to have multiple instances of the cacheAccessor.

nilo-ms:
General question about this bug fix, I don't have anything against this solution, I'm only wondering why the issue happened in the first place. What I understood the original problem is when we initialise multiple instances of the `MSIDKeychainTokenCache` class, the first time the correct cache group is used, while the other times the default one is used. Did we understand why is this happening? Reusing the first instance created avoid having the undesirable behaviour, but have we solving the issue at its root? Maybe I'm missing something

nilo-ms:
General question about this bug fix, I don't have anything against this solution, I'm only wondering why the issue happened in the first place. What I understood the original problem is when we initialise multiple instances of the `MSIDKeychainTokenCache` class, the first time the correct cache group is used, while the other times the default one is used. Did we understand why is this happening? Reusing the first instance created avoid having the undesirable behaviour, but have we solving the issue at its root? Maybe I'm missing something

rodhan-ms:
> General question about this bug fix, I don't have anything against this solution, I'm only wondering why the issue happened in the first place. What I understood the original problem is when we initialise multiple instances of the `MSIDKeychainTokenCache` class, the first time the correct cache group is used, while the other times the default one is used. Did we understand why is this happening? Reusing the first instance created avoid having the undesirable behaviour, but have we solving the issue at its root? Maybe I'm missing something

There are more details in the bug (https://identitydivision.visualstudio.com/Engineering/_boards/board/t/DevExDub%20-B2C/Backlog%20items/?workitem=2757242). The underlying issue was that we were not taking user configuration into account when initialising the cache for Native Auth. The fix is that instead of creating multiple cache instances whenever we need them in our Native Auth code, we now reuse the cache that was initialised by MSALPublicClientApplication as we know that that will be configured correctly, etc. The problem wasn't caused by the multiple instances of the cache, that was just a symptom.

MSAL/test/unit/native_auth/cache/MSALNativeAuthCacheAccessorTest.swift
spetrescu84: As discussed there shouldn't be an End To End test to check this but rather make sure that the `cacheAccessor` uses the correct config

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
nilo-ms: After a chat we agreed that the singleton pattern here is not ideal. We're using the same approach of existing MSAL, that is: one cache accessor per public client application instance

nilo-ms (approved):

MSAL/test/unit/native_auth/cache/MSALNativeAuthCacheAccessorTest.swift
rodhan-ms: I added a test for this in MSALNativeAuthPublicClientApplicationTest (along with associated factories, etc.)

spetrescu84 (approved):

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -31,7 +31,7 @@ import Foundation
 
     let authTokens: MSALNativeAuthTokens
     let configuration: MSALNativeAuthConfiguration
-    private let cacheAccessor: MSALNativeAuthCacheInterface
+    let cacheAccessor: MSALNativeAuthCacheInterface
```
nilo-ms, line 34: why this property isn't private anymore?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -102,11 +109,13 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
     init(
         controllerFactory: MSALNativeAuthControllerBuildable,
         inputValidator: MSALNativeAuthInputValidating,
-        internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
+        internalChallengeTypes: [MSALNativeAuthInternalChallengeType],
```
nilo-ms, line 112: nit: would be good to have the challenge type list as last method parameter

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
rodhan-ms: It was because it was used in an "internal" method in an extension. I have updated the method to make it a parameter and put it back to being private.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
rodhan-ms: Good point thanks. I reordered the parameters

PR #1968, opened 2024-01-11
Add documentation needed for NativeAuth

spetrescu84:
## Proposed changes

- Added documentation for all public classes, methods, enums and properties
- Changed documentation in some places to match the standard (/** */ instead of /// on each line)
- `#import <Foundation/Foundation.h>` was added to `MSALNativeAuthChallengeTypes.h` file so the `build_docs.sh` doesn't fail
## Type of change

- [ ] Feature work
- [ ] Bug fix
- [x] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
```diff
@@ -24,8 +24,14 @@
 
 import Foundation
 
+/// The possible Channel Types via which an OOB was sent
+
```
diegojerezba, line 28: Nit: there's an extra space that in other classes doesn't exist

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
diegojerezba, line 1: This class still uses the `///` format instead of the `/** ... */` one.

In fact, I don't understand why the latter is considered "standard", can you provide some support for claim? I see Apple uses both (for example, see the inline documentation for `Swift.String` and `XCTest`).

In any case I'm fine with whatever format we decide, but I think we should be consistent across our classes.

MSAL/src/native_auth/public/state_machine/MSALNativeAuthRequiredAttributes.swift
```diff
@@ -24,11 +24,16 @@
...
+/// Class that defines the structure of a Required Attribute
 @objc
-public class MSALNativeAuthRequiredAttributes: NSObject {
+public class MSALNativeAuthRequiredAttribute: NSObject {
+    /// The name of the Required Attribute
     public let name: String
+    /// The type of the Required Attribute
     public let type: String
+    /// Specifies if the Attribute is required or not
     public let required: Bool
```
diegojerezba, line 35: Now that I'm reading the name of this class and its documentation... does this variable make sense in this class? It's supposed to always be `true`?

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
```diff
@@ -24,6 +24,7 @@
 
 import Foundation
 
+/// Class that defines the basic strucutre of a Native Auth error
```
diegojerezba, line 27: typo `strucutre`

MSAL/src/native_auth/public/state_machine/error/VerifyCodeError.swift
```diff
@@ -24,6 +24,7 @@
 
 import Foundation
 
+/// Class that defines the structure and type of a VerfiyCode error
```
diegojerezba, line 27: typo `VerfiyCode`

MSAL/src/native_auth/network/errors/sign_up/MSALNativeAuthRequiredAttributesInternal.swift
diegojerezba, line 1: The name of the file is still `MSALNativeAuthRequiredAttributesInternal` (plural Attributes)

MSAL/src/native_auth/public/state_machine/MSALNativeAuthRequiredAttributes.swift
diegojerezba, line 1: The name of the file is still `MSALNativeAuthRequiredAttributes` (plural Attributes)

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
spetrescu84: By standard I meant in our SDK, and the "standard" is coming from the MSAL Objective C classes.
In my opinion we should use the following format:
-If the comment is multiline and is used to describe a class we should use /** */ (again to match Objective C)
-If the comment is single line or is inside the class it should use ///
Indeed if you look at MSAL Objective C comments with a few exceptions they use /** */
If we want to be aligned we should change all comments to /** */
If we don't want to aligned but rather homogenous in our SDK we should change all comments to /// regardless of single or multi-line
Example of MSAL Objective C comment:
Screenshot 2024-01-12 at 11 23 24

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
spetrescu84: Removed

MSAL/src/native_auth/public/state_machine/MSALNativeAuthRequiredAttributes.swift
spetrescu84: Asked APi team

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
spetrescu84: Corrected

MSAL/src/native_auth/public/state_machine/error/VerifyCodeError.swift
spetrescu84: Corrected

MSAL/src/native_auth/network/errors/sign_up/MSALNativeAuthRequiredAttributesInternal.swift
spetrescu84: Fixed

MSAL/src/native_auth/public/state_machine/MSALNativeAuthRequiredAttributes.swift
spetrescu84: Fixed

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
```diff
@@ -25,13 +25,21 @@
...
+
+/** 
+    The set of capabilities that an application wishes to support for Native Auth operations.
+
+    Valid options are:
+    * OOB: The application can support asking a user to supply a verification code that is sent by email.
+    * Password: The application can support asking a user to supply a password
+*/
+
 typedef NS_OPTIONS(NSInteger, MSALNativeAuthChallengeTypes) {
```
nilo-ms, line 38: Now that I'm thinking, shouldn't this option be named as `MSALNativeAuthChallengeType` instead of `MSALNativeAuthChallengeType*S*` ?

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
```diff
@@ -24,8 +24,13 @@
 
 import Foundation
 
+/// The possible Channel Types via which an OOB was sent
```
nilo-ms, line 27: can you use "code" instead of OOB?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -24,6 +24,28 @@
 
 import Foundation
 
+/**
+    Subclasses MSALPublicClientApplication to be used for Native Auth operations
```
nilo-ms, line 28: Why should we specify that MSALNativeAuthPublicClientApplication is a subclass of MSALPublicClientApplication? I think we shouldn't say that, it's an internal technical detail

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -24,6 +24,28 @@
...
+    
+    To create an instance of the MSALNativeAuthPublicClientApplication pass directly the clientId, tenantSubdomain, challengeTypes and redirectUri (if needed)
```
nilo-ms, line 30: What if we say:
```suggestion
    To create an instance of the MSALNativeAuthPublicClientApplication use the clientId, tenantSubdomain, challengeTypes and redirectUri (if needed)
```

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -24,6 +24,28 @@
...
+    to init(clientId: tenantSubdomain: challengeTypes: redirectUri)
```
nilo-ms, line 31:
```suggestion
    to the initialiser method.
```

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -24,6 +24,28 @@
...
+    For example:
+ 
+     <pre>
+         do {
+             nativeAuth = try MSALNativeAuthPublicClientApplication(
+                 clientId: "Enter_the_Application_Id_Here",
+                 tenantSubdomain: "Enter_the_Tenant_Subdomain_Here",
+                 challengeTypes: [.OOB]
+             )
+             print("Initialised Native Auth successfully.")
```
nilo-ms, line 42: Should we use "MSAL Native Auth" instead?

spetrescu84: This example was taken from [the documentation](https://github.com/microsoft/entra-previews/blob/PP4/docs/Native-Auth/Developer-guides/1-iOS-Swift/1-Add-authentication-prepare-app.md#create-sdk-instance) so it should stay in sync but can change if needed.

MSAL/src/native_auth/public/state_machine/MSALNativeAuthRequiredAttribute.swift
```diff
@@ -24,11 +24,16 @@
...
 
+/// Class that defines the structure of a Required Attribute
 @objc
-public class MSALNativeAuthRequiredAttributes: NSObject {
+public class MSALNativeAuthRequiredAttribute: NSObject {
+    /// The name of the Required Attribute
     public let name: String
+    /// The type of the Required Attribute
     public let type: String
+    /// Specifies if the Attribute is required or not
```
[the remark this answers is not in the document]
spetrescu84: Here I took the approach that "Required Attribute" is the name of the class and that's why I capitalised Attribute but I was also on the edge about it. Renamed. This whole property should not exist in the first place in my opinion as it contradicts the class name

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
spetrescu84: Changed as I agree it's more clean.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
spetrescu84: Changed as it's more direct

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
spetrescu84: This is based on the discussion with Rodhan but indeed looking at Android one, we should have something more aligned so I changed it - https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/nativeauth/INativeAuthPublicClientApplication.kt

MSAL/src/native_auth/public/state_machine/delegate/SignUpDelegates.swift
diegojerezba: Thanks for the clarification. But with this approach, we are following the MSAL standard only on larger comments, not on the one line comments. Shouldn't we follow the standard everywhere?

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
spetrescu84: Done

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
spetrescu84: That is actually a Set (NS_OPTIONS) so you can pass into this init:
```
public init(
        configuration config: MSALPublicClientApplicationConfig,
        challengeTypes: MSALNativeAuthChallengeTypes) throws {
```
A set, like so:
`MSALNativeAuthPublicClientApplication(configuration: config, challengeTypes: [.OOB, .password])`
and even if it's one, it's like this:
`MSALNativeAuthPublicClientApplication(configuration: config, challengeTypes: [.OOB])`
which still signifies a Set/Array as in multiple items

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
```diff
@@ -24,8 +24,13 @@
 
 import Foundation
 
+/// The possible Channel Types via which a Code was sent
 @objc
 public enum MSALNativeAuthChannelType: Int {
+    
+    /// Specifies if the channel type is Email
```
rodhan-ms, line 31: I don't see any value in this type of documentation/comment. It is really just stating the obvious isn't it?

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
nilo-ms: Thanks for the clarification Silviu. All good

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
```diff
@@ -24,8 +24,13 @@
 
 import Foundation
 
+/// The possible Channel Types via which a Code was sent
```
nilo-ms, line 27: Why the word "code" is capital?

MSAL/src/native_auth/public/state_machine/error/VerifyCodeError.swift
```diff
@@ -24,6 +24,7 @@
 
 import Foundation
 
+/// Class that defines the structure and type of a VerifyCodeE error
```
rodhan-ms, line 27: Typo in `VerifyCodeE`

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
```diff
@@ -25,13 +25,21 @@
...
 #define MSALNativeAuthChallengeTypes_h
 
-/// The set of capabilities that an application wishes to support for Native Auth operations.
-///
-/// Valid options are:
-/// * OOB: The application can support asking a user to supply a verification code that is sent by email.
-/// * Password: The application can support asking a user to supply a password
+#import <Foundation/Foundation.h>
+
+/**
```
rodhan-ms, line 30: As discussed in Teams, this should change back to `///` style comment

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
```diff
@@ -25,13 +25,21 @@
 #ifndef MSALNativeAuthChallengeTypes_h
 #define MSALNativeAuthChallengeTypes_h
 
-/// The set of capabilities that an application wishes to support for Native Auth operations.
-///
-/// Valid options are:
-/// * OOB: The application can support asking a user to supply a verification code that is sent by email.
-/// * Password: The application can support asking a user to supply a password
+#import <Foundation/Foundation.h>
```
rodhan-ms, line 28: Why do we have to add this import?

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
spetrescu84: Indeed it's obvious and there has to be something there otherwise it says undocumented. I can add anything but it has to be something

rodhan-ms: Thanks for the explanation. That's annoying but I understand.

MSAL/src/native_auth/network/errors/sign_up/MSALNativeAuthRequiredAttributesOptions.swift
```diff
@@ -24,6 +24,6 @@
 
 import Foundation
 
-class MSALNativeAuthRequiredAttributeOptions: Decodable {
+class MSALNativeAuthRequiredAttributesOptions: Decodable {
```
rodhan-ms, line 27: This change shouldn't be in the PR now. Also the PR Description still refers to the Attributes change.

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
spetrescu84: The import is needed because MSALNativeAuthChallengeTypes is imported in `MSAL.h` with `#import <MSAL/MSALLegacySharedAccountsProvider.h>` 
If the import is not added here, it will fail creating the documentation when `sourcekitten` is run because `NS_OPTIONS` is not found

MSAL/src/native_auth/public/MSALNativeAuthChannelType.swift
spetrescu84: it shouldn't be :) changed

MSAL/src/native_auth/public/state_machine/error/VerifyCodeError.swift
spetrescu84: Thanks for catching it! Fixed

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
spetrescu84: Yes, updated now

MSAL/src/native_auth/network/errors/sign_up/MSALNativeAuthRequiredAttributesOptions.swift
spetrescu84: I've only now removed it.

MSAL/src/native_auth/public/MSALNativeAuthChallengeTypes.h
rodhan-ms: Got it. Makes sense. Thanks.

rodhan-ms (approved):

MSAL/src/native_auth/public/state_machine/MSALNativeAuthRequiredAttributes.swift
diegojerezba: Let's follow up with them, but I'm resolving this to avoid blocking the PR

diegojerezba (approved):

nilo-ms (approved):

PR #1969, opened 2024-01-11
Include Swift code into reference doc

spetrescu84:
## Proposed changes

- Changed build_docs.sh to add Swift code to reference doc using https://github.com/realm/jazzy#mixed-objective-c--swift

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/build_docs.sh
+++ b/build_docs.sh
@@ -2,8 +2,8 @@
 # To build manually run "sudo sh build_docs.sh"
 gem install jazzy
 echo -e "Copying MSAL public files"
-mkdir -m 777 docs.temp
-mkdir -m 777 docs.temp/MSAL
+mkdir -m docs.temp
+mkdir -m docs.temp/MSAL
 cp `find MSAL/src/public` docs.temp/MSAL
 cp `find MSAL/src/native_auth/public` docs.temp/MSAL
 cp README.md docs.temp/
--- a/build_docs.sh
+++ b/build_docs.sh
@@ -5,8 +5,8 @@
 # details see https://github.com/realm/jazzy?tab=readme-ov-file#mixed-objective-c--swift
 gem install jazzy
 echo -e "Copying MSAL public files"
-mkdir docs.temp
-mkdir docs.temp/MSAL
+mkdir -m 777 docs.temp
+mkdir -m 777 docs.temp/MSAL
 cp `find MSAL/src/public` docs.temp/MSAL
 cp `find MSAL/src/native_auth/public` docs.temp/MSAL
 cp README.md docs.temp/
--- a/build_docs.sh
+++ b/build_docs.sh
@@ -1,11 +1,21 @@
 #!/bin/bash
+# To build manually run "sudo sh build_docs.sh"
 gem install jazzy
 echo -e "Copying MSAL public files"
-mkdir docs.temp
-mkdir docs.temp/MSAL
+mkdir -m 777 docs.temp
+mkdir -m 777 docs.temp/MSAL
 cp `find MSAL/src/public` docs.temp/MSAL
+cp `find MSAL/src/native_auth/public` docs.temp/MSAL
 cp README.md docs.temp/
-cd docs.temp
-echo -e "Generating MSAL documentation"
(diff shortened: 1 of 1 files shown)


build_docs.sh
```diff
@@ -1,11 +1,21 @@
 #!/bin/bash
+# To build manually run "sudo sh build_docs.sh"
 gem install jazzy
 echo -e "Copying MSAL public files"
-mkdir docs.temp
-mkdir docs.temp/MSAL
+mkdir -m 777 docs.temp
```
nilo-ms, line 5: Why did we change the permission of this file? Shouldn't we use the default permission instead of 777?

build_docs.sh
```diff
@@ -1,11 +1,21 @@
...
+mkdir -m 777 docs.temp/MSAL
 cp `find MSAL/src/public` docs.temp/MSAL
+cp `find MSAL/src/native_auth/public` docs.temp/MSAL
 cp README.md docs.temp/
-cd docs.temp
+
 echo -e "Generating MSAL documentation"
-jazzy --objc --umbrella-header MSAL/MSAL.h --framework-root . --sdk iphonesimulator --author Microsoft\ Corporation --author_url https://aka.ms/azuread  --github_url https://github.com/AzureAD/microsoft-authentication-library-for-objc --theme fullwidth
+# Generate Swift SourceKitten output
+sourcekitten doc -- -workspace MSAL.xcworkspace -scheme "MSAL (iOS Framework)" -configuration Debug RUN_CLANG_STATIC_ANALYZER=NO -sdk iphonesimulator CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO -destination 'platform=iOS Simulator,name=iPhone 14,OS=16.4' > docs.temp/swiftDoc.json
```
nilo-ms, line 13: are the "--" needed here?

build_docs.sh
spetrescu84: Sorry, I should have copied over this comments from ADO:
While running locally, the 777 permission needs to be added otherwise sourcekitten command fails because it cannot write into the json's. It could be on server side the creation of the folder works differently as the terminal might be started with different parameters. The 777 permission gives access to all users or scripts to write into the folder/files in folder.

build_docs.sh
spetrescu84: The lines are needed otherwise the following error happens:
```
Error: Unknown option '-x'
Usage: sourcekitten doc [--single-file] [--module-name <module-name>] [--spm] [--objc] [<arguments> ...]
  See 'sourcekitten doc --help' for more information.
/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/json/common.rb:156:in `parse': 416: unexpected token at ']' (JSON::ParserError)
```

build_docs.sh
nilo-ms: Can you please check if we need this permission also in the server side?

build_docs.sh
rodhan-ms, line 13: Would there be a way to avoid hardcoding the exact version and device details in the destination here but instead use latest or something? We don't want this to break in the future when minimum versions, etc. change.

build_docs.sh
spetrescu84: It doesn't need to be added so I removed it. Will check before deployment to make sure script runs correctly.

build_docs.sh
spetrescu84: if the name is supplied it has to be exact. But it's actually not required so I removed it. By default the first simulator installed will be used and since it's only building to create the documentation that will pass regardless of which it is.

build_docs.sh
```diff
@@ -1,11 +1,21 @@
 #!/bin/bash
+# To build manually run "sudo sh build_docs.sh"
```
rodhan-ms, line 2: As this is a seldom used file I think it would be good to add some high level explanation about the SourceKitten and Jazzy interaction, e.g.

```
As MSAL is a mixed Objective-C and Swift project we need to use SourceKitten to
first extract the doc comments from the source before building with Jazzy. For more
details see https://github.com/realm/jazzy?tab=readme-ov-file#mixed-objective-c--swift
```

spetrescu84: added

PR #1970, opened 2024-01-11
Print warning on console if URL scheme is not configured

rodhan-ms:
## Proposed changes

Updated public initializers of MSALNativeAuthPublicClientApplication to output a message when redirectUri is not set.

Note: To actually see this warning message the developer will need to have the MSAL logging callback configured correctly as [described here](https://learn.microsoft.com/en-us/entra/identity-platform/msal-logging-ios).

The message can be updated to point to a more relevant documentation page in the future (GA timeline?)

To test: Use this version of the library in the default sample app.  If a redirectUri is configured correctly the message will not be output.

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

https://identitydivision.visualstudio.com/Engineering/_boards/board/t/DevExDub%20-B2C/Backlog%20items/?workitem=2649566

### DIFF
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -83,7 +83,7 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
         self.inputValidator = MSALNativeAuthInputValidator()
 
         if config.redirectUri == nil {
-            print(MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
+            MSALLogger.log(level: .warning, context: nil, format: MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
         }
 
         try super.init(configuration: config)
@@ -122,7 +122,7 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
         )
 
         if redirectUri == nil {
-            print(MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
+            MSALLogger.log(level: .warning, context: nil, format: MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
         }
 
         // we need to bypass redirect URI validation because we don't need a redirect URI for Native Auth scenarios
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -82,6 +82,10 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
         self.cacheAccessorFactory = MSALNativeAuthCacheAccessorFactory()
         self.inputValidator = MSALNativeAuthInputValidator()
 
+        if config.redirectUri == nil {
+            print(MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
+        }
+
         try super.init(configuration: config)
     }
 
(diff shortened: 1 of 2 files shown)


diegojerezba:
Why using `print` instead of the MSAL logger? `print` statements are not removed from the prod builds AFAIK. My opinion is that every log produced by the SDK should be done via MSAL logger using the appropriate log level.

rodhan-ms:
> Why using `print` instead of the MSAL logger? `print` statements are not removed from the prod builds AFAIK. My opinion is that every log produced by the SDK should be done via MSAL logger using the appropriate log level.

I mentioned that in the description that we use `print` instead of using the MSAL logger on purpose. The add more detail it is because this is a message that we want the developer to see even in non-debug builds. This is not really part of the MSAL logs which are more of a debugging tool. This is a notice/warning to developers to remind them while they are developing an app using the SDK that they will need to add a redirectUri before they finish the code and ship an app.

spetrescu84 (approved):

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -56,6 +56,10 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
         self.controllerFactory = MSALNativeAuthControllerFactory(config: nativeConfiguration)
         self.inputValidator = MSALNativeAuthInputValidator()
 
+        if config.redirectUri == nil {
+            print(MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
```
nilo-ms, line 60: @rodhan-ms , continuing here the conversation started by @diegojerezba , could you please get advice from Sergei for this specific case? And what MSAL Android did it use for the same log?
MSAL team never used NSLog, they have this rule: "You should only ever log to NSLog in debug mode to prevent leaking potentially sensitive information". But I'm not sure if we need to threat this case as a special case.

rodhan-ms: Hi @antrix1989 can we get your input on this please? (See also https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1970#issuecomment-1889020399).

In short, basic Native Auth flows don't require the redirectUri to be set, especially when developers are taking their first steps with the SDK. However, as previously discussed, we want to make it clear to developers that they will need to add a redirect URI and implement the web fallback cases before their apps are in production. We agreed that we would log a warning to remind developers about this. The question now is where to put that log - should we use the regular MSAL log mechanism or should we make sure it shows in the Xcode console by using `print` directly.

antrix1989: I suggest using the MSAL logger for all log messages. NSLog can significantly slow down performance.

> they will need to add a redirect URI and implement the web fallback cases before their apps are in production

If they have to do it eventually anyway, does it really matter whether it's done at the beginning or the end of their project? Should we simply add an assertion() that will fail in debug mode and indicate that a redirect URI is required?

rodhan-ms: Thanks @antrix1989, I'll update the code to use the MSAL logger as suggested. 
The reason for logging and not, e.g. asserting is that we want to make early experimentation with the SDK as quick and easy as possible for developers. Thanks for your input.

nilo-ms: Thanks @antrix1989 , we wrote a document to describe the native auth approach at the web fallback (and redirect URI). https://github.com/microsoft/entra-previews/blob/PP4/docs/Native-Auth/Developer-guides/1-iOS-Swift/3-Additional-scenarios/7-support-web-fallback.md
We don't force the user to implement a fallback mechanism, but we inform of the issue that they can have if they don't do it.

nilo-ms (approved):

PR #1981, opened 2024-01-24
Read correlationId from server on 2xx and 4xx responses. Add correlationId and errorCodes to public errors.

diegojerezba:
## Proposed changes

- Read correlationId from the headers of every 2xx and 4xx response.
- Add correlationId and errorCodes to the public errors in sign-up, sign-in, reset-password and credentials-token flows.

Note: The MSAL-ObjC-CIAM team has been added because I've made a made an addition to the `MSAL/module.modulemap` file, but the rest of the changes don't affect MSAL objc.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
+++ b/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
@@ -155,11 +155,12 @@ class MSALNativeAuthBaseController {
     ) async -> Result<T, Error> {
         return await withCheckedContinuation { continuation in
             request.send { result, error in
-                if let error = error, let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
-                    context.setServerCorrelationId(errorWithCorrelationId.correlationId)
-                    continuation.resume(returning: .failure(error))
-                } else if let error = error {
-                    MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
+                if let error = error {
+                    if let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
+                        context.setServerCorrelationId(errorWithCorrelationId.correlationId)
+                    } else {
+                        MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
+                    }
                     continuation.resume(returning: .failure(error))
                 } else if let response = result as? T {
                     context.setServerCorrelationId(response.correlationId)
--- a/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
+++ b/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
@@ -149,12 +149,20 @@ class MSALNativeAuthBaseController {
         completion(response, error)
     }
 
-    func performRequest<T>(_ request: MSIDHttpRequest, context: MSIDRequestContext) async -> Result<T, Error> {
+    func performRequest<T: MSALNativeAuthResponseCorrelatable>(
+        _ request: MSIDHttpRequest,
+        context: MSALNativeAuthRequestContext
+    ) async -> Result<T, Error> {
         return await withCheckedContinuation { continuation in
             request.send { result, error in
-                if let error = error {
+                if let error = error, let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
+                    context.setServerCorrelationId(errorWithCorrelationId.correlationId)
+                    continuation.resume(returning: .failure(error))
+                } else if let error = error {
+                    MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
                     continuation.resume(returning: .failure(error))
                 } else if let response = result as? T {
+                    context.setServerCorrelationId(response.correlationId)
                     continuation.resume(returning: .success(response))
                 } else {
                     MSALLogger.log(level: .error, context: context, format: "Error request - Both result and error are nil")
--- a/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
+++ b/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
@@ -154,13 +154,18 @@ class MSALNativeAuthBaseController {
         context: MSALNativeAuthRequestContext
     ) async -> Result<T, Error> {
         return await withCheckedContinuation { continuation in
-            request.send { result, error in
+            request.send { [weak self] result, error in
                 if let error = error {
-                    if let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
+                    if let correlationId = self?.extractCorrelationIdFromUserInfo((error as NSError).userInfo) {
+                        context.setServerCorrelationId(UUID(uuidString: correlationId))
+                    } else if let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
                         context.setServerCorrelationId(errorWithCorrelationId.correlationId)
+                    } else if case MSALNativeAuthInternalError.responseSerializationError(let correlationId) = error {
+                        context.setServerCorrelationId(correlationId)
                     } else {
                         MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
                     }
+
                     continuation.resume(returning: .failure(error))
                 } else if let response = result as? T {
                     context.setServerCorrelationId(response.correlationId)
@@ -171,5 +176,16 @@ class MSALNativeAuthBaseController {
                 }
             }
         }
+    }
+
+    private func extractCorrelationIdFromUserInfo(_ userInfo: [String: Any]) -> String? {
+        guard
+            let headers = userInfo[MSIDHTTPHeadersKey] as? [String: Any],
+            let correlationId = headers[MSID_OAUTH2_CORRELATION_ID_REQUEST_VALUE] as? String
+        else {
+            return nil
+        }
+
+        return correlationId
     }
 }
--- a/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
+++ b/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
@@ -156,13 +156,19 @@ class MSALNativeAuthBaseController {
         return await withCheckedContinuation { continuation in
             request.send { [weak self] result, error in
                 if let error = error {
+                    // 5xx errors contain the server's returned correlation-id in userInfo.
                     if let correlationId = self?.extractCorrelationIdFromUserInfo((error as NSError).userInfo) {
                         context.setServerCorrelationId(UUID(uuidString: correlationId))
+
+                    // 4xx errors are decoded producing an error that conforms to MSALNativeAuthResponseCorrelatable protocol.
                     } else if let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
                         context.setServerCorrelationId(errorWithCorrelationId.correlationId)
+
+                    // If a 4xx error fails to decode, this error is returned from the error deserializer.
                     } else if case MSALNativeAuthInternalError.responseSerializationError(let correlationId) = error {
                         context.setServerCorrelationId(correlationId)
                     } else {
+                        context.setServerCorrelationId(nil)
                         MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
                     }
 
--- a/MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
+++ b/MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
@@ -70,7 +70,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
         return await handleSignUpStartResult(result, username: parameters.username, event: event, context: parameters.context)
     }
 
-    func resendCode(username: String, context: MSIDRequestContext, continuationToken: String) async -> SignUpResendCodeControllerResponse {
+    func resendCode(username: String, context: MSALNativeAuthRequestContext, continuationToken: String) async -> SignUpResendCodeControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdSignUpResendCode, context: context)
         let challengeResult = await performAndValidateChallengeRequest(continuationToken: continuationToken, context: context)
         return handleResendCodeResult(challengeResult, username: username, event: event, continuationToken: continuationToken, context: context)
@@ -80,7 +80,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
         _ code: String,
         username: String,
         continuationToken: String,
-        context: MSIDRequestContext
+        context: MSALNativeAuthRequestContext
     ) async -> SignUpSubmitCodeControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdSignUpSubmitCode, context: context)
         let params = MSALNativeAuthSignUpContinueRequestProviderParams(
@@ -98,7 +98,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
         _ password: String,
         username: String,
         continuationToken: String,
-        context: MSIDRequestContext
+        context: MSALNativeAuthRequestContext
     ) async -> SignUpSubmitPasswordControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdSignUpSubmitPassword, context: context)
 
@@ -122,7 +122,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
         _ attributes: [String: Any],
         username: String,
         continuationToken: String,
-        context: MSIDRequestContext
+        context: MSALNativeAuthRequestContext
     ) async -> SignUpSubmitAttributesControllerResponse {
         let event = makeAndStartTelemetryEvent(id: .telemetryApiIdSignUpSubmitAttributes, context: context)
         let params = MSALNativeAuthSignUpContinueRequestProviderParams(
@@ -147,7 +147,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
             request = try requestProvider.start(parameters: parameters)
         } catch {
             MSALLogger.log(level: .error, context: parameters.context, format: "Error while creating Start Request: \(error)")
-            return .unexpectedError(message: nil)
+            return .unexpectedError(nil)
         }
 
         MSALLogger.log(level: .info, context: parameters.context, format: "Performing signup/start request")
@@ -161,55 +161,61 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
         _ result: MSALNativeAuthSignUpStartValidatedResponse,
         username: String,
         event: MSIDTelemetryAPIEvent?,
-        context: MSIDRequestContext
+        context: MSALNativeAuthRequestContext
     ) async -> SignUpStartControllerResponse {
         switch result {
         case .success(let continuationToken):
             MSALLogger.log(level: .info, context: context, format: "Successful signup/start request")
             let challengeResult = await performAndValidateChallengeRequest(continuationToken: continuationToken, context: context)
             return handleSignUpChallengeResult(challengeResult, username: username, event: event, context: context)
-        case .attributeValidationFailed(let invalidAttributes):
+        case .attributeValidationFailed(let apiError, let invalidAttributes):
             MSALLogger.log(
                 level: .error,
                 context: context,
                 format: "attribute_validation_failed received from signup/start request for attributes: \(invalidAttributes)"
             )
             let message = String(format: MSALNativeAuthErrorMessage.attributeValidationFailedSignUpStart, invalidAttributes.description)
-            let error = SignUpStartError(type: .generalError, message: message)
+            let error = apiError.toSignUpStartPublicError(context: context, message: message)
             return .init(.attributesInvalid(invalidAttributes), telemetryUpdate: { [weak self] result in
                 // The telemetry event always fails because the attribute validation failed
                 self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result, controllerError: error)
             })
         case .redirect:
-            let error = SignUpStartError(type: .browserRequired)
+            let error = SignUpStartError(type: .browserRequired, correlationId: context.correlationId())
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Redirect error in signup/start request \(error.errorDescription ?? "No error description")")
             return .init(.error(error))
-        case .error(let apiError):
-            let error = apiError.toSignUpStartPublicError()
+        case .error(let apiError),
+             .unauthorizedClient(let apiError):
+            let error = apiError.toSignUpStartPublicError(context: context)
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in signup/start request \(error.errorDescription ?? "No error description")")
             return .init(.error(error))
         case .invalidUsername(let apiError):
-            let error = SignUpStartError(type: .invalidUsername, message: apiError.errorDescription)
+            let error = SignUpStartError(
+                type: .invalidUsername,
+                message: apiError.errorDescription,
+                correlationId: context.correlationId(),
+                errorCodes: apiError.errorCodes ?? [],
+                errorUri: apiError.errorURI
+            )
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "InvalidUsername in signup/start request \(error.errorDescription ?? "No error description")")
             return .init(.error(error))
-        case .unauthorizedClient(let apiError):
-            let error = SignUpStartError(type: .generalError, message: apiError.errorDescription)
-            stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Invalid Client Id in signup/start request \(error.errorDescription ?? "No error description")")
-            return .init(.error(error))
-        case .unexpectedError(let message):
-            let error = SignUpStartError(type: .generalError, message: message)
+        case .unexpectedError(let apiError):
+            let error = SignUpStartError(
+                type: .generalError,
+                message: apiError?.errorDescription,
+                correlationId: context.correlationId(),
+                errorCodes: apiError?.errorCodes ?? [],
+                errorUri: apiError?.errorURI
+            )
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
@@ -222,7 +228,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
 
     private func performAndValidateChallengeRequest(
         continuationToken: String,
-        context: MSIDRequestContext
+        context: MSALNativeAuthRequestContext
     ) async -> MSALNativeAuthSignUpChallengeValidatedResponse {
         let request: MSIDHttpRequest
 
@@ -230,7 +236,7 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
             request = try requestProvider.challenge(token: continuationToken, context: context)
         } catch {
             MSALLogger.log(level: .error, context: context, format: "Error while creating Challenge Request: \(error)")
-            return .unexpectedError(message: nil)
+            return .unexpectedError(nil)
         }
 
         MSALLogger.log(level: .info, context: context, format: "Performing signup/challenge request")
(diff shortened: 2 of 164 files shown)


MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
```diff
@@ -29,10 +29,18 @@ import Foundation
 public class MSALNativeAuthError: NSObject, LocalizedError {
     /// Describes why an error occurred and provides more information about the error.
     public var errorDescription: String? { message }
+    
+    /// Correlation ID used for the request
+    public let correlationId: UUID
+
+    /// Error codes returned along with the error
+    public let errorCodes: [Int]
```
nilo-ms, line 37: As agreed errorUri field will be added as String in the public SDK error

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -163,4 +163,11 @@ class MSALNativeAuthBaseController {
             }
         }
     }
+
+    func updateContext(_ previousContext: MSIDRequestContext, with correlationId: UUID?) -> MSALNativeAuthRequestContext {
+        return MSALNativeAuthRequestContext(
+            correlationId: correlationId ?? previousContext.correlationId(),
```
rodhan-ms, line 169: Just from the signature of this method I would expect if I passed a nil correlationId that the new context would use it (I know that's not possible). I would prefer if the nil coalescing was done at the call site, or (probably better) if this method was named differently.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -163,4 +163,11 @@ class MSALNativeAuthBaseController {
             }
         }
     }
+
+    func updateContext(_ previousContext: MSIDRequestContext, with correlationId: UUID?) -> MSALNativeAuthRequestContext {
+        return MSALNativeAuthRequestContext(
```
rodhan-ms, line 168: If the correlationId is nil why not just return the `previousContext` instead of instantiating a new one?

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
rodhan-ms, line 169: If I understand correctly, don't we expect the new correlationId to be the same as the old one in most cases? Would it be useful/interesting to log this as a warning when they are different, or even just at info level?

MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift
diegojerezba: Done

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
nilo-ms: I'm wondering if it makes more sense to add a new setCorrelationId method in the `MSALNativeAuthRequestContext` class

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -164,52 +164,63 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
...
+            let error = apiError.toSignUpStartPublicError(context: context)
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in signup/start request \(error.errorDescription ?? "No error description")")
             return .init(.error(error))
         case .invalidUsername(let apiError):
-            let error = SignUpStartError(type: .invalidUsername, message: apiError.errorDescription)
+            let context = updateContext(context, with: apiError.getHeaderCorrelationId())
+            let error = SignUpStartError(
```
nilo-ms, line 203: can you use the method `apiError.toSignUpStartPublicError()` in this case?

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -246,7 +257,8 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
         context: MSIDRequestContext
     ) -> SignUpStartControllerResponse {
         switch result {
-        case .codeRequired(let sentTo, let challengeType, let codeLength, let continuationToken):
+        case .codeRequired(let sentTo, let challengeType, let codeLength, let continuationToken, let correlationId):
+            let context = updateContext(context, with: correlationId)
             MSALLogger.log(level: .info, context: context, format: "Successful signup/challenge request")
             return SignUpStartControllerResponse(
                 .codeRequired(
```
nilo-ms, line 264: at line 268 we use the correlationId from the context instead of the one received from the API. Is this behaviour expected?

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -328,16 +351,29 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
                 correlationId: context.correlationId()
             )
             return .init(.error(error: error, newState: newState))
-        case .redirect,
-             .passwordRequired:
-            let error = ResendCodeError()
+        case .redirect:
+            let error = ResendCodeError(correlationId: context.correlationId())
```
nilo-ms, line 355: just as a confirmation, is the correlation not returned from the backend in case of redirect response. Is it right?

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
rodhan-ms: I don't think so because the request's correlation ID was still what it was so updating it doesn't feel right. If we want to store the new correlationId in the `MSALNativeAuthRequestContext` then maybe we could add a new attribute for `responseCorrelationId`. But if all of this is just because we want to use the new correlationId in the error message and logs then maybe we should find a way to pass the values there directly.

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -453,21 +519,35 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
...
             stopTelemetryEvent(event, context: context, error: error)
             MSALLogger.log(level: .error,
                            context: context,
                            format: "Error in signup/continue request \(error.errorDescription ?? "No error description")")
             return .init(.error(error: error, newState: nil))
-        case .attributeValidationFailed:
-            let error = VerifyCodeError(type: .generalError)
+        case .attributeValidationFailed(let apiError, _):
+            let context = updateContext(context, with: apiError.getHeaderCorrelationId())
+            let error = VerifyCodeError(
```
nilo-ms, line 531: is there a `apiError.toVerifyCodeError()` available?

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: The return type is `MSALNativeAuthRequestContext` because the SignIn and Token flows were using it instead of `MSIDRequestContext`. 

But you're right: it makes more sense to return the same instance if there aren't any changes needed, so I've updated these flows to accept an `MSIDRequestContext` instead.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: I've updated it to `updateCorrelationId(newCorrelationId, fromContext: context)`. Let me know if you think a different name would suit it better.

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
diegojerezba: No, because `apiError.toSignUpStartPublicError()` checks only for the `error: MSALNativeAuthSignUpStartOauth2ErrorCode` property. 

In this case, the property is `.invalidRequest`, and the type of public error is decided at the Validator level using the errorCodes.

This error handling across so many layers is also a good candidate to be simplified in the future.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: Yes, good suggestion 👍. I've added it as a warning. If this happens it means that something is wrong in the backend.

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
diegojerezba: I checked and yes, it is returned. When we decided to also get the correlationId from the 2xx responses I forgot to check for the `redirect` case, but it is done now. Thanks for pointing it out 👍

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
diegojerezba: Yes, we can use it here. I've merged the logic with the `case .error(let apiError)` 🙂

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -163,4 +163,20 @@ class MSALNativeAuthBaseController {
...
+        guard let correlationId = correlationId else {
+            return originalContext
+        }
+
+        let context = MSALNativeAuthRequestContext(correlationId: correlationId, telemetryRequestId: originalContext.telemetryRequestId())
+
+        MSALLogger.log(
+            level: .warning,
+            context: context,
+            format: "Inconsistency between the correlationId sent by the SDK the one received in the response"
```
spetrescu84, line 177: I'd say log the 2 values

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -149,12 +149,20 @@ class MSALNativeAuthBaseController {
...
 
-    func performRequest<T>(_ request: MSIDHttpRequest, context: MSIDRequestContext) async -> Result<T, Error> {
+    func performRequest<T: MSALNativeAuthResponseCorrelatable>(
+        _ request: MSIDHttpRequest,
+        context: MSALNativeAuthRequestContext
+    ) async -> Result<T, Error> {
         return await withCheckedContinuation { continuation in
             request.send { result, error in
-                if let error = error {
+                if let error = error, let errorWithHeaders = error as? MSALNativeAuthResponseCorrelatable {
```
rodhan-ms, line 158: `errorWithHeaders` should be `errorWithCorrelationId` or `correlatableError` if you want to use the new word :)

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -149,12 +149,20 @@ class MSALNativeAuthBaseController {
...
+                    context.setServerCorrelationId(errorWithHeaders.correlationId)
+                    continuation.resume(returning: .failure(error))
+                } else if let error = error {
+                    MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
```
rodhan-ms, line 162: Nice

MSAL/src/native_auth/network/MSALNativeAuthCustomErrorSerializer.swift
```diff
@@ -26,9 +26,12 @@ import Foundation
 
 @_implementationOnly import MSAL_Private
 
-final class MSALNativeAuthCustomErrorSerializer<T: Decodable & Error>: NSObject, MSIDResponseSerialization {
+// swiftlint:disable:next line_length
+final class MSALNativeAuthCustomErrorSerializer<T: Decodable & Error & MSALNativeAuthResponseCorrelatable>: NSObject, MSIDResponseSerialization {
     func responseObject(for httpResponse: HTTPURLResponse?, data: Data?, context: MSIDRequestContext?) throws -> Any {
-        let customError = try JSONDecoder().decode(T.self, from: data ?? Data())
+        var customError = try JSONDecoder().decode(T.self, from: data ?? Data())
+        customError.correlationId = customError.retrieveCorrelationIdFromHeaders(from: httpResponse) // DJB: Update the context at this level?
```
rodhan-ms, line 33: Should the `// DJB` comment still be here?

MSAL/src/native_auth/network/MSALNativeAuthUrlRequestSerializer.swift
```diff
@@ -29,7 +29,7 @@ enum MSALNativeAuthUrlRequestEncoding: String {
     case json = "application/json"
 }
 
-final class MSALNativeAuthUrlRequestSerializer: NSObject, MSIDRequestSerialization {
+final class MSALNativeAuthUrlRequestSerializer: NSObject, MSIDRequestSerialization { // DJB: where is this used?
```
rodhan-ms, line 32: Reminder about another `// DJB` comment

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: @spetrescu84 - Yesterday we talked about a temporary and a final solution for this: 

As a temporary fix, we are going to add to `MSALNativeAuthRequestContext` a variable called `_serverCorrelationId` and a method `setServerCorrelationId()`. This method checks if the initial correlationId and the serverCorrelationId are the same. If they aren't, we log it as a warning, including both correlationIds in the log. And from that moment, every time we call `context.correlationId()`, we return the serverCorrelationId.

But what we are doing is conceptually wrong: we are setting the serverCorrelationId in an object that should be constant, and is actually called `...RequestContext`, not `...serverContext`. So, in the future we will introduce a `FlowContext` object, that will be a wrapper of the `MSALNativeAuthRequestContext` + a server context that we will add. We will discuss the details in future 👍

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: Thank you, done

MSAL/src/native_auth/network/MSALNativeAuthCustomErrorSerializer.swift
diegojerezba: Done

MSAL/src/native_auth/network/parameters/MSALNativeAuthRequestContext.swift
```diff
@@ -59,4 +61,23 @@ final class MSALNativeAuthRequestContext: MSIDRequestContext {
...
+
+        guard _correlationId != serverCorrelationId else {
+            return
+        }
+
+        let log = """
+                  Inconsistency between the correlationId sent by the SDK the one received in the response.
+                  Original correlationId: \(_correlationId). Server correlationId \(serverCorrelationId)
+                  """
+        MSALLogger.log(level: .info, context: self, format: log)
```
nilo-ms, line 79: shouldn't this log be a warning?

MSAL/src/native_auth/network/parameters/MSALNativeAuthRequestContext.swift
```diff
@@ -59,4 +61,23 @@ final class MSALNativeAuthRequestContext: MSIDRequestContext {
             MSID_APP_VER_KEY: appVersion
         ]
     }
+
+    func setServerCorrelationId(_ serverCorrelationId: UUID?) {
+        guard let serverCorrelationId = serverCorrelationId else {
+            MSALLogger.log(level: .info, context: self, format: "correlationId not found in server response")
```
nilo-ms, line 67: even here, shouldn't be this a warning

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -149,12 +149,20 @@ class MSALNativeAuthBaseController {
...
 
-    func performRequest<T>(_ request: MSIDHttpRequest, context: MSIDRequestContext) async -> Result<T, Error> {
+    func performRequest<T: MSALNativeAuthResponseCorrelatable>(
+        _ request: MSIDHttpRequest,
+        context: MSALNativeAuthRequestContext
+    ) async -> Result<T, Error> {
         return await withCheckedContinuation { continuation in
             request.send { result, error in
-                if let error = error {
+                if let error = error, let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
```
nilo-ms, line 158: I'd add an inner if just for the errorWithCorrelationId so we don't need to repeat the code at line 160 also at line 163. 
```suggestion
                if let error = error {
                    if let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
                        context.setServerCorrelationId(errorWithCorrelationId.correlationId)
                    }
                    continuation.resume(returning: .failure(error))
                } else if let response = result as? T {
```

MSAL/src/native_auth/network/errors/reset_password/MSALNativeAuthResetPasswordChallengeResponseError.swift
```diff
@@ -40,30 +41,60 @@ struct MSALNativeAuthResetPasswordChallengeResponseError: MSALNativeAuthResponse
...
+        self.innerErrors = innerErrors
+        self.target = target
+        self.correlationId = correlationId
     }
 }
 
 extension MSALNativeAuthResetPasswordChallengeResponseError {
 
-    func toResetPasswordStartPublicError() -> ResetPasswordStartError {
+    func toResetPasswordStartPublicError(context: MSIDRequestContext) -> ResetPasswordStartError {
```
nilo-ms, line 68: why you accept the whole context if you just need the correlationId?

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
nilo-ms: Yeah, it's not so straightforward. Would be ideal to have an error creation more homogeneous.

MSAL/src/native_auth/network/MSALNativeAuthResponseCorrelatable.swift
```diff
@@ -0,0 +1,45 @@
...
+}
+
+extension MSALNativeAuthResponseCorrelatable {
+
+    func retrieveCorrelationIdFromHeaders(from httpResponse: HTTPURLResponse?) -> UUID? {
+        guard let headers = httpResponse?.allHeaderFields else {
+            return nil
+        }
+
+        if let correlationId = headers[MSID_OAUTH2_CORRELATION_ID_REQUEST_VALUE] as? String {
```
nilo-ms, line 39: nit: you could merge this "if" with the "guard" check at line 35

MSAL/src/native_auth/network/errors/sign_up/MSALNativeAuthSignUpContinueResponseError.swift
```diff
@@ -85,11 +127,22 @@ extension MSALNativeAuthSignUpContinueResponseError {
...
+                correlationId: context.correlationId(),
+                errorCodes: errorCodes ?? [],
+                errorUri: errorURI
+            )
         }
     }
 
-    func toAttributesRequiredPublicError() -> AttributesRequiredError {
-        return AttributesRequiredError(message: errorDescription)
+    func toAttributesRequiredPublicError(context: MSIDRequestContext, message: String? = nil) -> AttributesRequiredError {
```
nilo-ms, line 140: why in this case we have the message as method parameter?

MSAL/src/native_auth/network/responses/validator/reset_password/MSALNativeAuthResetPasswordResponseValidator.swift
```diff
@@ -73,24 +73,24 @@ final class MSALNativeAuthResetPasswordResponseValidator: MSALNativeAuthResetPas
...
+            return .error(.unauthorizedClient(apiError))
         case .userNotFound:
-            return .error(.userNotFound(message: apiError.errorDescription))
+            return .error(.userNotFound(apiError))
         case .unsupportedChallengeType:
-            return .error(.unsupportedChallengeType(message: apiError.errorDescription))
+            return .error(.unsupportedChallengeType(apiError))
         case .none:
-            return .error(.unexpectedError(message: apiError.errorDescription))
+            return .error(.unexpectedError(.init(errorDescription: apiError.errorDescription)))
```
nilo-ms, line 93: If I understood correctly, if the apiError.error field is not recognised (is not a valid MSALNativeAuthResetPasswordStartOauth2ErrorCode value) we return just the error description, right?

We should return all the error details (description, errorURI, correlationId, etc.) in any case

MSAL/src/native_auth/network/responses/validator/reset_password/MSALNativeAuthResetPasswordValidatedResponses.swift
```diff
@@ -22,33 +22,58 @@
...
-    case unexpectedError(message: String?)
+    case invalidRequest(MSALNativeAuthResetPasswordStartResponseError)
+    case unauthorizedClient(MSALNativeAuthResetPasswordStartResponseError)
+    case userNotFound(MSALNativeAuthResetPasswordStartResponseError)
+    case unsupportedChallengeType(MSALNativeAuthResetPasswordStartResponseError)
+    case userDoesNotHavePassword(MSALNativeAuthResetPasswordStartResponseError)
+    case unexpectedError(MSALNativeAuthResetPasswordStartResponseError?)
 
-    func toResetPasswordStartPublicError() -> ResetPasswordStartError {
+    func toResetPasswordStartPublicError(context: MSIDRequestContext) -> ResetPasswordStartError {
```
nilo-ms, line 42: same for this class, can we just accept the correlation id ?

MSAL/src/native_auth/network/responses/validator/token/validated_response/MSALNativeAuthTokenValidatedResponse.swift
```diff
@@ -31,98 +31,189 @@ enum MSALNativeAuthTokenValidatedResponse {
...
         case .generalError:
-            return SignInStartError(type: .generalError)
-        case .userNotFound(let message):
-            return SignInStartError(type: .userNotFound, message: message)
-        case .invalidPassword(let message):
-            return SignInStartError(type: .invalidCredentials, message: message)
-        case .strongAuthRequired(let message):
-            return SignInStartError(type: .browserRequired, message: message)
-        case .expiredRefreshToken(let message):
+            return SignInStartError(type: .generalError, correlationId: context.correlationId())
```
nilo-ms, line 67: Can we return the same amount of information also for this generale error?

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignUpDelegateDispatchers.swift
```diff
@@ -47,7 +51,11 @@ final class SignUpStartDelegateDispatcher: DelegateDispatcher<SignUpStartDelegat
             telemetryUpdate?(.success(()))
             await onSignUpAttributesInvalid(attributeNames)
         } else {
-            let error = SignUpStartError(type: .generalError, message: requiredErrorMessage(for: "onSignUpAttributesInvalid"))
+            let error = SignUpStartError(
+                type: .generalError,
+                message: requiredErrorMessage(for: "onSignUpAttributesInvalid"),
+                correlationId: correlationId
```
nilo-ms, line 57: instead of using the correlationId from the DelegateDispatcher, what if we add a new method parameter at line 49?

MSAL/src/native_auth/network/parameters/MSALNativeAuthRequestContext.swift
```diff
@@ -59,4 +61,23 @@ final class MSALNativeAuthRequestContext: MSIDRequestContext {
...
+            return
```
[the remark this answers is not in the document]
diegojerezba: That's a good point, I've updated it and created a test that covers it, thanks.

MSAL/src/native_auth/network/parameters/MSALNativeAuthRequestContext.swift
```diff
@@ -59,4 +61,23 @@ final class MSALNativeAuthRequestContext: MSIDRequestContext {
...
+        }
+
+        guard _correlationId != serverCorrelationId else {
+            return
+        }
+
+        let log = """
+                  Inconsistency between the correlationId sent by the SDK the one received in the response.
```
rodhan-ms, line 76: Typo: Missing "and" between "SDK" and "the one" - should be "...the SDK **and** the one received..."

MSAL/src/native_auth/network/parameters/MSALNativeAuthRequestContext.swift
diegojerezba: Yes, done.

MSAL/src/native_auth/network/parameters/MSALNativeAuthRequestContext.swift
diegojerezba: Yes, done.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: ok, done

MSAL/src/native_auth/network/errors/reset_password/MSALNativeAuthResetPasswordChallengeResponseError.swift
diegojerezba: nice idea, I've modified them all.

MSAL/src/native_auth/network/MSALNativeAuthResponseCorrelatable.swift
diegojerezba: Done.

MSAL/src/native_auth/network/errors/sign_up/MSALNativeAuthSignUpContinueResponseError.swift
diegojerezba: Because the sign-up controller creates a custom message: 

> attribute_validation_failed from signup/continue submitAttributes request. Make sure these attributes are correct: attribute1, attribute2.

MSAL/src/native_auth/network/responses/validator/reset_password/MSALNativeAuthResetPasswordResponseValidator.swift
diegojerezba: Yes you're right, I've changed in all the `case .none:` cases, since I was only returning the errorDescription in all of them.

MSAL/src/native_auth/network/responses/validator/reset_password/MSALNativeAuthResetPasswordValidatedResponses.swift
diegojerezba: Done in all of them

MSAL/src/native_auth/network/responses/validator/token/validated_response/MSALNativeAuthTokenValidatedResponse.swift
diegojerezba: Yes, we can. But I've seen in some cases the `.generalError` is returned without having an ApiError available, so I've set it as optional.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -149,12 +149,26 @@ class MSALNativeAuthBaseController {
...
+            request.send { [weak self] result, error in
                 if let error = error {
+                    if let correlationId = self?.extractCorrelationIdFromUserInfo((error as NSError).userInfo) {
+                        context.setServerCorrelationId(UUID(uuidString: correlationId))
+                    } else if let errorWithCorrelationId = error as? MSALNativeAuthResponseCorrelatable {
+                        context.setServerCorrelationId(errorWithCorrelationId.correlationId)
+                    } else if case MSALNativeAuthInternalError.responseSerializationError(let correlationId) = error {
+                        context.setServerCorrelationId(correlationId)
+                    } else {
+                        MSALLogger.log(level: .warning, context: context, format: "Error request - cannot decode error headers. Continuing")
```
nilo-ms, line 166: I think we should set serverCorrelation Id to nil here. If don't do it, the serverCorrelationId would be still the one from the previous response, so once we retrieve a correlation Id from the context, we will use the one from the wrong one

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -149,12 +149,26 @@ class MSALNativeAuthBaseController {
...
+    func performRequest<T: MSALNativeAuthResponseCorrelatable>(
+        _ request: MSIDHttpRequest,
+        context: MSALNativeAuthRequestContext
+    ) async -> Result<T, Error> {
         return await withCheckedContinuation { continuation in
-            request.send { result, error in
+            request.send { [weak self] result, error in
                 if let error = error {
+                    if let correlationId = self?.extractCorrelationIdFromUserInfo((error as NSError).userInfo) {
+                        context.setServerCorrelationId(UUID(uuidString: correlationId))
```
nilo-ms, line 160: can you add a comment for each code block explaining when it should be called?

MSAL/src/native_auth/network/MSALNativeAuthCustomErrorSerializer.swift
```diff
@@ -26,10 +26,18 @@ import Foundation
...
-        throw customError
+        do {
+            var customError = try JSONDecoder().decode(T.self, from: data ?? Data())
+            customError.correlationId = T.retrieveCorrelationIdFromHeaders(from: httpResponse)
+
+            // the successfuly constructed "customError" needs to be thrown, 
+            // since the previous "try" command just validates the object (error) decoding
+            throw customError
+        } catch is DecodingError {
+            MSALLogger.log(level: .error, context: context, format: "CustomErrorSerializer failed decoding. Key not found")
```
nilo-ms, line 39: can you specify "correlation Id key" in this error message?

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: Nice one, thanks.

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
diegojerezba: Done

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignUpDelegateDispatchers.swift
diegojerezba: As agreed, I've added a correlation-id in the response of every controller. This correlation-id is being used in every method of the dispatchers now.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -204,14 +211,20 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
         delegate: SignInStartDelegate
     ) {
         Task {
+            let correlationId = correlationId ?? .init()
```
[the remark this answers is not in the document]
diegojerezba: Removed that part of the log to make it less confusing.

MSAL/src/native_auth/network/MSALNativeAuthCustomErrorSerializer.swift
diegojerezba: Removed that part of the log to make it less confusing.

nilo-ms (approved):

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -30,7 +30,7 @@ import Foundation
 /// to the initialiser method.
 ///
 /// For example:
- 
+
```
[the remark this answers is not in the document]
diegojerezba: It was an extra space that I removed. Swiftlint was complaining about that

PR #2019, opened 2024-02-20
Fix static framework - Add dedicated pod subspec for native auth code

nilo-ms:
## Proposed changes

Swift code currently doesn't work when cocoa pod - static framework is used. To avoid the introduction of a breaking change for existing MSAL customers that uses the static framework, a new pod subspec has been created. Swift code (and native auth features) has been added to this subspec, this means that customers that use cocoa pod and will want to use native auth features will need to specify this subspec in the podfile:
`pod 'MSAL/native-auth'`
The default subspec will not contains any swift files, so cocoa pod - static framework will continue working as expected.
## Type of change

- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information
The only "native auth" code that has been added to the default subspec is the file "MSALNativeAuthChallengeTypes.h",   we did it, so we don't need to have two different version of the MSAL.h file.

### DIFF
--- a/MSAL.podspec
+++ b/MSAL.podspec
@@ -30,7 +30,7 @@ Pod::Spec.new do |s|
     app.pod_target_xcconfig = { 'CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF' => 'NO' }
     app.source_files = "MSAL/src/**/*.{h,m}", "MSAL/IdentityCore/IdentityCore/src/**/*.{h,m}", "MSAL/src/native_auth/public/*.{h,m}"
     app.ios.public_header_files = "MSAL/src/public/*.h","MSAL/src/public/ios/**/*.h", "MSAL/src/public/configuration/**/*.h", "MSAL/src/native_auth/public/*.h"
-    app.osx.public_header_files = "MSAL/src/public/mac/*.h","MSAL/src/public/*.h", "MSAL/src/public/configuration/**/*.h", "MSAL/src/native_auth/public/*.h"
+    app.osx.public_header_files = "MSAL/src/public/mac/*.h","MSAL/src/public/*.h", "MSAL/src/public/configuration/**/*.h"
     app.ios.exclude_files = "MSAL/src/**/mac/*", "MSAL/IdentityCore/IdentityCore/src/**/mac/*"  
     app.osx.exclude_files = "MSAL/src/**/ios/*", "MSAL/IdentityCore/IdentityCore/src/**/ios/*"
     app.requires_arc = true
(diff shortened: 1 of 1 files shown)


spetrescu84 (approved):

MSAL.podspec
```diff
@@ -20,34 +20,45 @@ Pod::Spec.new do |s|
...
+  
   s.default_subspecs ='app-lib'
   
   s.prefix_header_file = "MSAL/src/MSAL.pch"
   s.header_dir = "MSAL"
 
   s.subspec 'app-lib' do |app|
-    app.source_files = "MSAL/src/**/*.{h,m}", "MSAL/src/native_auth/**/*.{h,m,swift}", "MSAL/IdentityCore/IdentityCore/src/**/*.{h,m}", "MSAL/module.modulemap"
+    app.pod_target_xcconfig = { 'CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF' => 'NO' }
+    app.source_files = "MSAL/src/**/*.{h,m}", "MSAL/IdentityCore/IdentityCore/src/**/*.{h,m}", "MSAL/src/native_auth/public/*.{h,m}"
```
diegojerezba, line 31: Regarding the line: `..., "MSAL/src/native_auth/public/*.{h,m}"`, shouldn't we exclude everything native_auth related?

nilo-ms: I added the reason we added this line in the "additional information" in the PR description. In brief, we add the Objective-C code to avoid having two different MSAL.h files

MSAL.podspec
```diff
@@ -20,34 +20,45 @@ Pod::Spec.new do |s|
...
   
-  	# There is currently a bug in CocoaPods where it doesn't combine the public headers
-  	# for both the platform and overall.
-  	ext.ios.exclude_files = "MSAL/src/**/mac/*", "MSAL/IdentityCore/IdentityCore/src/**/mac/*"
-  	ext.osx.exclude_files = "MSAL/src/**/ios/*", "MSAL/IdentityCore/IdentityCore/src/**/ios/*", "MSAL/src/native_auth/**/*", "MSAL/module.modulemap"
-  	ext.requires_arc = true
+    # There is currently a bug in CocoaPods where it doesn't combine the public headers
+    # for both the platform and overall.
+    ext.ios.exclude_files = "MSAL/src/**/mac/*", "MSAL/IdentityCore/IdentityCore/src/**/mac/*"
+    ext.osx.exclude_files = "MSAL/src/**/ios/*", "MSAL/IdentityCore/IdentityCore/src/**/ios/*"
```
diegojerezba, line 60: In the line 53 we are including all of the native_auth files: ` "MSAL/src/native_auth/public/*.{h,m}"`, but in the line 60 we are not excluding them for osx.

Also, we are also not including the modulemap, not sure if that's intended or not.

nilo-ms: The only file that it will be copied is the MSALNativeAuthChallengeTypes.h. There is no harm of leaving for osx because it's not used, but for cleanliness I'll exclude it.
About the modulemap, we're not adding it because we're not adding any swift files, so yes, it's expected

PR #2141, opened 2024-05-01
ID Token should be updated when Access Token is refreshed/acquired silently

spetrescu84:
## Proposed changes

The Auth Tokens present in the UserAccountResult and specifically the ID Token alongside the MSALAccount get updated when a refresh token operation happens

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information
To test this, the test case Verify split issuer needs to be run. The UserAccountResult returned when the getAccount is called, needs to have the ID Token updated when refreshToken is called and the MSALAccount should have the Account Claims updated

### DIFF
--- a/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
+++ b/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
@@ -83,7 +83,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
         return nil
     }
 
-    func refreshToken(context: MSALNativeAuthRequestContext, authTokens: MSALNativeAuthTokens) async -> RefreshTokenCredentialControllerResponse {
+    func refreshToken(context: MSALNativeAuthRequestContext, authTokens: MSALNativeAuthTokens, userAccountResult: MSALNativeAuthUserAccountResult) async -> RefreshTokenCredentialControllerResponse {
         MSALLogger.log(level: .verbose, context: context, format: "Refresh started")
         let telemetryEvent = makeAndStartTelemetryEvent(id: .telemetryApiIdRefreshToken, context: context)
         let scopes = authTokens.accessToken.scopes.array as? [String] ?? []
@@ -104,7 +104,8 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
             response,
             scopes: scopes,
             context: context,
-            telemetryEvent: telemetryEvent
+            telemetryEvent: telemetryEvent,
+            userAccountResult: userAccountResult
         )
     }
 
@@ -147,7 +148,8 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
         _ response: MSALNativeAuthTokenValidatedResponse,
         scopes: [String],
         context: MSALNativeAuthRequestContext,
-        telemetryEvent: MSIDTelemetryAPIEvent?
+        telemetryEvent: MSIDTelemetryAPIEvent?,
+        userAccountResult: MSALNativeAuthUserAccountResult
     ) -> RefreshTokenCredentialControllerResponse {
         let config = factory.makeMSIDConfiguration(scopes: scopes)
         switch response {
@@ -156,7 +158,8 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
                 tokenResponse: tokenResponse,
                 telemetryEvent: telemetryEvent,
                 context: context,
-                config: config
+                config: config,
+                userAccountResult: userAccountResult
             )
         case .error(let errorType):
             let error = errorType.convertToRetrieveAccessTokenError(correlationId: context.correlationId())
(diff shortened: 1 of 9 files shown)


MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -182,12 +189,40 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
                 context: context,
                 format: "Refresh Token completed successfully")
             // TODO: Handle tokenResult.refreshToken as? MSIDRefreshToken in a safer way
```
diegojerezba, line 191: Now that we are changing this part, can you handle this `TODO` comment?

diegojerezba (approved):
Left a small comment but LGTM, there isn't a reference cycle between `MSALNativeAuthUserAccountResult` and `CredentialsController` (I was a bit wary of the UserAccountResult passing itself to the CredentialsController, but it seems the object is being correctly deallocated).

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -27,13 +27,13 @@ import Foundation
...
 
-    let authTokens: MSALNativeAuthTokens
+    var authTokens: MSALNativeAuthTokens
     let configuration: MSALNativeAuthConfiguration
     private let cacheAccessor: MSALNativeAuthCacheInterface
 
-    /// Get the ID token for the account.
+    /// Get the latest ID token for the account.
     @objc public var idToken: String? {
         authTokens.rawIdToken
```
diegojerezba, line 38: @spetrescu84 did you consider retrieving the id token from the cache here? Maybe that would be a cleaner way, as you wouldn't have to make the `account` property mutable and would avoid most of the new changes.

I'm not sure if the performance would be affected, though.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -169,11 +174,13 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
         }
     }
 
+    // swiftlint:disable:next function_body_length
```
nilo-ms, line 177: instead of adding an exception for swiftlint, can you move the claims extraction/account creation in another private method?

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -182,12 +189,40 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
...
+                        context: context,
+                        format: "Initialising account without claims")
+                }
+            } catch {
+                MSALLogger.log(
+                    level: .error,
+                    context: context,
+                    format: "Claims for account could not be created - \(error)" )
+            }
+            guard let account = MSALAccount.init(msidAccount: tokenResult.account,
```
nilo-ms, line 213: Can we reuse the code from the MSALNativeAuthResultFactory class?

MSAL/test/unit/native_auth/controllers/MSALNativeAuthCredentialsControllerTests.swift
```diff
@@ -156,15 +159,18 @@ final class MSALNativeAuthCredentialsControllerTests: MSALNativeAuthTestCase {
         factory.mockMakeUserAccountResult(userAccountResult)
         tokenResult.accessToken = MSIDAccessToken()
         tokenResult.accessToken.accessToken = expectedAccessToken
+        tokenResult.rawIdToken = expectedIdToken
         responseValidatorMock.tokenValidatedResponse = .success(tokenResponse)
-        cacheAccessorMock.mockUserAccounts = [account]
+        cacheAccessorMock.mockUserAccounts = [newAccount!]
```
nilo-ms, line 164: can we avoid the force unwrap here?

MSAL/test/unit/native_auth/controllers/MSALNativeAuthCredentialsControllerTests.swift
```diff
@@ -121,13 +121,14 @@ final class MSALNativeAuthCredentialsControllerTests: MSALNativeAuthTestCase {
 
         let expectedContext = MSALNativeAuthRequestContext(correlationId: defaultUUID)
```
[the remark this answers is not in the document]
spetrescu84: Added, but not here, but rather in `MSALNativeAuthResultFactoryTests` as that's where it should be. Also changed how the code works because the `MSALAccount` always gets created no matter what as it's not an optional (we had it wrong because of Objective C/Swift interoperability).
Changed the names of the tests on this class to highlight the `userAccountResult` doesn't get created rather than the `account`

MSAL/test/unit/native_auth/controllers/MSALNativeAuthCredentialsControllerTests.swift
nilo-ms: The only point is that if newAccount is nil, we crash, and all the next tests are not executed. While if we gracefully unwrap it only this test will fail

PR #2152, opened 2024-05-08
Feature/multiple tokens 3

borgesmb:
## Proposed changes

Added support for multiple access tokens.

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
+++ b/MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
@@ -28,11 +28,14 @@ class MSALNativeAuthBaseController {
 
     typealias TelemetryInfo = (event: MSIDTelemetryAPIEvent?, context: MSALNativeAuthRequestContext)
     let clientId: String
+    weak var application: MSALNativeAuthPublicClientApplication?
 
     init(
-        clientId: String
+        clientId: String,
+        application: MSALNativeAuthPublicClientApplication?
     ) {
         self.clientId = clientId
+        self.application = application
     }
 
     func makeAndStartTelemetryEvent(
--- a/MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
+++ b/MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
@@ -42,7 +42,8 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
         cacheAccessor: MSALNativeAuthCacheInterface,
         factory: MSALNativeAuthResultBuildable,
         signInResponseValidator: MSALNativeAuthSignInResponseValidating,
-        tokenResponseValidator: MSALNativeAuthTokenResponseValidating
+        tokenResponseValidator: MSALNativeAuthTokenResponseValidating,
+        application: MSALNativeAuthPublicClientApplication?
     ) {
         self.signInRequestProvider = signInRequestProvider
         self.signInResponseValidator = signInResponseValidator
@@ -51,12 +52,13 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
             requestProvider: tokenRequestProvider,
             cacheAccessor: cacheAccessor,
             factory: factory,
-            responseValidator: tokenResponseValidator
+            responseValidator: tokenResponseValidator,
+            application: application
         )
     }
 
-    convenience init(config: MSALNativeAuthConfiguration, cacheAccessor: MSALNativeAuthCacheInterface) {
-        let factory = MSALNativeAuthResultFactory(config: config, cacheAccessor: cacheAccessor)
+    convenience init(config: MSALNativeAuthConfiguration, cacheAccessor: MSALNativeAuthCacheInterface, application: MSALNativeAuthPublicClientApplication?) {
+        let factory = MSALNativeAuthResultFactory(config: config, cacheAccessor: cacheAccessor, application: application)
         self.init(
             clientId: config.clientId,
             signInRequestProvider: MSALNativeAuthSignInRequestProvider(
@@ -68,7 +70,8 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
             signInResponseValidator: MSALNativeAuthSignInResponseValidator(),
             tokenResponseValidator: MSALNativeAuthTokenResponseValidator(
                 factory: factory,
-                msidValidator: MSIDTokenResponseValidator())
+                msidValidator: MSIDTokenResponseValidator()), 
+                application: application
         )
     }
 
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
@@ -88,7 +88,7 @@ extension MSALNativeAuthPublicClientApplication {
     ) async -> MSALNativeAuthResetPasswordControlling.ResetPasswordStartControllerResponse {
         let context = MSALNativeAuthRequestContext(correlationId: correlationId)
         let correlationId = context.correlationId()
-        
+
         guard inputValidator.isInputValid(username) else {
             return .init(.error(ResetPasswordStartError(type: .invalidUsername, correlationId: correlationId)), correlationId: correlationId)
         }
@@ -118,5 +118,20 @@ extension MSALNativeAuthPublicClientApplication {
 
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
+    }
+
+    func copy(with zone: NSZone? = nil) -> Any? {
+        var challenges: MSALNativeAuthChallengeTypes = []
+        for challenge in internalChallengeTypes {
+            switch challenge {
+            case .oob:
+                challenges.insert(.OOB)
+            case .password:
+                challenges.insert(.password)
+            default:
+                break
+            }
+        }
+        return try? MSALNativeAuthPublicClientApplication(configuration: configuration, challengeTypes: challenges)
     }
 }
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -307,41 +307,5 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
                 }
             }
         }
-
-//        Task {
-//            
-//            let params = MSALSilentTokenParameters(scopes: scopes ?? [], account: account)
-//            params.forceRefresh = forceRefresh
-//            
-//            do {
-//                let response = try await withCheckedThrowingContinuation({ [weak self] (continuation: CheckedContinuation<MSALResult, Error>) in
-//                    guard let self = self else { return }
-//                    
-//                    self.acquireTokenSilent(with: params) { result, error in
-//                        
-//                        guard let result = result else {
-//                            continuation.resume(throwing: error!)
-//                            return
-//                        }
-//                        
-//                        continuation.resume(returning: result)
-//                    }
-//                })
-//                
-//                let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: nil) // controllerResponse.telemetryUpdate)
-//                let accessTokenResult = MSALNativeAuthTokenResult(accessToken: response.accessToken,
-//                                                            scopes: response.scopes,
-//                                                            expiresOn: response.expiresOn)
-//                
-//                await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(
-//                    result: accessTokenResult,
-//                    correlationId: response.correlationId
-//                )
-//                
-//            } catch {
-//                let e = RetrieveAccessTokenError(type: .generalError, correlationId: correlationId ?? UUID())
-//                await delegate.onAccessTokenRetrieveError(error: e)
-//            }
-//        }
     }
 }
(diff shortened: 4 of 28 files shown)


MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,6 +78,7 @@ import Foundation
     ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
     ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
     ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    @available(*, deprecated, message: "Use 'getAccessToken' method from 'MSALNativeAuthPublicClientApplication' instance")
```
spetrescu84, line 81: I don't understand this message, is the method not in `MSALNativeAuthUserAccountResult`? Rather than `MSALNativeAuthPublicClientApplication`?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -32,6 +32,7 @@ import Foundation
     var authTokens: MSALNativeAuthTokens
     let configuration: MSALNativeAuthConfiguration
     private let cacheAccessor: MSALNativeAuthCacheInterface
+    private let application: MSALNativeAuthPublicClientApplication?
```
nilo-ms, line 35: should this reference be weak?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +100,38 @@ import Foundation
             }
         }
     }
+
```
nilo-ms, line 103: @borgesmb could you align with Android comment for getAccessToken and getAccessToken with scopes?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
```diff
@@ -119,4 +119,19 @@ extension MSALNativeAuthPublicClientApplication {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
     }
+
+    func copy(with zone: NSZone? = nil) -> Any? {
```
nilo-ms, line 123: @borgesmb can you please clarify why this method is need?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: This should be a strong reference, in order to hold a the public client application and have access to the get access token method, in case the user nullify the original NativeAuthPublicClientApplication object. This property is filled with a copy of the object passed in its constructor (line 53).

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Sorry, I changed the code base on the agreement we had in that meeting, but forgot to correct the deprecation comment.

nilo-ms:
@borgesmb can we get rid of the `MSALNativeAuthTokens` class? It should not be used anymore

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
borgesmb: it is needed in order to produce a copy of the client application, to be used in the account result. So, the result will have access to the client application, and the get access token method, even in case the user nullify the original client application object.

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -51,12 +52,13 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
-            responseValidator: tokenResponseValidator
+            responseValidator: tokenResponseValidator,
+            application: application
         )
     }
 
-    convenience init(config: MSALNativeAuthConfiguration, cacheAccessor: MSALNativeAuthCacheInterface) {
-        let factory = MSALNativeAuthResultFactory(config: config, cacheAccessor: cacheAccessor)
+    convenience init(config: MSALNativeAuthConfiguration, cacheAccessor: MSALNativeAuthCacheInterface, application: MSALNativeAuthPublicClientApplication?) {
+        let factory = MSALNativeAuthResultFactory(config: config, cacheAccessor: cacheAccessor, application: application)
```
nilo-ms, line 61: Personally, I'm not a fan of using a reference of the publicClientApplication, but I don't have a better solution for this. So, no complaint.
The only thing that I suggest is to have this instance only in the classes that we need it. If the application instance is only needed to create this factory, I'd add this factory as parameter in the convenience init.
I'm also wondering why we don't use the ControllerFactory to create an instance of the SignInController in the SignUp and ResetPassword controller. If we do, we could avoid sharing around the codebase this "application" instance.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +100,38 @@ import Foundation
...
+    ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
+    ///   - scopes: Optional. Permissions that should be included in the access token received after sign in flow has completed
+    ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
+    ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    public func getAccessToken(forceRefresh: Bool = false,
+                               scopes: [String]? = nil,
+                               correlationId: UUID? = nil,
+                               delegate: CredentialsDelegate) {
+
+        let params = MSALSilentTokenParameters(scopes: scopes ?? [], account: account)
```
nilo-ms, line 116: can you assign correlationId too?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +100,38 @@ import Foundation
...
+        params.forceRefresh = forceRefresh
+
+        application?.acquireTokenSilent(with: params) { result, error in
+
+            if let error = error as? NSError {
+                let accessTokenError = RetrieveAccessTokenError(type: .generalError,
+                                                                correlationId: correlationId ?? result?.correlationId ?? UUID(),
```
nilo-ms, line 123: we should always use result.correlationID in this case

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +100,38 @@ import Foundation
...
+    ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
+    ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    public func getAccessToken(forceRefresh: Bool = false,
+                               scopes: [String]? = nil,
+                               correlationId: UUID? = nil,
+                               delegate: CredentialsDelegate) {
+
+        let params = MSALSilentTokenParameters(scopes: scopes ?? [], account: account)
+        params.forceRefresh = forceRefresh
+
```
nilo-ms, line 118: can you log something related to native auth here? So, we can understand that the next silent token retrieval is triggered by native auth.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +100,38 @@ import Foundation
...
+                Task { await delegate.onAccessTokenRetrieveError(error: accessTokenError) }
+            }
+
+            if let result = result {
+                let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: nil)
+                let accessTokenResult = MSALNativeAuthTokenResult(accessToken: result.accessToken,
+                                                                  scopes: result.scopes,
+                                                                  expiresOn: result.expiresOn)
+                Task { await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(result: accessTokenResult, correlationId: result.correlationId) }
+            }
```
nilo-ms, line 134: in the rare case both "if" are not true, we don't dispatch any callback. Can you please dispatch an error callback here?

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -28,11 +28,14 @@ class MSALNativeAuthBaseController {
 
     typealias TelemetryInfo = (event: MSIDTelemetryAPIEvent?, context: MSALNativeAuthRequestContext)
     let clientId: String
+    weak var application: MSALNativeAuthPublicClientApplication?
```
diegojerezba, line 31: why is it needed for all the controllers to hold an instance of `MSALNativeAuthPublicClientApplication`?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +100,38 @@ import Foundation
...
+            if let error = error as? NSError {
+                let accessTokenError = RetrieveAccessTokenError(type: .generalError,
+                                                                correlationId: correlationId ?? result?.correlationId ?? UUID(),
+                                                                errorCodes: [error.code])
+                Task { await delegate.onAccessTokenRetrieveError(error: accessTokenError) }
+            }
+
+            if let result = result {
+                let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: nil)
+                let accessTokenResult = MSALNativeAuthTokenResult(accessToken: result.accessToken,
```
nilo-ms, line 130: we should update the idToken too here.
@spetrescu84 am I wrong?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -47,11 +47,12 @@ import Foundation
 @objcMembers
 public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplication {
```
diegojerezba, line 48: @borgesmb I don't like this approach in general. It doesn't feel good. We are adding the `application` class to all the controllers, to the ControllerFactory, to the ResultFactory...

If all that we need is to call `MSALPublicClientApplication` from the `MSALNativeAuthUserAccountResult`, is it possible to instantiate `MSALPublicClientApplication` directly from there?

I see that all we need is the config that is set during the creation of `MSALNativeAuthPublicClientApplication`, but I'd be willing to have it as a global variable accessible by any class. It's not good either, but less intrusive than keeping a copy of the Application class everywhere.

This is what I mean with the global configuration:

```
var globalConfiguration: MSALPublicClientApplicationConfig?

@objcMembers
public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplication
```

And then instantiate PublicClientApplication directly in the `MSALNativeAuthUserAccountResult`:

In MSALNativeAuthUserAccountResult.swift:
```
    public func getAccessToken(forceRefresh: Bool = false,
                               scopes: [String]? = nil,
                               correlationId: UUID? = nil,
                               delegate: CredentialsDelegate) {

            let msal = try MSALPublicClientApplication(configuration: globalConfiguration!)

            let params = MSALSilentTokenParameters(scopes: scopes ?? [], account: account)
            params.forceRefresh = forceRefresh

            msal.acquireTokenSilent(with: params) { result, error in
               ...
            }
    }
```

Do you think something like this would be possible?

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
borgesmb: Not anymore. Please take a look on the new PR; https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Addressed. Please take a look on the new PR; https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Addressed. Please take a look on the new PR; https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: I didn't get exactly the question.

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
borgesmb: Addressed. Please take a look on the new PR; https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
borgesmb: Addressed. Please take a look on the new PR; https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Do you have any idea of what to log exactly?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
spetrescu84: it should yes

PR #2159, opened 2024-05-10
Feature/multiple tokens 4

borgesmb:
## Proposed changes

Implements support for multiple access tokens.

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -68,6 +71,8 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
             throw MSALNativeAuthInternalError.invalidAuthority
         }
 
+        MSALNativeAuthPublicClientApplication.sharedConfiguration = config
```
nilo-ms, line 74: Should we set this shared config/challenge types after the super.init? I'm saying this because the super.init can throw an error, and we want to assign this shared config only when we know that a new publicClientApplication instance is created. Same comment applies for the other init

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -120,6 +125,8 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
             redirectUri: redirectUri,
             authority: ciamAuthority
         )
+        MSALNativeAuthPublicClientApplication.sharedConfiguration = configuration
+        MSALNativeAuthPublicClientApplication.sharedChallengeTypes = challengeTypes
 
         if redirectUri == nil {
             MSALLogger.log(level: .warning, context: nil, format: MSALNativeAuthErrorMessage.redirectUriNotSetWarning)
```
nilo-ms, line 132: Should we also set the shared variables also for the init at line 144?

MSAL/src/native_auth/public/MSALNativeAuthTokenResult.swift
```diff
@@ -26,28 +26,32 @@ import Foundation
 
 public class MSALNativeAuthTokenResult: NSObject {
 
-    let authTokens: MSALNativeAuthTokens
+    let internalAccessToken: String
```
nilo-ms, line 29: nit: can these be private?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,6 +75,7 @@ import Foundation
     ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
     ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
     ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    @available(*, deprecated, message: "Use the 'getAccessToken' method that also receives a list of scopes to be included in the token")
     @objc public func getAccessToken(forceRefresh: Bool = false, correlationId: UUID? = nil, delegate: CredentialsDelegate) {
         Task {
```
nilo-ms, line 80: This method and the one with scopes, can (and need IMO) use the same logic.
We should be able to get rid of the `getAccessTokenInternal` method

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +97,47 @@ import Foundation
...
+    ///   - scopes: Optional. Permissions that should be included in the access token received after sign in flow has completed
+    ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
+    ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    public func getAccessToken(forceRefresh: Bool = false,
+                               scopes: [String]? = nil,
+                               correlationId: UUID? = nil,
+                               delegate: CredentialsDelegate) {
+
+        let params = MSALSilentTokenParameters(scopes: scopes ?? [], account: account)
+        params.forceRefresh = forceRefresh
```
nilo-ms, line 114: we need to set the correlation Id in this param

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +97,47 @@ import Foundation
...
+
+        guard let config = MSALNativeAuthPublicClientApplication.sharedConfiguration,
+              let challengeTypes = MSALNativeAuthPublicClientApplication.sharedChallengeTypes,
+              let client = try? MSALNativeAuthPublicClientApplication(configuration: config, challengeTypes: challengeTypes)
+        else { return }
```
nilo-ms, line 119: can you call the delegate error callback here? The user will not receive any callback in this case

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -96,4 +97,47 @@ import Foundation
...
+    ///   - client: The instance of Native Auth public client application.
+    ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
+    ///   - scopes: Optional. Permissions that should be included in the access token received after sign in flow has completed
+    ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
+    ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    public func getAccessToken(forceRefresh: Bool = false,
+                               scopes: [String]? = nil,
+                               correlationId: UUID? = nil,
+                               delegate: CredentialsDelegate) {
+
```
nilo-ms, line 112: Can you log here something like:
"Retrieving access token with scopes started". 
This can be useful for debugging purpose.
We usually log something at the beginning of the controller methods, but in this case, we don't call the controller

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,6 +75,7 @@ import Foundation
     ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
     ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
     ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    @available(*, deprecated, message: "Use the 'getAccessToken' method that also receives a list of scopes to be included in the token")
```
diegojerezba, line 78: For a user that simply wants to do `result.getAccessToken(delegate: self)` this is marked as deprecated because Swift doesn't have a way to know which `getAccessToken` method the user is calling. The only way to silence this warning is by passing scopes.

Screenshot 2024-05-13 at 14 15 01

IMO this worsens the Swift developer experience, because it gives the impression that they are using a deprecated method (when they're not) in those use cases that don't require more scopes.

However if we don't add a second method, this could become a breaking change for objective-c users. Ideally we would have the following:

1. In the same Swift method, we add the `scopes` parameter with a default value. This method will be see from both Swift and Objc.
2. We create an obj-only "old" method without the scopes and we deprecated it, so the existing Objc users will see be able to continue using it and will see the deprecation warning. I'm not sure if this is possible or not tbh. Can you check it?

nilo-ms: Hey Diego, this method will not be deprecated anymore :)
Marcos is going to update the PR.

diegojerezba: Nice. But what happens with the existing objc users that may have used it @nilo-ms ?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Logic is now in one place.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
```diff
@@ -119,4 +119,19 @@ extension MSALNativeAuthPublicClientApplication {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
     }
+    
+    func getChallengeTypesFromInternalChallengeTypes(_ internalChallengeTypes: [MSALNativeAuthInternalChallengeType]) -> MSALNativeAuthChallengeTypes {
```
diegojerezba, line 123: nit: the function above `getInternalChallengeTypes()` is static. Can you make this static too?

Also there are two warnings from swiftlint.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,20 +26,45 @@ import Foundation
...
+            if let result = result {
+                let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: nil)
+                let accessTokenResult = MSALNativeAuthTokenResult(accessToken: result.accessToken,
+                                                                  scopes: result.scopes,
+                                                                  expiresOn: result.expiresOn)
+                Task { await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(result: accessTokenResult, correlationId: result.correlationId) }
+                return
+            }
+
+            Task { await delegate.onAccessTokenRetrieveError(error: RetrieveAccessTokenError(type: .generalError,
```
diegojerezba, line 66: nit: swiftlint warning

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
```diff
@@ -31,6 +31,7 @@ public class RetrieveAccessTokenError: MSALNativeAuthError {
         case browserRequired
         case refreshTokenExpired
         case generalError
+        case invalidScope
```
diegojerezba, line 34: We need to add a new public utility function `isInvalidScope()` so the developers can check for this.

Remember to add a new test to the `RetrieveAccessTokenErrorTests` to test this new method too.

MSAL/test/unit/native_auth/public/MSALNativeAuthUserAccountResultTests.swift
```diff
@@ -54,17 +54,20 @@ class MSALNativeAuthUserAccountResultTests: XCTestCase {
...
-        let expectation = expectation(description: "CredentialsController")
-        let authTokens = MSALNativeAuthUserAccountResultStub.authTokens
-        let mockDelegate = CredentialsDelegateSpy(expectation: expectation, expectedResult: MSALNativeAuthTokenResult(authTokens: authTokens))
-        mockDelegate.expectedAccessToken = authTokens.accessToken.accessToken
-        mockDelegate.expectedExpiresOn = authTokens.accessToken.expiresOn
-        mockDelegate.expectedScopes = authTokens.accessToken.scopes.array as? [String] ?? []
-        sut.getAccessToken(delegate: mockDelegate)
-        wait(for: [expectation], timeout: 1)
-    }
+    // To-do: change this test by moching MSAL cache accessor
```
diegojerezba, line 57: What prevents us from fixing this test in this PR?

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -88,7 +88,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
                       userAccountResult: MSALNativeAuthUserAccountResult) async -> RefreshTokenCredentialControllerResponse {
```
nilo-ms, line 88: we can get rid of this "refreshToken" method. 
Can you please remove it also from the `MSALNativeAuthCredentialsControlling` interface?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,20 +26,45 @@ import Foundation
...
+                                delegate: CredentialsDelegate) {
+
+        let params = MSALSilentTokenParameters(scopes: scopes, account: account)
+        params.forceRefresh = forceRefresh
+        params.correlationId = correlationId
+
+        guard let config = MSALNativeAuthPublicClientApplication.sharedConfiguration,
+              let challengeTypes = MSALNativeAuthPublicClientApplication.sharedChallengeTypes,
+              let client = try? MSALNativeAuthPublicClientApplication(configuration: config, challengeTypes: challengeTypes)
+        else {
```
nilo-ms, line 41: can you log an error here:
```suggestion
        else {
        MSALLogger.log(
                level: .error,
                context: context,
                format: "Config or challenge types unexpectedly found nil"
            )
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,25 +75,41 @@ import Foundation
     ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
```
nilo-ms, line 75: can you please align with MSAL Android comment for this method?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,25 +75,41 @@ import Foundation
...
+                               scopes: [],
+                               correlationId: correlationId,
+                               delegate: delegate)
+    }
 
-            let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: controllerResponse.telemetryUpdate)
+    /// Retrieves an access token for the account.
+    /// - Parameters:
+    ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
+    ///   - scopes: Optional. Permissions that should be included in the access token received after sign in flow has completed
```
nilo-ms, line 90: scopes are not optional

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,25 +75,41 @@ import Foundation
...
+                                     correlationId: UUID? = nil,
+                                     delegate: CredentialsDelegate) {
+        getAccessTokenInternal(forceRefresh: forceRefresh,
+                               scopes: [],
+                               correlationId: correlationId,
+                               delegate: delegate)
+    }
 
-            let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: controllerResponse.telemetryUpdate)
+    /// Retrieves an access token for the account.
```
nilo-ms, line 87: also for this method, can you please align with MSAL Android?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
nilo-ms, line 90: can we use something more similar to what MSAL is using today?
```suggestion
    ///   - scopes: Permissions you want included in the access token received
                    in the result. Not all scopes are
                    guaranteed to be included in the access token returned.
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,25 +75,41 @@ import Foundation
...
-            switch controllerResponse.result {
-            case .success(let accessTokenResult):
-                await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(
-                    result: accessTokenResult,
-                    correlationId: controllerResponse.correlationId
-                )
-            case .failure(let error):
-                await delegate.onAccessTokenRetrieveError(error: error)
-            }
+        guard !scopes.isEmpty else {
```
nilo-ms, line 98: can you please add new method to the MSALNativeAuthInputValidator class and use it to check the input scopes?

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
nilo-ms, line 34: @borgesmb are we aligned with MSAL Android for this error name?

MSAL/src/native_auth/public/state_machine/error/RetrieveAccessTokenError.swift
```diff
@@ -65,4 +68,9 @@ public class RetrieveAccessTokenError: MSALNativeAuthError {
     public var isRefreshTokenExpired: Bool {
         return type == .refreshTokenExpired
     }
+    
+    /// Returns `true` if the scope is invalid.
```
nilo-ms, line 72: What if we use something like:
```suggestion
    /// Returns `true` if the scope list is empty.
```

nilo-ms:
@borgesmb can you please update the Changelog file?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
diegojerezba: Sorry my bad, the way it is now won't affect existing users 👍

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
SammyO, line 88: Is it not possible to completely use existing logic? I.e. why do we as native auth need our own refresh token logic?

Even if we do need our own function for this, why do we need specific native auth classes for tokens and account result? Native auth tokens are no different than browser-delegated tokens, or accounts. so can both flows not use the same classes?

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -88,7 +88,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
...
         MSALLogger.log(level: .verbose, context: context, format: "Refresh started")
         let telemetryEvent = makeAndStartTelemetryEvent(id: .telemetryApiIdRefreshToken, context: context)
-        let scopes = authTokens.accessToken.scopes.array as? [String] ?? []
+        let scopes = authTokens.accessToken.scopes?.array as? [String] ?? []
```
SammyO, line 91: Why is this nullable (given all the conversations we've had about not allowing null and empty scopes)?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
```diff
@@ -119,4 +119,21 @@ extension MSALNativeAuthPublicClientApplication {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
     }
+
+    static func getChallengeTypesFromInternalChallengeTypes(
```
SammyO, line 123: I would change the name of this method to something that reflects what it does, and/or add comments. Looking at this one vs. the method above, it's not clear to me what each of them does, and why we need two.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -49,8 +49,11 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
 
     let controllerFactory: MSALNativeAuthControllerBuildable
     let inputValidator: MSALNativeAuthInputValidating
-    private let internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
 
+    private(set) static var sharedConfiguration: MSALPublicClientApplicationConfig?
```
SammyO, line 53: We can't have static classes like this, as this prevents multiple SDK instances from running in an application in parallel. Since 1 app can only have 1 user flow, and a user flow can only have 1 auth method, we have always said that we support developers from using multiple apps (to support multiple user flows), meaning they need to have multiple SDK instances running in parallel.

cc @nilo-ms, as this an important tenet.

MSAL/src/native_auth/public/MSALNativeAuthTokenResult.swift
```diff
@@ -26,28 +26,32 @@ import Foundation
...
+        internalExpiresOn = expiresOn
     }
 
     /**
      The Access Token requested.
      Note that if access token is not returned in token response, this property will be returned as an empty string.
      */
     @objc public var accessToken: String {
-        authTokens.accessToken.accessToken
+        internalAccessToken
```
SammyO, line 44: Out of curiosity: why were these changes made?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
diegojerezba: Great catch, I believe we all overlooked this. @borgesmb, I think we can use a global variable instead of a `static` one. 

Can you test if this would work? @nilo-ms do you agree with this approach?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
borgesmb: Renamed to "convertChallengeTypes", which is more aligned with what it does.

borgesmb (commented):
.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
nilo-ms: Yes, please :)

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -75,25 +75,41 @@ import Foundation
...
-        Task {
-            let controllerResponse = await getAccessTokenInternal(
-                forceRefresh: forceRefresh,
-                correlationId: correlationId,
-                cacheAccessor: cacheAccessor
-            )
+    @objc public func getAccessToken(forceRefresh: Bool = false,
+                                     correlationId: UUID? = nil,
+                                     delegate: CredentialsDelegate) {
+        getAccessTokenInternal(forceRefresh: forceRefresh,
```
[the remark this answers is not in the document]
spetrescu84: Should we say, "without defined scopes" ? As the default ones are there I think, they get added when the parameters class gets created

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
borgesmb: Done.

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -70,30 +72,54 @@ import Foundation
...
-            case .success(let accessTokenResult):
-                await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(
-                    result: accessTokenResult,
-                    correlationId: controllerResponse.correlationId
-                )
-            case .failure(let error):
-                await delegate.onAccessTokenRetrieveError(error: error)
-            }
+    /// Retrieves the access token for the currently signed in account from the cache such that
+    /// the scope of retrieved access token is a superset of requested scopes. If the access token
```
spetrescu84, line 95: The part with "If the access token has expired, it will be refreshed using the refresh token that's stored in the cache." I don't think it should be here or if it is, should also be in the other function because the behaviour is the same. I don't think we need to explain to the user what happens under the hood.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -70,30 +72,54 @@ import Foundation
...
-            switch controllerResponse.result {
-            case .success(let accessTokenResult):
-                await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(
-                    result: accessTokenResult,
-                    correlationId: controllerResponse.correlationId
-                )
-            case .failure(let error):
-                await delegate.onAccessTokenRetrieveError(error: error)
-            }
+    /// Retrieves the access token for the currently signed in account from the cache such that
```
spetrescu84, line 94: I would change the text (because I feel it's a bit convoluted):
`/// Retrieves the access token for the currently signed in account from the cache such that the scope of retrieved access token is a superset of requested scopes.`
into 
`/// Retrieves the access token for the currently signed in account from the cache which contains the requested scopes or a superset of those.`

Also what happens if there are 2 access tokens in the cache with the following scopes:
`'openid profile offline_access scope1'`
`'openid profile offline_access scope2'`
Retrieved in this order (so the one with `scope2` is last.
Which one gets retrieved? Why?
@nilo-ms @diegojerezba what do you guys think should happen in this scenario?

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -194,7 +194,9 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
                 return .init(.failure(error), correlationId: context.correlationId())
             }
             userAccountResult.refreshData(authTokens: authTokens, account: account)
-            return .init(.success(MSALNativeAuthTokenResult(authTokens: authTokens)),
+            return .init(.success(MSALNativeAuthTokenResult(accessToken: authTokens.accessToken.accessToken,
+                                                            scopes: authTokens.accessToken.scopes?.array as? [String] ?? [],
```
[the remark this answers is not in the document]
borgesmb: MSAL ObjC allocates an ordered set, initialised with the passed scopes. So it will never be null: " scopes:[[NSOrderedSet alloc] initWithArray:parameters.scopes copyItems:YES]"

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
nilo-ms: If an external developer uses the "getAccessToken" function without any scopes, they do mind only receiving a valid access token that have default OIDC permission. In this case both cached tokens have this permission, so it's not important which one is returned.

About the suggested description: when a user specifies some scopes, it's not guaranteed that the access token returned will have all these scopes granted.

spetrescu84: I meant if they pass `openid profile offline_access` as scopes. Which one should they get.
Or let's say another case:
`openid profile offline_access scope1 scope2` stored
'openid profile offline_access scope1 scope3` stored
User retrieves `openid profile offline_access scope`. Which one should they get?

Regarding the `superset` description. Then maybe we shouldn't say they get back a superset, if that's not correct

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
borgesmb: Just pushed some changes based on an excellent tip from @spetrescu84. We are able to achieve the result without shared properties or spreading references of the client app around.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,20 +26,50 @@ import Foundation
...
+    func getAccessTokenInternal(forceRefresh: Bool = false,
+                                scopes: [String],
+                                correlationId: UUID? = nil,
+                                delegate: CredentialsDelegate) {
+
+        let params = MSALSilentTokenParameters(scopes: scopes, account: account)
+        params.forceRefresh = forceRefresh
+        params.correlationId = correlationId
+
+        guard let config = MSALNativeAuthPublicClientApplication.sharedConfiguration,
```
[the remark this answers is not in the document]
borgesmb: We got rid of this shared configuration.

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,20 +26,56 @@ import Foundation
...
+                error: RetrieveAccessTokenError(type: .generalError,
+                                                correlationId: correlationId ?? context.correlationId())) }
+            return
+        }
+
+        client.acquireTokenSilent(with: params) { result, error in
+
+            if let error = error as? NSError {
+                let accessTokenError = RetrieveAccessTokenError(type: .generalError,
+                                                                correlationId: result?.correlationId ?? UUID(),
```
nilo-ms, line 62: Can you use the context here?
```suggestion
                                                                correlationId: result?.correlationId ?? context.correlationId(),
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,20 +26,56 @@ import Foundation
...
+                let delegateDispatcher = CredentialsDelegateDispatcher(delegate: delegate, telemetryUpdate: nil)
+                let accessTokenResult = MSALNativeAuthTokenResult(accessToken: result.accessToken,
+                                                                  scopes: result.scopes,
+                                                                  expiresOn: result.expiresOn)
+                Task { await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(result: accessTokenResult, correlationId: result.correlationId) }
+                return
+            }
+
+            Task { await delegate.onAccessTokenRetrieveError(error: RetrieveAccessTokenError(type: .generalError,
+                                                                                             correlationId: correlationId ?? UUID())) }
```
nilo-ms, line 78: can you use the context here?
```suggestion
                                                                                             correlationId: correlationId ?? context.correlationId())) }
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
nilo-ms: I can't say specifically what access token will be returned. I suppose none, because scope3 isn't linked to any access token stored.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
nilo-ms: Thanks a lot Sammy for this suggestion! We completely missed it. This static configuration could create weird behaviour even when multiple nativeAuthPublicApplication instances are created (like in our sample app). 
We fixed this issue, I'll let @borgesmb explains the details :)

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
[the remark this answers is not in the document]
nilo-ms: This comment is only needed to identify when the getAccessToken without scopes is called

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
nilo-ms: @spetrescu84 can we consider this conversation resolved?

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -83,34 +83,6 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
         return nil
```
nilo-ms, line 83: Given that multiple access tokens can now exist for a single account, we should:
- get rid of the comment at line 72 because it states something that it isn't true anymore
- retrieve only IdToken  in the`retrieveUserAccountResult` function. We will retrieve Access and Refresh token only when the user requests for it.
- Now we can get rid of the `MSALNativeAuthTokens` class

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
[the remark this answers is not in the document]
spetrescu84: Ok, then my change is not needed

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
borgesmb: @nilo-ms , I already removed handleMSIDTokenResponse, handleTokenResponse, the comment on line 72 and retrieved only IdToken in `theretrieveUserAccountResult` method. I do agree that we could get rid of `MSALNativeAuthTokens` class and refactor our cache accessor, but this will be a big change, it affects dozens of existing unit test, and it seems beyond the scope of this feature. I suggest creating a PBI in our backlog for this refactor.

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
spetrescu84: I tested this with requesting `openid profile offline_access scope1` and you get back `openid profile offline_access scope1 scope2 scope3` so there will be only one token all the time

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
nilo-ms: I believe that retrieving only IdToken in the `retrieveUserAccountResult` and as consequence removing the `MSALNativeAuthTokens` is part of this ticket. This feature is to allow external developer retrieving and storing multiple access tokens in the SDK. So, this means that the user needs to specify scopes to retrieve access token. This PR makes the retrieval of the access token and refresh token as part of the account retrieval not needed anymore. So, in my opinion, this code needs to be removed in this PR.

borgesmb: Done.

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
borgesmb: Sure, done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
borgesmb: Sure, done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: This desciption is aligned with Android. If we change it here, we need to change in both. What do you think @nilo-ms , @SaurabhMSFT ?

borgesmb (commented):
.

MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
```diff
@@ -38,26 +38,17 @@ final class MSALNativeAuthCacheAccessor: MSALNativeAuthCacheInterface {
...
-                context: context)
-            let accessToken = try tokenCacheAccessor.getAccessToken(
-                forAccount: account.lookupAccountIdentifier,
-                configuration: accountConfiguration,
-                context: context)
-            return MSALNativeAuthTokens(accessToken: accessToken, refreshToken: refreshToken, rawIdToken: idToken.rawIdToken)
+            return idToken.rawIdToken
         }
 
     func getAllAccounts(configuration: MSIDConfiguration) throws -> [MSALAccount] {
```
nilo-ms, line 54: We can remove the `validateAndSaveTokensAndAccount` method too. It's used only in tests

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -29,25 +29,28 @@ import Foundation
     /// The account object that holds account information.
     @objc public var account: MSALAccount
 
-    var authTokens: MSALNativeAuthTokens
+//    var authTokens: MSALNativeAuthTokens
```
nilo-ms, line 32: nit: commented code

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -70,30 +73,54 @@ import Foundation
...
+    /// Retrieves the access token for the currently signed in account from the cache such that
+    /// the scope of retrieved access token is a superset of requested scopes. If the access token
+    /// has expired, it will be refreshed using the refresh token that's stored in the cache. If no
+    /// access token matching the requested scopes is found in cache then a new access token is fetched.
+    /// - Parameters:
+    ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
+    ///   - scopes: Permissions you want included in the access token received in the result. Not all scopes are guaranteed to be included in the access token returned.
+    ///   - correlationId: Optional. UUID to correlate this request with the server for debugging.
+    ///   - delegate: Delegate that receives callbacks for the Get Access Token flow.
+    public func getAccessToken(forceRefresh: Bool = false,
```
nilo-ms, line 104: I think default parameters should go after mandatory one. So, this should be
```suggestion
    public func getAccessToken(scopes: [String],
                                                     forceRefresh: Bool = false,
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -70,30 +73,54 @@ import Foundation
         }
     }
 
-    /// Retrieves an access token for the account.
+    /// Retrieves the access token for the default OIDC(openid, offline_access, profile) scopes from the cache.
     /// - Parameters:
     ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
```
nilo-ms, line 78:
```suggestion
    ///   - forceRefresh: Optional. Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
```

MSAL/src/native_auth/controllers/factories/MSALNativeAuthResultFactory.swift
```diff
@@ -71,33 +71,31 @@ final class MSALNativeAuthResultFactory: MSALNativeAuthResultBuildable {
                                 accountClaims: jsonDictionary)
     }
 
-    func makeAuthTokens(tokenResult: MSIDTokenResult, context: MSIDRequestContext) -> MSALNativeAuthTokens? {
+    func makeRawIdToken(tokenResult: MSIDTokenResult, context: MSIDRequestContext) -> String? {
         guard let refreshToken = tokenResult.refreshToken as? MSIDRefreshToken else {
```
nilo-ms, line 75: We don't need this guard block. Once removing this block I think we don't this factory method anymore

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -69,48 +69,19 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
...
         if let account = accounts.first {
             // We pass an empty array of scopes because that will return all tokens for that account identifier
-            // Because we expect to be only one access token per account at this point, it's ok for the array to be empty
-            guard let tokens = retrieveTokens(account: account,
-                                              scopes: [],
-                                              context: context) else {
+            guard let rawIdToken = retrieveIdToken(account: account,
+                                                   scopes: [],
+                                                   context: context) else {
                 MSALLogger.log(level: .verbose, context: nil, format: "No tokens found")
```
nilo-ms, line 75:
```suggestion
                MSALLogger.log(level: .verbose, context: nil, format: "No Id token found")
```

MSAL/test/unit/native_auth/public/MSALNativeAuthUserAccountResultTests.swift
```diff
@@ -46,25 +46,28 @@ class MSALNativeAuthUserAccountResultTests: XCTestCase {
...
+    
+//    func test_whenAccountAndTokenExist_itReturnsCorrectData() {
```
nilo-ms, line 59: Why is this method now commented? What we changed in the code in order to invalidate this test?

borgesmb (commented):
.

borgesmb (commented):
.

MSAL/test/unit/native_auth/public/MSALNativeAuthUserAccountResultTests.swift
nilo-ms: More than mocking the CacheAccessor we should make "mockable" the "acquireTokenSilent" function. A possible way to do tha is this onet:
- Create a new interface with the "acquireTokenSilenty(configuration, challengeTypes, MSALSilentTokenParameters)
- Create a new wrapper class for the MSALNativeAuthPublicClientApplication that implements the interface created in the previous point
- Inject this dependency in the MSALNativeAuthUserAccountResult class (use the class created in the previous point as default value)
- Inject a spy version of the interface created at point 1 to check how the MSALNativeAuthUserAccountResult operates when custom error/result are returned to that class

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -70,30 +73,54 @@ import Foundation
...
-                )
-            case .failure(let error):
-                await delegate.onAccessTokenRetrieveError(error: error)
-            }
+    /// Retrieves the access token for the currently signed in account from the cache such that
+    /// the scope of retrieved access token is a superset of requested scopes. If the access token
+    /// has expired, it will be refreshed using the refresh token that's stored in the cache. If no
+    /// access token matching the requested scopes is found in cache then a new access token is fetched.
+    /// - Parameters:
+    ///   - forceRefresh: Ignore any existing access token in the cache and force MSAL to get a new access token from the service.
```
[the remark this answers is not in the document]
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Done.

MSAL/src/native_auth/controllers/factories/MSALNativeAuthResultFactory.swift
borgesmb: Done.

MSAL/src/native_auth/controllers/factories/MSALNativeAuthResultFactory.swift
```diff
@@ -71,33 +71,31 @@ final class MSALNativeAuthResultFactory: MSALNativeAuthResultBuildable {
...
     }
 
     func makeUserAccountResult(tokenResult: MSIDTokenResult, context: MSIDRequestContext) -> MSALNativeAuthUserAccountResult? {
         let account =  makeAccount(tokenResult: tokenResult, context: context)
-        guard let authTokens = makeAuthTokens(tokenResult: tokenResult, context: context) else {
+        guard let rawIdToken = makeRawIdToken(tokenResult: tokenResult, context: context) else {
             MSALLogger.log(
                 level: .error,
                 context: context,
                 format: "Auth Tokens could not be created")
```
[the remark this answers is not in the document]
borgesmb: Removed this log, since the validation is not necessary anymore.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -128,14 +99,14 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
...
         context: MSALNativeAuthRequestContext
-    ) -> MSALNativeAuthTokens? {
+    ) -> String? {
         do {
             let config = factory.makeMSIDConfiguration(scopes: scopes)
-            return try cacheAccessor.getTokens(account: account, configuration: config, context: context)
+            return try cacheAccessor.getIdToken(account: account, configuration: config, context: context)
         } catch {
             MSALLogger.log(
                 level: .error,
```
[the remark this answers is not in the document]
borgesmb: Done.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
borgesmb: Done.

borgesmb (commented):
.

MSAL/src/native_auth/cache/MSALNativeAuthCacheAccessor.swift
borgesmb: This method is also used by the MSALNativeAuthTokenController (`cacheTokenResponseRetrieveTokenResult`).

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -29,25 +29,27 @@ import Foundation
...
+        rawIdToken: String?,
         configuration: MSALNativeAuthConfiguration,
         cacheAccessor: MSALNativeAuthCacheInterface
     ) {
         self.account = account
-        self.authTokens = authTokens
+        self.rawIdToken = rawIdToken
         self.configuration = configuration
         self.cacheAccessor = cacheAccessor
+        inputValidator = MSALNativeAuthInputValidator()
```
diegojerezba, line 52: In these cases we always try to use dependency injection and set a default parameter in the method's declaration, or create a `convenience init`.

```
    init(
        account: MSALAccount,
        authTokens: MSALNativeAuthTokens,
        rawIdToken: String?,
        configuration: MSALNativeAuthConfiguration,
        cacheAccessor: MSALNativeAuthCacheInterface,
        inputValidator: MSALNativeAuthInputValidating = MSALNativeAuthInputValidator()
    ) 
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -29,25 +29,27 @@ import Foundation
     /// The account object that holds account information.
     @objc public var account: MSALAccount
 
-    var authTokens: MSALNativeAuthTokens
+    var rawIdToken: String?
```
diegojerezba, line 32: Can this be private?

MSAL/test/unit/native_auth/mock/MSALNativeAuthFactoriesMocks.swift
```diff
@@ -43,16 +43,12 @@ class MSALNativeAuthResultFactoryMock: MSALNativeAuthResultBuildable {
         return makeAccount ?? MSALAccount.init(msidAccount: tokenResult.account, createTenantProfile: false)
     }
 
-    func mockMakeNativeAuthTokens(_ authTokens: MSALNativeAuthTokens) {
-        self.makeNativeAuthTokens = authTokens
+    func mockMakeNativeAuthRawIdToken(_ rawIdToken: String) {
```
diegojerezba, line 46: Both functions `mockMakeNativeAuthRawIdToken()` and `makeRawIdToken()` are not used anywhere, can we remove them (along with the variable `makeNativeRawIdToken`)?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Yes. Done.

MSAL/test/unit/native_auth/mock/MSALNativeAuthFactoriesMocks.swift
borgesmb: Done.

MSAL/src/configuration/MSALPublicClientApplicationConfig.m
```diff
@@ -130,6 +130,7 @@ - (id)copyWithZone:(NSZone *)zone
     item->_verifiedRedirectUri = [_verifiedRedirectUri copyWithZone:zone];
     item->_extraQueryParameters = [_extraQueryParameters copyWithZone:zone];
     item->_multipleCloudsSupported = _multipleCloudsSupported;
+    item->_bypassRedirectURIValidation = _bypassRedirectURIValidation;
```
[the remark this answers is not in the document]
borgesmb: I'll do it in the other PR, that I opened with changes into MSAL Objective-C. This code is here because it was merged from the other working branch, so you guys would be able to test the feature.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
borgesmb: Totally makes sense. Done.

MSAL/test/unit/native_auth/public/MSALNativeAuthUserAccountResultTests.swift
borgesmb: Ok, let's do it in another PR. Removing the commented code.

borgesmb (commented):
.

MSAL/src/native_auth/input_validator/MSALNativeAuthInputValidator.swift
```diff
@@ -26,10 +26,15 @@ import Foundation
 
 protocol MSALNativeAuthInputValidating {
     func isInputValid(_ input: String) -> Bool
+    func isInputValid(_ input: Array<Any>) -> Bool
```
nilo-ms, line 29: nit: swiftlint is complaining about this Syntactic sugar violation.
```suggestion
    func isInputValid(_ input: [Any]) -> Bool
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,57 @@ import Foundation
 
 extension MSALNativeAuthUserAccountResult {
 
-    func getAccessTokenInternal(
-        forceRefresh: Bool,
-        correlationId: UUID?,
-        cacheAccessor: MSALNativeAuthCacheInterface
-    ) async -> MSALNativeAuthCredentialsControlling.RefreshTokenCredentialControllerResponse {
+    func getAccessTokenInternal(forceRefresh: Bool = false,
```
nilo-ms, line 29: we should not use default value in this method, but default value should be used in public methods

MSAL/src/native_auth/input_validator/MSALNativeAuthInputValidator.swift
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
borgesmb: Done.

MSAL/src/native_auth/controllers/factories/MSALNativeAuthResultFactory.swift
```diff
@@ -71,33 +69,14 @@ final class MSALNativeAuthResultFactory: MSALNativeAuthResultBuildable {
...
-            return nil
-        }
-        return .init(account: account, authTokens: authTokens, configuration: config, cacheAccessor: cacheAccessor)
+        let rawIdToken = tokenResult.rawIdToken
+        return .init(account: account, rawIdToken: rawIdToken, configuration: config, cacheAccessor: cacheAccessor)
     }
 
-    func makeUserAccountResult(account: MSALAccount, authTokens: MSALNativeAuthTokens) -> MSALNativeAuthUserAccountResult? {
-        return .init(account: account, authTokens: authTokens, configuration: config, cacheAccessor: cacheAccessor)
+    func makeUserAccountResult(account: MSALAccount, rawIdToken: String?) -> MSALNativeAuthUserAccountResult? {
```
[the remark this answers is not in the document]
nilo-ms: Sorry, I overlooked the methods signature.

borgesmb (commented):
.

MSAL/src/native_auth/extension/NSError+Extensions.swift
```diff
@@ -0,0 +1,61 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+extension NSError {
+    func messageCodeAndCorrelationIdFromMSALError() -> (message: String?, code: Int?, correlationId: UUID?) {
```
nilo-ms, line 28: I think that we don't need to create a NSError extension for this "translation". This code needs to be used specifically in `MSALNativeAuthUserAccountResult+Internal` class. So, I suggest moving this code in a private method in that class. 
You could create something like:
`func createRetrieveAccessTokenError(nsError, context) -> RetrieveAccessTokenError `

borgesmb: Done.

nilo-ms:
@borgesmb can you please update the CHANGELOG.md file?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,91 @@ import Foundation
...
+
+            Task { await delegate.onAccessTokenRetrieveError(error: RetrieveAccessTokenError(type: .generalError,
+                                                                                             correlationId: correlationId ?? context.correlationId())) }
         }
     }
+    
+    func createRetrieveAccessTokenError(error: NSError, context: MSALNativeAuthRequestContext) -> RetrieveAccessTokenError {
+        if let innerError = error.userInfo[NSUnderlyingErrorKey] as? NSError,
+           let message = innerError.userInfo[MSALErrorDescriptionKey] as? String,
+           let code = codeAndCorrelationIdFromMSALError(error: innerError).code,
```
nilo-ms, line 85: if you split the `codeAndCorrelationIdFromMSALError` in two functions you can avoid retrieving twice the code and correlationId at line 86 and 85.
So, this would be 
```suggestion
           let code = codeFromMSALError(error: innerError),
           let correlationId = correlationIdFromMSALError(error: innerError)
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,91 @@ import Foundation
...
+        }
+        
+        return RetrieveAccessTokenError(type: .generalError,
+                                        message: error.localizedDescription,
+                                        correlationId: context.correlationId(),
+                                        errorCodes: [error.code])
+    }
+
+    private func codeAndCorrelationIdFromMSALError(error: NSError) -> (code: Int?, correlationId: UUID?) {
+        var errorCode: Int?
```
nilo-ms, line 103: nit: this method can be shortened with:
```suggestion
        let errorCode = error.userInfo[MSALInternalErrorCodeKey] as? Int
        let errorCorrelationId = UUID(uuidString: error.userInfo[MSALCorrelationIDKey] as? String ?? "")
        return (errorCode, errorCorrelationId)
```

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,91 @@ import Foundation
...
+                Task { await delegateDispatcher.dispatchAccessTokenRetrieveCompleted(result: accessTokenResult, correlationId: result.correlationId) }
+                return
+            }
+
+            Task { await delegate.onAccessTokenRetrieveError(error: RetrieveAccessTokenError(type: .generalError,
+                                                                                             correlationId: correlationId ?? context.correlationId())) }
         }
     }
+    
+    func createRetrieveAccessTokenError(error: NSError, context: MSALNativeAuthRequestContext) -> RetrieveAccessTokenError {
```
nilo-ms, line 82: nit: no action required here, but just giving another point of view: it's not ideal to make a private method internal to make it testable. This tell us that this class isn't testable by design, and it doesn't smell right. I think following the approach suggested [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159#discussion_r1610149398) would make this class easier to test. Let's consider this as a potential future improvement :)

borgesmb: I completely agree. When we make this class testable, this function will be private.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
borgesmb: Done.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
borgesmb: Good idea. Done.

borgesmb (commented):
.

nilo-ms (approved):

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,84 @@ import Foundation
...
+        }
+
+        return RetrieveAccessTokenError(type: .generalError,
+                                        message: error.localizedDescription,
+                                        correlationId: context.correlationId(),
+                                        errorCodes: [error.code])
+    }
+
+    private func codeFromMSALError(error: NSError) -> Int? {
+        return error.userInfo[MSALInternalErrorCodeKey] as? Int
```
nilo-ms, line 103: when we parse this error, we're not returning the API error code as we usually do, but the internal MSAL Objective-C code. For example, In the case of the API error AADSTS28000 ("Provided value for the input parameter scope is not valid because it contains more than one resource.") we return -42002 that it's an internal error. Or we parse the error code from the error message, or we just return an empty error code. So, we don't return a misleading information

borgesmb: Removing error code from MSAL internal errors.

borgesmb (commented):
.

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,83 @@ import Foundation
...
-        self.account = account
+    func createRetrieveAccessTokenError(error: NSError, context: MSALNativeAuthRequestContext) -> RetrieveAccessTokenError {
+        if let innerError = error.userInfo[NSUnderlyingErrorKey] as? NSError,
+           let message = innerError.userInfo[MSALErrorDescriptionKey] as? String,
+           let correlationId = correlationIdFromMSALError(error: innerError) {
+            return RetrieveAccessTokenError(type: .generalError, message: message, correlationId: correlationId, errorCodes: [])
+        }
+
+        if let message = error.userInfo[MSALErrorDescriptionKey] as? String,
+            let code = codeFromMSALError(error: error),
```
nilo-ms, line 90: is "MSALInternalErrorCodeKey" always going to return an internal error code? Can you please double check. If the answer is yes, we should not return this code in our error codes

MSAL/test/unit/native_auth/input_validator/MSALNativeAuthInputValidatorTest.swift
```diff
@@ -39,4 +39,15 @@ final class MSALNativeAuthInputValidatorTest: XCTestCase {
     func testInput_whenInvalidStringInputIsUsed_resultShouldBeInvalid() {
         XCTAssertFalse(validator.isInputValid(""))
     }
+    
+    func testInput_whenValidArrayInputIsUsed_resultShouldBeValid() {
+        XCTAssertTrue(validator.isInputValid(["email@contoso.com"]))
+        XCTAssertTrue(validator.isInputValid(["email", "password"]))
+        XCTAssertTrue(validator.isInputValid([1, 2, 3]))
+        XCTAssertTrue(validator.isInputValid([""]))
```
diegojerezba, line 47: One question, do we want `[""]` to be considered a valid input from the user?

MSAL/test/unit/native_auth/public/MSALNativeAuthUserAccountResultTests.swift
```diff
@@ -46,30 +87,55 @@ class MSALNativeAuthUserAccountResultTests: XCTestCase {
...
+    func test_errorWithoutInnerError() {
+        let contextCorrelationId = UUID()
+        let context = MSALNativeAuthRequestContext(correlationId: contextCorrelationId)
+        
+        let result = sut.createRetrieveAccessTokenError(error: errorWithoutInnerErrorMock,
+                                                        context: context)
+        
+        XCTAssertEqual(result.errorDescription, "user_info_error_description")
+        XCTAssertEqual(result.errorCodes, [-3003])
+        XCTAssertEqual(result.correlationId.uuidString, withoutInnerCorrelationId)
```
diegojerezba, line 127: There's something I don't understand from here. The context is created in line 120.
Shouldn't the assert in the line 127 be equal to the context of line 120?

borgesmb: We get the correlation id from the error. If not possible, we get it from the context. Correct me if I'm wrong @nilo-ms.

MSAL/test/unit/native_auth/input_validator/MSALNativeAuthInputValidatorTest.swift
borgesmb: Good question. The method is generic, it only takes into account the existence of items in the array, not its contents. That's why I tested it even with numbers. I can change it to receive an array of strings, and also check if the items are empty strings or not (this validation is correct for scopes, since MSAL ignores empty strings in the list). In this case the method won't be generic, but we can create new methods in the future if we need to validate other input types. What do you think @diegojerezba ?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
borgesmb: "MSALInternalErrorCodeKey" is the same for error codes, or inner error codes (although the name can be misleading). In line 90 it is going to get the error code (because it doesn't have an inner error, otherwise the function would have returned in line 86).

borgesmb (commented):
.

MSAL/test/unit/native_auth/input_validator/MSALNativeAuthInputValidatorTest.swift
diegojerezba: I think it's fine the way it is now, thanks for the clarification 👍

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -26,25 +26,83 @@ import Foundation
...
+
+            Task { await delegate.onAccessTokenRetrieveError(error: RetrieveAccessTokenError(type: .generalError,
+                                                                                             correlationId: correlationId ?? context.correlationId())) }
         }
     }
 
-    func refreshData(authTokens: MSALNativeAuthTokens, account: MSALAccount) {
-        self.authTokens = authTokens
-        self.account = account
+    func createRetrieveAccessTokenError(error: NSError, context: MSALNativeAuthRequestContext) -> RetrieveAccessTokenError {
```
diegojerezba, line 82: Coming from this [comment](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159#discussion_r1621194657), I have a couple questions about this function. It does three attempts to get the error:

1. Get the error from error.userInfo[NSUnderlyingErrorKey] + message + correlationId.
2. If any of the 1) fails, it tries to get the message + code + correlationId.
3. If any of the 1) fails, it creates a generic error.

Here are the questions:

- Why the code is not retrieved in 1)?
- Why the code in ⁠2) differs from the code in 3)? (I think this is the same as this [comment](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159/files#r1621077026))
    - In 2) is error.userInfo[MSALInternalErrorCodeKey]
    - In 3) is error.code
- In both 1) and 2) we still can get an empty correlationId (the function `correlationIdFromMSALError()` can return an empty String. Shouldn't we pass the context to that function and return the `context.correlationId` in case the one from the api error is not available?

MSAL/test/unit/native_auth/public/MSALNativeAuthUserAccountResultTests.swift
diegojerezba: I'm taking a look at the implementation and I've written a [comment](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2159/files#r1622184786) related to this one there, so I'll resolve this to avoid duplication.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -128,86 +99,21 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
...
                 context: context,
-                format: "Error retrieving tokens: \(error)"
+                format: "Error retrieving IdToken"
```
spetrescu84, line 114: We've lost the error here as were are no longer logging it?

MSAL/src/native_auth/controllers/factories/MSALNativeAuthResultFactory.swift
```diff
@@ -71,33 +69,14 @@ final class MSALNativeAuthResultFactory: MSALNativeAuthResultBuildable {
...
         let account =  makeAccount(tokenResult: tokenResult, context: context)
-        guard let authTokens = makeAuthTokens(tokenResult: tokenResult, context: context) else {
-            MSALLogger.log(
-                level: .error,
-                context: context,
-                format: "Auth Tokens could not be created")
-            return nil
-        }
-        return .init(account: account, authTokens: authTokens, configuration: config, cacheAccessor: cacheAccessor)
+        let rawIdToken = tokenResult.rawIdToken
```
spetrescu84, line 74: Do we need this here? We can just have tokenResult.rawIdToken in the next line

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
```diff
@@ -119,4 +119,21 @@ extension MSALNativeAuthPublicClientApplication {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
     }
+
+    static func convertChallengeTypes(
```
spetrescu84, line 123: Looking at these 2 functions, I think we took the paradigm from `MSALPublicClientApplication` but should they even be here? They don't use anything from the `MSALNativeAuthPublicClientApplication`, they could just as easily go into the Enums themselves like so:
Screenshot 2024-06-04 at 15 36 34

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
nilo-ms: Yes, we decided to not logged it because it can contain sensitive information

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
nilo-ms: Given that these functions are now used by more than one class, they should go in a dedicated class (ex: challengeTypeConverter). I don't think that moving them to the MSALNativeAuthInternalChallengeType enum will make the situation better. I'd say to leave those functions in the publicClientApplication for now

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
nilo-ms: If this key can also be used also for internal error, we should not return them as part of our "errorCodes" array. I'm going to remove the parse of the error code

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
spetrescu84: Ok

spetrescu84 (approved):

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
nilo-ms: Me and Diego discussed this comment in a call. What we clarified:
- the code isn't retrieved in 1) because `MSALInternalErrorCodeKey` contains only error codes generated internally by the SDK. We should return only error codes related to eSTS
- We should not parse the error in this case. Code updated
- `correlationIdFromMSALError()` function can return an UUID nil (UUID of an empty string is nil)

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult+Internal.swift
nilo-ms: If "MSALInternalErrorCodeKey" can contains internal error, we should parse it. Code updated

diegojerezba (approved):

MSAL/src/native_auth/controllers/factories/MSALNativeAuthResultFactory.swift
nilo-ms: Updated ;)

PR #2181, opened 2024-05-29
Add SPM validation in each PR

diegojerezba:
## Proposed changes

This PR adds a new check in the pr-validation.yml. It consists in an integration of the changes of the current branch into the [NativeAuth Sample App](https://github.com/Azure-Samples/ms-identity-ciam-native-auth-ios-sample) using a Swift Package.

These are the steps that are done:

- Creates MSAL.zip of the xcframework.
	- It configures the URL and the checksum so the Sample App will use the changes we want to test.
- Pushes it to a temporary branch.
- Clones the NativeAuth Sample App repo, and configures it to download the Swift Package from the temporary branch.
- Builds the Sample App.
- Cleans up.

Note: I have checked the `build.py` script and I think it's different enough from what we are trying to do here, so I thought it is better to have two separate jobs, even if both scripts build the framework. I'm open to discussion on this.

The time it takes the PR-validation hasn't been incremented, as it runs in parallel and this job takes ~6-10 min.

The following tests have been done:

- Test 1: (Pre)configure the sample app to use a new public method published in the current branch of MSAL.
- Test 2: Make a breaking change in MSAL, the Sample App shouldn't work.
- Test 3: If any build fails, the pipeline fails (and the temporary branch is deleted).
- Test 4: If for any reason the checksum calculation doesn't work, the pipeline fails (and the temporary branch is deleted).

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [x ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -98,7 +98,7 @@ jobs:
     clean: true
     submodules: true
     fetchDepth: 1
-    persistCredentials: true
+    persistCredentials: false
     path: s
 
   - script: |
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -110,8 +110,7 @@ jobs:
     inputs:
       targetType: 'inline'
       script: |
-        UUID_LOCAL=$(uuidgen)
-        BRANCH_NAME_LOCAL="$(Build.SourceBranchName)-temp-${UUID_LOCAL}"
+        BRANCH_NAME_LOCAL="$(Build.SourceBranchName)-temp"
         echo "##vso[task.setvariable variable=BRANCH_NAME]${BRANCH_NAME_LOCAL}"
       
   - task: Bash@3
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -98,7 +98,7 @@ jobs:
     clean: true
     submodules: true
     fetchDepth: 1
-    persistCredentials: false
+    persistCredentials: true
     path: s
 
   - script: |
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -127,9 +127,10 @@ jobs:
       targetType: 'inline'
       script: |
         sh spm-integration-test.sh "${BRANCH_NAME}"
-    continueOnError: true
+    continueOnError: false
 
   - task: Bash@3
+    condition: always()
     displayName: Cleanup
     inputs:
       targetType: 'inline'
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -83,4 +83,31 @@ jobs:
       testResultsFiles: '$(Agent.BuildDirectory)/s/build/reports/*'
       failTaskOnFailedTests: true
       testRunTitle: 'Test Run - $(target)'
-  
+
+- job: 'Validate_SPM_Integration'
+  displayName: Validate SPM Integration
+  pool:
+    vmImage: 'macOS-13'
+    timeOutInMinutes: 15
+  workspace:
+    clean: all
+
+  steps:
+
+  - checkout: self
+    clean: true
+    submodules: true
+    fetchDepth: 1
+    persistCredentials: true
+    path: s
+
+  - script: |
+      /bin/bash -c "sudo xcode-select -s /Applications/Xcode_15.0.1.app"
+    displayName: 'Switch to use Xcode 15.0.1'
+
+  - task: Bash@3
+    displayName: Run SPM integration test script
+    inputs:
+      targetType: 'inline'
+      script: 
+        sh spm-integration-test.sh "$(Build.SourceBranchName)"
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -102,12 +102,43 @@ jobs:
     path: s
 
   - script: |
-      /bin/bash -c "sudo xcode-select -s /Applications/Xcode_15.0.1.app"
-    displayName: 'Switch to use Xcode 15.0.1'
+        /bin/bash -c "sudo xcode-select -s /Applications/Xcode_14.3.app"
+    displayName: 'Switch to use Xcode 14.3'
+
+  - task: Bash@3
+    displayName: Set variable BRANCH_NAME to a temporary branch
+    inputs:
+      targetType: 'inline'
+      script: |
+        UUID_LOCAL=$(uuidgen)
+        BRANCH_NAME_LOCAL="$(Build.SourceBranchName)-temp-${UUID_LOCAL}"
+        echo "##vso[task.setvariable variable=BRANCH_NAME]${BRANCH_NAME_LOCAL}"
+      
+  - task: Bash@3
+    displayName: Checkout to temporary branch
+    inputs:
+      targetType: 'inline'
+      script: |
+        git checkout -b "${BRANCH_NAME}"
 
   - task: Bash@3
     displayName: Run SPM integration test script
     inputs:
       targetType: 'inline'
+      script: |
+        sh spm-integration-test.sh "${BRANCH_NAME}"
+    continueOnError: true
+
+  - task: Bash@3
+    displayName: Cleanup
+    inputs:
+      targetType: 'inline'
       script: 
-        sh spm-integration-test.sh "$(Build.SourceBranchName)"
+        cd ../..
+        rm -rf "$SAMPLE_APP_TEMP_DIR" archive framework MSAL.zip
+        git checkout -- .
+        git fetch --quiet
+        git switch "$(Build.SourceBranchName)"
+        git branch -D "$BRANCH_NAME"
+        git push origin --delete "$BRANCH_NAME"
+
--- a/spm-integration-test.sh
+++ b/spm-integration-test.sh
@@ -78,8 +78,8 @@ cd ../..
 rm -rf "$SAMPLE_APP_TEMP_DIR" archive framework MSAL.zip
 
 git checkout -- .
-git fetch
-git switch dev
+git fetch --quiet
+git switch "$1"
 
 git branch -D "$BRANCH_NAME"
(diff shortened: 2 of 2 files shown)


azure_pipelines/pr-validation.yml
```diff
@@ -83,4 +83,31 @@ jobs:
...
+
+  - checkout: self
+    clean: true
+    submodules: true
+    fetchDepth: 1
+    persistCredentials: true
+    path: s
+
+  - script: |
+      /bin/bash -c "sudo xcode-select -s /Applications/Xcode_15.0.1.app"
```
nilo-ms, line 105: nit: is there a reason we're using Xcode 15 while in the previous job we use Xcode 14?

spm-integration-test.sh
```diff
@@ -0,0 +1,90 @@
...
+BRANCH_NAME="$1-temp-${UUID}"
+SAMPLE_APP_TEMP_DIR="NativeAuthSampleAppTemp"
+current_date=$(date +"%Y-%m-%d %H:%M:%S")
+
+# Build framework
+
+git checkout -b "$BRANCH_NAME"
+rm -rf archive framework MSAL.zip
+
+xcodebuild -sdk iphonesimulator -configuration Release -workspace MSAL.xcworkspace -scheme "MSAL (iOS Framework)" archive SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES -archivePath archive/iOSSimulator CODE_SIGNING_ALLOWED=NO -quiet > build.log 2>&1
```
nilo-ms, line 11: can you print something to explain what each task is doing?

spm-integration-test.sh
```diff
@@ -0,0 +1,90 @@
...
+if [ $XCBUILD_STATUS_SIM -ne 0 ]; then
+  echo "** Build xcframework Error **"
+  exit 1
+fi
+
+xcodebuild -sdk iphoneos -configuration Release -workspace MSAL.xcworkspace -scheme "MSAL (iOS Framework)" archive SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES -archivePath archive/iOS CODE_SIGNING_ALLOWED=NO -quiet > build.log 2>&1
+XCBUILD_STATUS_IPHONE=$?
+if [ $XCBUILD_STATUS_IPHONE -ne 0 ]; then
+  echo "** Build xcframework Error **"
+  exit 1
```
nilo-ms, line 22: should remove the folders created before exit?

spetrescu84: As long as the machine shuts down after the test it's not necessary as it gets cleaned up, but what needs to happen for sure is that the machine has to shut down. I know there is a way to keep it on so the tests use previous data but that should not be the case for this validation

spm-integration-test.sh
```diff
@@ -0,0 +1,90 @@
...
+fi
+
+zip -r MSAL.zip framework/MSAL.xcframework -y -v
+CHECKSUM=$(swift package compute-checksum MSAL.zip)
+if [ -z "$CHECKSUM" ]; then
+  echo "** Checksum could not be obtained **"
+  exit 1
+fi
+
+NEW_URL="https://github.com/AzureAD/microsoft-authentication-library-for-objc/raw/$BRANCH_NAME/MSAL.zip/"
```
spetrescu84, line 46: Is the `/` after `MSAL.zip/` needed?

spm-integration-test.sh
```diff
@@ -0,0 +1,90 @@
...
+git clone https://github.com/Azure-Samples/ms-identity-ciam-native-auth-ios-sample.git
+cd ms-identity-ciam-native-auth-ios-sample
+
+sed -i '' 's#kind = upToNextMinorVersion;#kind = branch;#' NativeAuthSampleApp.xcodeproj/project.pbxproj
+sed -i '' "s#minimumVersion = [0-9.]*;#branch = $BRANCH_NAME;#" NativeAuthSampleApp.xcodeproj/project.pbxproj
+
+rm -f NativeAuthSampleApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+
+xcodebuild -resolvePackageDependencies
+xcodebuild -scheme NativeAuthSampleApp -configuration Release -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 14,OS=latest' clean build
```
spetrescu84, line 72: The `OS=latest` could cause problems down the line, it has before. maybe set it to a specific version like 17.0

spm-integration-test.sh
```diff
@@ -0,0 +1,90 @@
...
+
+cd ../..
+rm -rf "$SAMPLE_APP_TEMP_DIR" archive framework MSAL.zip
+
+git checkout -- .
+git fetch --quiet
+git switch "$1"
+
+git branch -D "$BRANCH_NAME"
+git push origin --delete "$BRANCH_NAME"
```
spetrescu84, line 85: So if the script fails, the branches don't get deleted. Should we delete the branches all the time?

azure_pipelines/pr-validation.yml
diegojerezba: not really, I've adjusted it to Xcode 14

spm-integration-test.sh
diegojerezba: Done

spm-integration-test.sh
diegojerezba: I have updated the job. Now it does the cleanup in a different step after having run the script. The structure is similar to this:

- Step: select Xcode
- Step: Run script, with `continueOnError: true`, so it continues to the next step even if the scrip fails.
- Step: Cleanup

[Here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1313778&view=logs&j=1494a728-882c-596b-7b5c-9fa51f14f0ef) there's an example of the cleanup if the SPM scrip fails:

spm-integration-test.sh
diegojerezba: Removed

spm-integration-test.sh
diegojerezba: See [this comment](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2181#discussion_r1635277508).

spm-integration-test.sh
diegojerezba: Thank you for the advice, I've updated it to the latest possible version that runs with Xcode 14.3 (iOS 16.4)

spetrescu84 (approved):

nilo-ms (approved):

azure_pipelines/pr-validation.yml
```diff
@@ -83,4 +83,62 @@ jobs:
...
+      script: |
+        git checkout -b "${BRANCH_NAME}"
+
+  - task: Bash@3
+    displayName: Run SPM integration test script
+    inputs:
+      targetType: 'inline'
+      script: |
+        sh spm-integration-test.sh "${BRANCH_NAME}"
+    continueOnError: true
```
ameyapat, line 130: Should this be continueOnError: false? How will errors during build with this branch's MSAL swift package be caught and cause the pipeline to fail?

azure_pipelines/pr-validation.yml
```diff
@@ -83,4 +83,62 @@ jobs:
...
+
+  - task: Bash@3
+    displayName: Cleanup
```
ameyapat, line 133: Add condition: always() for this task so that it always runs?

azure_pipelines/pr-validation.yml
```diff
@@ -83,4 +83,62 @@ jobs:
...
+  - script: |
+        /bin/bash -c "sudo xcode-select -s /Applications/Xcode_14.3.app"
+    displayName: 'Switch to use Xcode 14.3'
+
+  - task: Bash@3
+    displayName: Set variable BRANCH_NAME to a temporary branch
+    inputs:
+      targetType: 'inline'
+      script: |
+        UUID_LOCAL=$(uuidgen)
```
ameyapat, line 113: nit: We don't need to create a UUID guid if the branch is going to be deleted in the last step anyways. We can create and delete a branch with the same name

ameyapat:
Added a few comments. Otherwise, lgtm

azure_pipelines/pr-validation.yml
diegojerezba: You're right, thanks for the catch.

azure_pipelines/pr-validation.yml
diegojerezba: Absolutely, thank you.

azure_pipelines/pr-validation.yml
diegojerezba: Done

azure_pipelines/pr-validation.yml
```diff
@@ -83,4 +83,62 @@ jobs:
...
+  workspace:
+    clean: all
+
+  steps:
+
+  - checkout: self
+    clean: true
+    submodules: true
+    fetchDepth: 1
+    persistCredentials: true
```
nilo-ms, line 101: Should this be false?

ameyapat (approved):
I'd suggest to manually run the pipeline on this branch as a sanity check

azure_pipelines/pr-validation.yml
diegojerezba: I tried, but got this [error](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1320612&view=logs&j=1494a728-882c-596b-7b5c-9fa51f14f0ef&t=7be9455c-7140-52f8-03cd-b09f3cf00ddb): `fatal: could not read Username for 'https://github.com'/: terminal prompts disabled`

The [documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines) says that it is necessary in order to leave the OAuth token in the Git config after the initial fetch.

diegojerezba:
> I'd suggest to manually run the pipeline on this branch as a sanity check

Done: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1320628&view=results

PR #2217, opened 2024-06-21
Native auth base E2E tests and 6 test cases

nilo-ms:
## Proposed changes

This PR includes the basic code for end-to-end tests of native authentication and six test cases. The tests run concurrently and can retry up to three times if any of them fails.

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [X] Test
- [ ] Logging/Telemetry

## Risk


MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -26,54 +26,65 @@ import XCTest
...
-            correlationId: correlationId,
-            responses: [response]
-        )
+    class Constants {
+        static let nativeAuthKey = "native_auth"
+        static let clientIdEmailPasswordKey = "email_password_client_id"
+        static let clientIdEmailCodeKey = "email_code_client_id"
+        static let clientIdEmailPasswordAttributesKey = "email_password_attributes_client_id"
+        static let clientIdEmailCodeAttributesKey = "email_code_attributes_client_id"
+        static let tenantSubdomainKey = "tenant_subdomain"
```
spetrescu84, line 35: We need to make sure to remove this domain and others that we might use after from the Kusto query on the OCE dashboard

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -26,54 +26,65 @@ import XCTest
...
-        if usingMockAPI {
-            print("🤖🤖🤖 Using mock API: \(Configuration.authorityURLString)")
-        } else {
-            print("👩‍💻👩‍💻👩‍💻 Using test tenant: \(Configuration.authorityURLString)")
+        let confFile = try? JSONSerialization.jsonObject(with: configurationData, options: []) as? [String: Any]
+        confFileContent = confFile?[Constants.nativeAuthKey] as? [String: String]
+    }
+    
+    func initialisePublicClientApplication(
+        useEmailPasswordClientId: Bool = true,
```
spetrescu84, line 59: Should we have these 2 bools concatenated into an enum as maybe in the future we want even more cases?
The `if` below might grow quite a bit.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -26,54 +26,65 @@ import XCTest
...
+        guard let clientId = confFileContent?[clientIdKey] as? String, let tenantSubdomain = confFileContent?[Constants.tenantSubdomainKey] as? String else {
+            XCTFail("ClientId or tenantSubdomain not found in conf.json")
+            return nil
         }
+        return try? MSALNativeAuthPublicClientApplication(clientId: clientId, tenantSubdomain: tenantSubdomain, challengeTypes: challengeTypes)
+    }
+    
+    func generateSignUpRandomEmail() -> String {
+        let randomId = UUID().uuidString.prefix(8)
+        return "native-auth-signup-\(randomId)@1secmail.org"
```
spetrescu84, line 80: Emails are with `.org` and API is with `.com`?

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
```diff
@@ -0,0 +1,100 @@
...
+        var request = URLRequest(url: url)
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        do {
+            let (data, response) = try await URLSession.shared.data(for: request)
+
+            guard let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode == 200 else {
+                return nil
+            }
+            let dataDictionary = try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
+            return (dataDictionary?["textBody"] as? String)?.components(separatedBy: CharacterSet.newlines).first(where: {$0.range(of: "[0-9]{4,}$", options: .regularExpression, range: nil, locale: nil) != nil })
```
spetrescu84, line 94: This is next to impossible to read or debug so if the position of the OTP changes it gets really complex to change this.
Maybe separate on multiple lines and add a comment?

MSAL/NativeAuthEndToEndTestPlan.xctestplan
```diff
@@ -62,6 +62,8 @@
     }
```
spetrescu84, line 62: I think the Mock API test plan configuration should be removed

MSAL/test/integration/native_auth/end_to_end/sign_in/MSALNativeAuthSignInUsernameEndToEndTests.swift
```diff
@@ -118,131 +97,23 @@ final class MSALNativeAuthSignInUsernameEndToEndTests: MSALNativeAuthEndToEndBas
         let username = ProcessInfo.processInfo.environment["existingOTPUserEmail"] ?? "<existingOTPUserEmail not set>"
```
spetrescu84, line 97: Should this also be changed to `let username = "missing email+code signIn username"`?

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: I think the lab tenant is already removed from our and eSTS dashboard.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Can you clarify how the enum will looks like? I didn't fully get it

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Correct

MSAL/NativeAuthEndToEndTestPlan.xctestplan
nilo-ms: No, it should not. I already created a new schema dedicated to E2E tests in the branch where I want to run those tests in the pipeline. I can do this change here too

MSAL/test/integration/native_auth/end_to_end/sign_in/MSALNativeAuthSignInUsernameEndToEndTests.swift
nilo-ms: All the skipped tests need to be revisited once enabled. So, I'd say to do this once we enable them

MSAL/test/integration/native_auth/end_to_end/sign_up/MSALNativeAuthSignUpUsernameEndToEndTests.swift
```diff
@@ -27,39 +27,37 @@ import MSAL
...
+            return
+        }
         let codeRequiredExp = expectation(description: "code required")
         let signUpStartDelegate = SignUpStartDelegateSpy(expectation: codeRequiredExp)
-
-        if usingMockAPI {
-            try await mockResponse(.signUpStartSuccess, endpoint: .signUpStart)
-            try await mockResponse(.challengeTypeOOB, endpoint: .signUpChallenge)
-        }
+        let usernameOTP = generateSignUpRandomEmail()
```
diegojerezba, line 40: From what I've tested this particular test, I'm getting ~1 failure out of 10 tries. If we multiply this for every test, the possibility of getting random failures is a bit high.

As we've talked this is due to OTP codes getting stuck on the Microsoft queue.

Right now if a test fails, we try again (up to 3 times), but this generates another email. It's the equivalent to a user that can't get the OTP trying to sign up again with a different email.

Do you know if we can simulate a user clicking on the "resend code" button instead of signing up with a new email?

I'm not sure how this works on the backend side, but (as a user) I've always thought that when I click on resend code it "forces" the backend to send it.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
diegojerezba, line 59: I wouldn't set a default value to `useEmailPasswordClientId: Bool = true,`, as we have an entirely set of tests (the emailAndCode ones) for which this is false.

In fact, I would use an enum rather an an boolean. I think this is what @spetrescu84 is saying in [his comment](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2217/files#r1654979578).

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -26,54 +26,65 @@ import XCTest
...
+        confFileContent = confFile?[Constants.nativeAuthKey] as? [String: String]
+    }
+    
+    func initialisePublicClientApplication(
+        useEmailPasswordClientId: Bool = true,
+        attributeRequiredClientId: Bool = false,
+        challengeTypes: MSALNativeAuthChallengeTypes = [.OOB, .password]
+    ) -> MSALNativeAuthPublicClientApplication? {
+        var clientIdKey = Constants.clientIdEmailPasswordKey
+        if useEmailPasswordClientId && attributeRequiredClientId {
```
diegojerezba, line 64: Is not possible to have `useEmailPasswordClientId && !attributeRequiredClientId`?

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -26,54 +26,65 @@ import XCTest
...
-                authority: authority
-            ),
-            challengeTypes: [.OOB, .password]
-        )
-
-        if usingMockAPI {
-            print("🤖🤖🤖 Using mock API: \(Configuration.authorityURLString)")
-        } else {
-            print("👩‍💻👩‍💻👩‍💻 Using test tenant: \(Configuration.authorityURLString)")
+        let confFile = try? JSONSerialization.jsonObject(with: configurationData, options: []) as? [String: Any]
```
diegojerezba, line 54: I would deserialize the `configurationData` catching the error because it can help us to understand possible errors in the future.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -26,54 +26,65 @@ import XCTest
...
+        static let clientIdEmailPasswordAttributesKey = "email_password_attributes_client_id"
+        static let clientIdEmailCodeAttributesKey = "email_code_attributes_client_id"
+        static let tenantSubdomainKey = "tenant_subdomain"
+        static let signInEmailPasswordUsernameKey = "sign_in_email_password_username"
+        static let signInEmailCodeUsernameKey = "sign_in_email_code_username"
+        static let resetPasswordUsernameKey = "reset_password_username"
     }
+    
+    let correlationId = UUID()
+    let defaultTimeout: TimeInterval = 20
```
diegojerezba, line 42: Are we synced with Android on the default TimeOut?

MSAL/MSAL.xcodeproj/project.pbxproj
```diff
@@ -5230,6 +5298,26 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		28A277CD2C21F0B400D95E00 /* Copy conf.json file */ = {
```
diegojerezba, line 5301: Is this expected to be here?

MSAL/test/integration/native_auth/end_to_end/sign_up/MSALNativeAuthSignUpUsernameEndToEndTests.swift
nilo-ms: It can be, but we need to double check this with the API team before to make this assumption. I personally would prefer to rerun the test for the following reason:
- it doesn't affect the test code. If we ask to the SDK to resend the code, we're interacting with the SDK under test in a way that it is not part of the actual test.
- using the test config is pretty easy to setup the retry logic and we don't need to add new code
- each time we rerun the tests, a new email is created, making it highly unlikely that a new email won't be sent to this address.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Can you please suggest how this will look in the code? If we have to define all scenarios as enum cases, the enum might become quite large. For instance, we would need to include 4 use cases for this particular scenario. However, it's possible I might have misunderstood your suggestion.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Yeah, but this is the default case, so I didn't add an else case

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: I don't see a reason why we should be in sync 🧐

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Are you suggesting failing the test if the JSONSerialization fails?

MSAL/MSAL.xcodeproj/project.pbxproj
nilo-ms: Yeah, this is the script used to copy the conf.json

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
```diff
@@ -0,0 +1,102 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.  
+
+import XCTest
+
+class MSALNativeAuthEmailCodeRetriever: XCTestCase {
+
+    private let baseURLString = "https://www.1secmail.com/api/v1/?action="
```
diegojerezba, line 29: Nit: the references to the email provider are spread out (here and in `MSALNativeAuthEndToEndBaseTestCase`), maybe we could have them in one `Constants` place, in case we want to change the provider at some point?

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
```diff
@@ -0,0 +1,102 @@
...
+        try? await Task.sleep(nanoseconds: UInt64(secondsToWait * Double(NSEC_PER_SEC)))
+        var request = URLRequest(url: url)
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        do {
+            let (data, response) = try await URLSession.shared.data(for: request)
+            
+            guard let httpResponse = response as? HTTPURLResponse,
+                    httpResponse.statusCode == 200,
+                    var dataDictionary = try JSONSerialization.jsonObject(with: data, options: []) as? [[String: Any]] else {
+                return nil
```
diegojerezba, line 66: If during the pipeline a test fails, the engineer will look at the logs to see what happened. I think it would be useful to print rich logs. 

For example, a test might fail because the 1sec service is performing some throttling. Or because their service is down. I think printing the status code in case of failure might be interesting.

Also as we discussed before, the email can be empty, in which case is a Microsoft issue on delivering the email OTP. You can add that info to the print (if the body is an empty array, say that Microsoft didn't delivered the email OTP) so we will know exactly why that test failed.

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
```diff
@@ -0,0 +1,102 @@
...
+            }
+            if dataDictionary.count > 0 {
+                dataDictionary.sort(by: {($0["id"] as? Int ?? 0) > ($1["id"] as? Int ?? 0)})
```
diegojerezba, line 69: nit: I think the readability of this code would be better if we used `Decodable` rather than handling an array of dictionaries `[[String: Any]]`.

But given that this code will not be looked at/changed often, I'm also okay with the current approach. Feel free to discard this comment if you consider it's not worth the effort.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
diegojerezba: I think it may be useful to have the same configuration when we interact with the same third party dependency.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
diegojerezba: Yes, it's a rare case, but it is just to fail as soon as possible, rather than waiting for when the confFile is used to find out that is nil because it couldn't be deserialized.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
diegojerezba: I think we can just differentiate between email otp and email with password for now.

```
enum Userflow {
    case emailOtp
    case emailPassword
}
```

I named it `UserFlow` to match what is in the portal, but feel free to use another name.

MSAL/test/integration/native_auth/end_to_end/sign_up/MSALNativeAuthSignUpUsernameEndToEndTests.swift
diegojerezba: > it doesn't affect the test code. If we ask to the SDK to resend the code, we're interacting with the SDK under test in a way that it is not part of the actual test.

This is a good point, and it's very valid, but I'm of the opinion that an end to end test should handle a full flow and simulate a real world situation.

> using the test config is pretty easy to setup the retry logic and we don't need to add new code

Another good point, I agree that my suggestion will increase the complexity.

> each time we rerun the tests, a new email is created, making it highly unlikely that a new email won't be sent to this address.

I don't agree with this, if the email is delayed once, I don't understand why signing up with a different email would make the other email be sent on time.

Anyway, all of this is based on the assumption that the API forces the "pending emails" to be sent if a user clicks on "resend code". I would ask the API team first, and if we see the complexity increases too much, we can first monitor the current solution for some time. If we see we have multiple failures for this reason, we can implement it then.

nilo-ms: > but I'm of the opinion that an end to end test should handle a full flow and simulate a real world situation.
The test scenario should be designed for consistent replication, ensuring it evaluates the same behavior each time. Introducing additional steps for certain situations go against this requirement.
> don't agree with this, if the email is delayed once, I don't understand why signing up with a different email would make the other email be sent on time.
The issue isn't with the entire email OTP service being non-functional; it's specific to a particular email address where the process seems to halt. From what I've observed, initiating the signup using an alternative email address resolves the problem.
If we notice that we have many test emails don't receive the email OTP on time, it might be good to consult eSTS for further info, as this could potentially impact our customers.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Thanks for the suggestion. Let's wait for Silviu's input because he suggested something similar in a previous comment.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: Sure, I'll change it

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
nilo-ms: The idea was to have everything related to the email retriever in one single class. Currently, the base class contains just one element related to 1secmail, which is the signUp email creator. I can move that method to this class.

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
nilo-ms: I think using decodable is too much overengineering for this code. Resolving this comment

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
nilo-ms: The goal is that a test should never fails for something related to email OTP code retriever. If an email OTP is not found the test is retried with a different email address. 
I added some basic log in the `retrieveEmailOTPCode` method. I didn't add logs in this method because it can be executed up to three times for each test execution, so I didn't log anything to don't create too much noise. 
We can talk more about it in a call if you want.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
spetrescu84: Something like this:
```
enum ClientIdType {
        case password
        case passwordAndAttributes
        case emailCode
        case emailCodeAndAttributes
    }
```
Or like Diego's [comment](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2217/files/2cda8d67f7af1438645960768ddb8162c4492f45#r1659027154)

MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/MSALNativeAuthEmailCodeRetriever.swift
```diff
@@ -0,0 +1,115 @@
...
+            }
+            guard var dataDictionary = try JSONSerialization.jsonObject(with: data, options: []) as? [[String: Any]] else {
+                return nil
+            }
+            if dataDictionary.count > 0 {
+                dataDictionary.sort(by: {($0["id"] as? Int ?? 0) > ($1["id"] as? Int ?? 0)})
+                return dataDictionary.first?["id"] as? Int
+            } else {
+                // log only for the final retry
+                if (retryCounter == 1) {
```
diegojerezba, line 82: Can you do this check comparing it to the `numberOfRetry` variable? Just in case we decide to update the number of retries in the future, to avoid changing it here as well.

diegojerezba: Oh right, sorry my bad! 🤦‍♂️

diegojerezba (approved):

spetrescu84 (approved):

MSAL/test/integration/native_auth/end_to_end/sign_in/MSALNativeAuthSignInUserNameAndPasswordEndToEndTests.swift
```diff
@@ -94,4 +83,32 @@ final class MSALNativeAuthSignInUsernameAndPasswordEndToEndTests: MSALNativeAuth
         XCTAssertNotNil(signInDelegateSpy.result?.idToken)
         XCTAssertEqual(signInDelegateSpy.result?.account.username, username)
     }
+    
+    func test_signInAndSendingIncorrectPasswordResultsInError() async throws {
+        guard let sut = initialisePublicClientApplication(clientIdType: .code), let username = getSignInUsernamePassword() else {
```
spetrescu84, line 88: Should this be `.password`?

nilo-ms: It can be anything to be precise. A user will be required to signIn based how they signUp, not based on the client Id in use. 
But let me remove the clientIdType for correctness

PR #2236, opened 2024-07-03
Run native auth E2E tests in the pipeline

nilo-ms:
## Proposed changes

- Create a dedicated test schema to native auth E2E tests
- Run native auth E2E tests as part of the PR validation

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [X] Test
- [ ] Logging/Telemetry

## Risk


spetrescu84 (approved):

azure_pipelines/pr-validation.yml
```diff
@@ -84,6 +84,21 @@ jobs:
       failTaskOnFailedTests: true
       testRunTitle: 'Test Run - $(target)'
 
+- job: e2e_test_native_auth
+  displayName: 'Run MSAL E2E tests for native auth'
```
spetrescu84, line 88: Should we rename this to 'Run iOS MSAL E2E tests for native auth'?

spetrescu84: We will update this when the macOS job and schema gets added

azure_pipelines/templates/tests-with-conf-file.yml
```diff
@@ -0,0 +1,73 @@
...
+steps:
+  - checkout: self
+    clean: true
+    submodules: true
+    fetchDepth: 1
+    persistCredentials: true
+    path: s
+    
+  - task: Bash@3
+    displayName: 'Select Xcode version'
```
diegojerezba, line 14: We use different Xcode versions in this file and in `pr-validation.yml`. 

What do you think about creating a new `parameter` here and in `pr-validation.yml`, so we set the same version in both files?

Ideally we would have only one single place where to put all these variables, but I don't know if that's possible.

azure_pipelines/templates/tests-with-conf-file.yml
```diff
@@ -0,0 +1,73 @@
...
+    displayName: Run automations
+    inputs:
+      targetType: 'inline'
+      script: |
+        ls build/Build/Products/
+        xcodebuild test-without-building \
+            -xctestrun 'build/Build/Products/${{ parameters.build }}_iphonesimulator17.0-x86_64.xctestrun' \
+            -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.2' \
+            -retry-tests-on-failure \
+            -parallel-testing-enabled NO \
```
diegojerezba, line 63: I thought we had enabled it to run on parallel simulators. Does it not work faster in that way?

azure_pipelines/templates/tests-with-conf-file.yml
nilo-ms: Yeah, I removed the macOS declaration for all jobs because the pipeline was failing (it was selecting ubuntu as image instead of mac). Here is the commit:
https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2236/commits/21cde46e9f1eda51eef3952e9b91b4f4b52bb333
I used the same approach as the automation.yml file, but it didn't work

nilo-ms: Are you suggesting creating a new parameter to set different xcode version from both script?

diegojerezba (approved):

azure_pipelines/templates/tests-with-conf-file.yml
diegojerezba: Resolving this comment as discussed internally.

nilo-ms: Sharing here too: variables can't be created in a steps templates, but only in variables template

kaisong1990 (approved):

PR #2253, opened 2024-07-19
Mask Native Auth logs subject to PII

diegojerezba:
## Proposed changes

In this PR the Native Auth logs have been reviewed and updated, by using `MSALLogger.logPII()` in conjunction with `MSALLogMask` and its different types of masking options.

In general, this is the logic I have followed:

- All of the API errors are masked.
- None of the request errors have been masked (because I don't think they are not leaking any PII).
- The cache-related errors have been masked.

Important note 1:

In the controllers, when we log the errorDescription, we had in all of them: "\(error.errorDescription ?? "No error description")". I have removed the "no error description" part, so now if there isn't an error description, it uses the error description that we provide in the public errors (for example, see from `SignUpStartError` the `errorDescription` property). 

Before we were not using those descriptions, IMO they are richer than simply returning "no error description". But let me know if you don't agree with this change, I can revert it.

Important note 2:

I have made the decision of **not** to mask the following logs, and I would like to highlight them so everybody is aware.

- `MSALNativeAuthUrlRequestSerializer` class, line 70: when there is an error serializing a request.
- `MSALNativeAuthRequestConfigurator` class, line 265: when there is a problem creating the endpoint url.
- `MSALNativeAuthRequestContext` class, line 78: when there is a mismatch of the correlation id set in the SDK and the correlation id received in the response. Since all of the logs show the correlation id, I think it's not needed to mask them.

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [x] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
+++ b/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
@@ -94,7 +94,7 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
             MSALLogger.logPII(
                 level: .error,
                 context: nil,
-                format: "Error retrieving accounts \(MSALLogMask.maskPII(error))")
+                format: "Error retrieving accounts \(MSALLogMask.maskEUII(error))")
         }
         return []
     }
--- a/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
+++ b/MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
@@ -91,10 +91,10 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController, 
             let config = factory.makeMSIDConfiguration(scopes: [])
             return try cacheAccessor.getAllAccounts(configuration: config)
         } catch {
-            MSALLogger.log(
+            MSALLogger.logPII(
                 level: .error,
                 context: nil,
-                format: "Error retrieving accounts \(error)")
+                format: "Error retrieving accounts \(MSALLogMask.maskPII(error))")
         }
         return []
     }
--- a/MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
+++ b/MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
@@ -156,16 +156,16 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
                 correlationId: context.correlationId()
             )
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Redirect error in resetpassword/start request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "Redirect error in resetpassword/start request \(MSALLogMask.maskPII(error.errorDescription))")
             return .init(.error(error), correlationId: context.correlationId())
         case .error(let validatedError):
             let error = validatedError.toResetPasswordStartPublicError(context: context)
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Error in resetpassword/start request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "Error in resetpassword/start request \(MSALLogMask.maskPII(error.errorDescription))")
             return .init(.error(error), correlationId: context.correlationId())
         case .unexpectedError(let apiError):
             let error = ResetPasswordStartError(
@@ -176,9 +176,9 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
                 errorUri: apiError?.errorURI
             )
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Unexpected error in resetpassword/start request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "Unexpected error in resetpassword/start request \(MSALLogMask.maskPII(error.errorDescription))")
             return .init(.error(error), correlationId: context.correlationId())
         }
     }
@@ -231,9 +231,9 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
         case .error(let apiError):
             let error = apiError.toResetPasswordStartPublicError(correlationId: context.correlationId())
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Error in resetpassword/challenge request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "Error in resetpassword/challenge request \(MSALLogMask.maskPII(error.errorDescription))")
             return .init(.error(error), correlationId: context.correlationId())
         case .redirect:
             let error = ResetPasswordStartError(
@@ -242,9 +242,9 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
                 correlationId: context.correlationId()
             )
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Redirect error in resetpassword/challenge request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "Redirect error in resetpassword/challenge request \(MSALLogMask.maskPII(error.errorDescription))")
             return .init(.error(error), correlationId: context.correlationId())
         case .unexpectedError(let apiError):
             let error = ResetPasswordStartError(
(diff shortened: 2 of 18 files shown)


MSAL/src/native_auth/controllers/reset_password/MSALNativeAuthResetPasswordController.swift
```diff
@@ -156,16 +156,16 @@ final class MSALNativeAuthResetPasswordController: MSALNativeAuthBaseController,
                 correlationId: context.correlationId()
             )
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "Redirect error in resetpassword/start request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "Redirect error in resetpassword/start request \(MSALLogMask.maskPII(error.errorDescription))")
```
nilo-ms, line 161: Can you please check with MSAL Apple team (or Sergei) if we still need to mask PII if we log using PII? I'd expect that if we notify that the log contains PII there is no need to mask the content. At least this is how native auth Android operates and it makes sense to me. I know I wrote differently in the PBI description, but this is something came later :/

diegojerezba: This sample code is to show what we talked today in the morning:

Screenshot 2024-07-23 at 17 33 41

If the developer has the **default configuration** for MSAL:

`MSALGlobalConfig.loggerConfig.logMaskingLevel = .settingsMaskAllPII`

These are the resulting logs:

Screenshot 2024-07-23 at 17 45 17

In the first log the username is leaked. Note how in both logs the `containsPII` bool is false. This is because the `logMaskingLevel` is set to `.settingsMaskAllPII`.

===

Now let's configure the SDK with the most permissive masking level:

`MSALGlobalConfig.loggerConfig.logMaskingLevel = .settingsMaskSecretsOnly`

These are the resulting logs:

Screenshot 2024-07-23 at 17 40 06

Our masked log now shows the full errorDescription and the `containsPII` bool is true.

nilo-ms: Thanks for the clarification Diego

MSAL/src/native_auth/network/MSALNativeAuthResponseSerializer.swift
```diff
@@ -40,7 +40,7 @@ final class MSALNativeAuthResponseSerializer<T: Decodable & MSALNativeAuthRespon
             response.correlationId = T.retrieveCorrelationIdFromHeaders(from: httpResponse)
```
nilo-ms, line 40: @diegojerezba could you also check the log  `MSALNativeAuthUrlRequestSerializer` at line 67? I think we should log that as PII because the request body can contains PII too.

MSAL/src/native_auth/network/responses/validator/sign_in/MSALNativeAuthSignInResponseValidator.swift
```diff
@@ -74,10 +74,10 @@ final class MSALNativeAuthSignInResponseValidator: MSALNativeAuthSignInResponseV
             return .error(.unexpectedError(.init(errorDescription: MSALNativeAuthErrorMessage.unexpectedResponseBody)))
         case .failure(let responseError):
             guard let initiateResponseError = responseError as? MSALNativeAuthSignInInitiateResponseError else {
-                MSALLogger.log(
+                MSALLogger.logPII(
                     level: .error,
                     context: context,
-                    format: "signin/initiate: Unable to decode error response: \(responseError)")
+                    format: "signin/initiate: Unable to decode error response: \(MSALLogMask.maskPII(responseError))")
                 return .error(.unexpectedError(.init(errorDescription: MSALNativeAuthErrorMessage.unexpectedResponseBody)))
```
nilo-ms, line 81: Is there a reason you didn't mark as PII also the log at line 104? I think we should mark that log as PII

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -204,9 +204,9 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
                 errorUri: apiError.errorURI
             )
             stopTelemetryEvent(event, context: context, error: error)
-            MSALLogger.log(level: .error,
-                           context: context,
-                           format: "InvalidUsername in signup/start request \(error.errorDescription ?? "No error description")")
+            MSALLogger.logPII(level: .error,
+                              context: context,
+                              format: "InvalidUsername in signup/start request \(MSALLogMask.maskPII(error.errorDescription))")
```
nilo-ms, line 209: Username is EUII, can you use the correct mark? Thanks

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
```diff
@@ -91,10 +91,10 @@ final class MSALNativeAuthCredentialsController: MSALNativeAuthTokenController,
             let config = factory.makeMSIDConfiguration(scopes: [])
             return try cacheAccessor.getAllAccounts(configuration: config)
         } catch {
-            MSALLogger.log(
+            MSALLogger.logPII(
                 level: .error,
                 context: nil,
-                format: "Error retrieving accounts \(error)")
+                format: "Error retrieving accounts \(MSALLogMask.maskPII(error))")
```
nilo-ms, line 97: this error can contains EUII, can you mark it as such?

nilo-ms (approved):

spetrescu84:
Do we need to mask the data we pass to `stopTelemetryEvent` that is sent to the telemetry endpoint?

spetrescu84:
Not sure about this, but requests that fail on creation, but are secondary such as `/challenge` leading to this log:
`MSALLogger.log(level: .error, context: context, format: "Error while creating Challenge Request: \(error)")`
Could they contain PII in that error because of the parameters used? Like the `continuation_token`?

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
```diff
@@ -65,10 +65,10 @@ import Foundation
                 context: context
             )
         } catch {
-            MSALLogger.log(
+            MSALLogger.logPII(
                 level: .error,
                 context: context,
-                format: "Clearing MSAL token cache for the current account failed with error %@: \(error)"
+                format: "Clearing MSAL token cache for the current account failed with error %@: \(MSALLogMaskWrapper.maskPII(error))"
```
spetrescu84, line 71: Should this be MSALLogMaskWrapper.maskEUII(error)?

spetrescu84:
Should the error in the file `MSALNativeAuthUrlRequestSerializer`: 
`"HTTP body request serialization failed with error: \(error.localizedDescription)"`
Be marked as PII in case the decoded field contains it and fails on decode?

diegojerezba:
> Not sure about this, but requests that fail on creation, but are secondary such as `/challenge` leading to this log: `MSALLogger.log(level: .error, context: context, format: "Error while creating Challenge Request: \(error)")` Could they contain PII in that error because of the parameters used? Like the `continuation_token`?

The `RequestConfigurator` only throws custom errors (`MSALNativeAuthInternalError.invalidUrl` and `MSALNativeAuthInternalError.invalidRequest`).

MSAL/src/native_auth/network/MSALNativeAuthResponseSerializer.swift
diegojerezba: I decided not to add it because I think the exception thrown only contains the key. But I'm adding it just in case.

MSAL/src/native_auth/network/responses/validator/sign_in/MSALNativeAuthSignInResponseValidator.swift
diegojerezba: Done

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
```diff
@@ -169,10 +169,10 @@ final class MSALNativeAuthSignUpController: MSALNativeAuthBaseController, MSALNa
             let challengeResult = await performAndValidateChallengeRequest(continuationToken: continuationToken, context: context)
             return handleSignUpChallengeResult(challengeResult, username: username, event: event, context: context)
         case .attributeValidationFailed(let apiError, let invalidAttributes):
-            MSALLogger.log(
+            MSALLogger.logPII(
                 level: .error,
                 context: context,
-                format: "attribute_validation_failed received from signup/start request for attributes: \(invalidAttributes)"
+                format: "attribute_validation_failed received from signup/start request for attributes: \(MSALLogMask.maskEUII(invalidAttributes))"
```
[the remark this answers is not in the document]
diegojerezba: Yes

MSAL/src/native_auth/controllers/sign_up/MSALNativeAuthSignUpController.swift
diegojerezba: Done, thank you.

MSAL/src/native_auth/controllers/credentials/MSALNativeAuthCredentialsController.swift
diegojerezba: Done, thank you

MSAL/src/native_auth/public/MSALNativeAuthUserAccountResult.swift
diegojerezba: Done

diegojerezba:
> Do we need to mask the data we pass to `stopTelemetryEvent` that is sent to the telemetry endpoint?

No, the fields that are sent after a response (error code, oauth2 error code and error domain) don't contain any PII. You can see what we send in `MSALNativeAuthBaseController`, starting at line 98.

The developer can configure these telemetry events with the following: `MSALGlobalConfig.telemetryConfig.piiEnabled = true`. This is what is shared in the callback depending on the developer's config:

Screenshot 2024-07-25 at 16 44 37

spetrescu84 (approved):

borgesmb (approved):

PR #2302, opened 2024-08-26
Email OTP MFA business logic

nilo-ms:
## Proposed changes

This PR contains the real implementation of email OTP MFA. 
I've developed dedicated results and protocols for MFA and integrated them into the SignInController class, as it already contained the majority of the necessary code.

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


MSAL/src/native_auth/network/responses/validator/sign_up/MSALNativeAuthSignUpResponseValidator.swift
```diff
@@ -213,7 +213,7 @@ final class MSALNativeAuthSignUpResponseValidator: MSALNativeAuthSignUpResponseV
                 MSALLogger.log(level: .error, context: context, format: "Missing expected fields in signup/continue for attributes_required error")
                 return .unexpectedError(.init(errorDescription: MSALNativeAuthErrorMessage.unexpectedResponseBody))
             }
-        // TODO: .verificationRequired is not supported by the API team yet. We treat it as an unexpectedError
+        // TODO: .verificationRequired is returned by server when user submits attribute but email isn't verified yet. It needs to be handled by SDK
```
nilo-ms, line 216: the previous comment wasn't accurate, so I updated it

MSAL/src/native_auth/public/state_machine/error/MFAError.swift
```diff
@@ -0,0 +1,60 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.  
+
+import Foundation
+
+/// Class that defines the structure and type of a MFAError
+@objcMembers
+public class MFAError: MSALNativeAuthError {
```
nilo-ms, line 29: This error is used by `sendChallenge` and `getAuthMethods` operations

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -358,13 +319,222 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
+        )
+        switch response.result {
+        case .completed(let accountResult):
+            return .init(
+                .completed(accountResult),
+                correlationId: response.correlationId,
+                telemetryUpdate: { [weak self] result in
+                    self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result)
+                })
+        case .error(let error, let newState):
```
SammyO, line 440: this calls the /token endpoint, right? Where is the handling for the API returning mfa_required again? I know it can't happen practically right now, but it can happen in the future. Will that response result in an .error case?

MSAL/test/integration/native_auth/requests/sign_in/MSALNativeAuthSignInIntrospectIntegrationTests.swift
```diff
@@ -63,7 +63,7 @@ class MSALNativeAuthSignInIntrospectIntegrationTests: MSALNativeAuthIntegrationB
             return XCTFail("No authentication method returned")
         }
         XCTAssertEqual(firstMethod.id, "F37D8C55-BE83-449F-8F99-131F6553871D")
-        XCTAssertEqual(firstMethod.challengeChannel, .email)
+        XCTAssertEqual(firstMethod.challengeChannel, "email")
```
SammyO, line 66: Are these E2E tests executed on the public SDK interface (rather than tests that test some internal component)? If so, I'd recommend adding some (mock-API-based) tests for MFA here as well, so you can test the entire chain.

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -140,6 +141,12 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
                         self?.stopTelemetryEvent(telemetryInfo.event, context: context, delegateDispatcherResult: result)
                     }))
                 },
+                onAwaitingMFA: { _ in
```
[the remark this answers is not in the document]
nilo-ms: 1. We don't need [weak self] in this case (and TBH I don't think we can use it). We use weak self in the telemetry callback from the delegate dispatcher.
2. Yes, this code is also used in the signIn after SSPR flow

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -358,13 +319,222 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
+            }
+        case .error(let errorType):
+            return processSubmitCodeFailure(
+                errorType: errorType,
+                telemetryInfo: telemetryInfo,
+                scopes: scopes,
+                continuationToken: continuationToken,
+                context: context
+            )
+        case .strongAuthRequired(_):
```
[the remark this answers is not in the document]
nilo-ms: I didn't remove it because of swiftlint warning, but now I disable it. Thanks

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -358,13 +319,222 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
+            let error = VerifyCodeError(type: .generalError, correlationId: context.correlationId())
+            MSALLogger.log(level: .error, context: context, format: "Submit code: received unexpected MFA required API result")
```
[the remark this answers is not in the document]
nilo-ms: I don't think is useful here because we need to inform the user that an unexpected server response has been received. The error description that we're printing at line 851 is an internal error description, not the one from the server

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -304,19 +249,35 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
                 continuationToken: continuationToken,
                 scopes: scopes
             )
-        case .strongAuthRequired(_):
-            // TODO: this will be handled in a separate PBI
-            return .init(.error(error: PasswordRequiredError(type: .generalError, correlationId: UUID()), newState: nil), correlationId: UUID())
+        case .strongAuthRequired(let newContinuationToken):
```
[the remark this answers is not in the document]
nilo-ms: I prefer to call it "newContinuationToken" so it can't be confused with the method parameter called "continuationToken" (line 183)

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -358,13 +319,222 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
+        continuationToken: String,
+        authMethod: MSALAuthMethod?,
+        context: MSALNativeAuthRequestContext,
+        scopes: [String]
+    ) async -> MFARequestChallengeControllerResponse {
+        let event = makeAndStartTelemetryEvent(id: .telemetryApiIdMFARequestChallenge, context: context)
+        let result = await performAndValidateChallengeRequest(
+            continuationToken: continuationToken,
+            context: context,
+            errorDescription: "MFA RequestChallenge: cannot create challenge request object",
```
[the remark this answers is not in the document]
nilo-ms: Now that you pointed out, this message should only be included in the log and not the in the error description. Updated

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -605,29 +829,47 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
 
+    private func performAndValidateIntrospectRequest(
+        continuationToken: String,
+        context: MSALNativeAuthRequestContext
+    ) async -> MSALNativeAuthSignInIntrospectValidatedResponse {
+        guard let introspectRequest = createIntrospectRequest(
+            continuationToken: continuationToken,
+            context: context
+        ) else {
+            let errorDescription = "Unable to create signIn/initiate request"
```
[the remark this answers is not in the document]
nilo-ms: Good catch

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -605,29 +829,47 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
         case .introspectRequired:
-            // TODO: this will be handled in a separate PBI
-            return .init(.error(
-                SignInStartError(
-                    type: .generalError,
-                    correlationId: telemetryInfo.context.correlationId()
-                )), correlationId: telemetryInfo.context.correlationId()
-            )
+            let error = SignInStartError(type: .generalError, correlationId: telemetryInfo.context.correlationId())
+            MSALLogger.log(level: .error, context: telemetryInfo.context, format: "SignIn, received unexpected introspect required API result")
```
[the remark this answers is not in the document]
nilo-ms: this result is expected only after the user triggered the "requestChallenge". It is correctly handled at line 384

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
nilo-ms: In the `MSALNativeAuthTokenResponseValidator` class the "mfa_required" is converted to a dedicated internal validated result that is "strongAuthRequired". This type of result will be converted to an error for the following scenario:
- signIn after signUp/SSPR
- signIn with username and code
Only during the signIn with password this result is converted to the state "awaitingMFA" .

SammyO: That's not exactly my point. What I mean is: when you submit a challenge during the sign in with password flow, i.e. in the MFA flow, and the API *again* returns mfa_required (e.g. if another MFA cycle is needed), what happens?

MSAL/test/integration/native_auth/requests/sign_in/MSALNativeAuthSignInIntrospectIntegrationTests.swift
nilo-ms: Those are integration tests that uses mock APIs. This particular class check that the Introspect request is created correctly, and the response is handled as expected. We plan to add E2E tests to test the entire chain (using only public interface)

SammyO: You plan to add them when? Shouldn't they be part of this PR, given that this is the business logic PR?

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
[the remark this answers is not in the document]
diegojerezba: For 2, what I mean is that we are returning a `SignInAfterSignUpError` in case the user comes from the SignInAfter SSPR flow (instead of `SignInAfterResetPasswordError`).

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
nilo-ms: Ah, now I got you. We will return an error because we don't expect a strong authentication after submitting a code (second factor after signIn with username and code and after submitting a challenge). Check line 531

MSAL/test/integration/native_auth/requests/sign_in/MSALNativeAuthSignInIntrospectIntegrationTests.swift
SammyO: I don't think we're talking about the same thing. I'm asking about E2E tests (executed on the interface, flowing through the entire SDK), that use the mock API (rather than real production environment). I believe they aren't part of this PR. Why is that?

You don't need the lab tenant for these tests. Nor are they the same as unit tests.

MSAL/src/native_auth/public/state_machine/state/MFAStates.swift
```diff
@@ -24,51 +24,97 @@
...
+        controller: MSALNativeAuthMFAControlling,
+        scopes: [String],
+        continuationToken: String,
+        correlationId: UUID) {
+        self.controller = controller
+        self.scopes = scopes
+        super.init(continuationToken: continuationToken, correlationId: correlationId)
+    }
+
+    func baseRequestChallenge(authMethod: MSALAuthMethod?, delegate: MFARequestChallengeDelegate) {
```
spetrescu84, line 42: This is not public right?

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInControlling.swift
```diff
@@ -58,4 +58,5 @@ protocol MSALNativeAuthSignInControlling {
     ) async -> SignInSubmitPasswordControllerResponse
 
     func resendCode(continuationToken: String, context: MSALNativeAuthRequestContext, scopes: [String]) async -> SignInResendCodeControllerResponse
+
```
spetrescu84, line 61: nit: extra spacing

MSAL/src/native_auth/network/responses/sign_in/MSALNativeAuthInternalAuthenticationMethod.swift
```diff
@@ -30,4 +30,13 @@ struct MSALNativeAuthInternalAuthenticationMethod: Decodable, Equatable {
     let challengeType: MSALNativeAuthInternalChallengeType
     let challengeChannel: String
     let loginHint: String
+
+    func toPublicAuthMethod() -> MSALAuthMethod {
+        return MSALAuthMethod(
```
spetrescu84, line 35: Why is this class not called MSALNativeAuthMethod? It can be use for non native also?

MSAL/src/native_auth/public/MSALAuthMethod.swift
```diff
@@ -20,7 +20,7 @@
 // AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
+// THE SOFTWARE.
```
spetrescu84, line 23: nit: extra spacing

MSAL/test/integration/native_auth/requests/sign_in/MSALNativeAuthSignInIntrospectIntegrationTests.swift
nilo-ms: We don't have such tests. Integration tests check that some components integrate as expected. We use this type of tests for tests like Validator-Network-Server (mock API) or Delegate-Delegate Dispatcher - Controller.
But we test the full flow only using the real server (E2E tests). We don't test that check the full flow that use the mock API

MSAL/src/native_auth/public/state_machine/state/MFAStates.swift
nilo-ms: Correct, can be called only from other states

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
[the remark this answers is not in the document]
nilo-ms: This is expected. The `SignInAfterSignUpError` is translated to `SignInAfterResetPasswordError` in the SSPR state class. But, I noticed that we use the wrong the telemetry id for the SSPR flow, so I'm going to fix it in this PR

MSAL/src/native_auth/network/responses/sign_in/MSALNativeAuthInternalAuthenticationMethod.swift
nilo-ms: There is no plan that this class can be used by the web flow. But the name with NativeAuth would be sound not appropriate because the "auth" is not part of the header, but part of the name of the class (authentication method). So, I decided to don't use "NativeAuth" in this case, to make the name of the class clearer.

MSAL/test/unit/native_auth/public/state_machine/mfa/AwaitingMFAStateTests.swift
```diff
@@ -0,0 +1,113 @@
...
+            newState: expectedState
+        )
+        controller.requestChallengeResponse = .init(expectedResult, correlationId: correlationId)
+
+        let delegate = MFARequestChallengeDelegateSpy(expectation: exp, expectedError: expectedError)
+
+        sut.requestChallenge(delegate: delegate)
+        wait(for: [exp], timeout: 1)
+
+        XCTAssertNil(delegate.newMFARequiredState)
```
diegojerezba, line 62: Maybe you can add a bool variable to the `MFARequestChallengeDelegateSpy` that is set to `true` if the `onMFARequestChallengeError` is called, and check for it here. Just to make the test easier for the reader.

I mean something like `onMFARequestChallengeErrorWasCalled: Bool`

MSAL/test/unit/native_auth/controllers/MSALNativeAuthMFAControllerTests.swift
```diff
@@ -0,0 +1,408 @@
...
+// THE SOFTWARE.  
+
+import XCTest
+@testable import MSAL
+@_implementationOnly import MSAL_Private
+@_implementationOnly import MSAL_Unit_Test_Private
+
+class MSALNativeAuthMFAControllerTests: MSALNativeAuthSignInControllerTests {
+    
+    func test_signInWithCodeSubmitCodeReceiveStrongAuthRequired_anErrorShouldBeReturned() {
```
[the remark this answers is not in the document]
nilo-ms: I initially added the tests that involves signIn in the signIn controller test class, but then I moved them here because kind of related to MFA. No strong opinion, but I think it's more accurate to leave them here.

MSAL/test/unit/native_auth/controllers/MSALNativeAuthMFAControllerTests.swift
```diff
@@ -0,0 +1,408 @@
...
+        checkTelemetryEventResult(id: .telemetryApiIdMFASubmitChallenge, isSuccessful: false)
+        if case .error(let error, let newState) = result.result {
+            XCTAssertEqual(error.type, .generalError)
+            XCTAssertNotNil(newState)
+        } else {
+            XCTFail("Expected error result")
+        }
+    }
+    
+    func test_whenSubmitChallengeTokenRequestReturnError_correctErrorShouldBeReturned() async {
```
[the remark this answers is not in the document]
nilo-ms: I think this test is ok to exists, because it tests a behaviour of our code base. And we're sure that it works as expected.
What I see not optimal in our unit test suite is that a single function is tested in multiple tests. This means that when that function change, many tests (not directly related to that function) needs to be changed.

MSAL/test/unit/native_auth/public/state_machine/mfa/AwaitingMFAStateTests.swift
nilo-ms: In the MFAGetAuthMethodsDelegateSpy we specify how the expected error should be. Once this attribute is set, the `onMFAGetAuthMethodsError` method check that the returned error is the same as the expected one. If this error is not set and the error callback is called the test fail.
Please, see MFAGetAuthMethodsDelegateSpy line 103

MSAL/test/unit/native_auth/controllers/MSALNativeAuthMFAControllerTests.swift
```diff
@@ -0,0 +1,408 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.  
+
+import XCTest
+@testable import MSAL
+@_implementationOnly import MSAL_Private
+@_implementationOnly import MSAL_Unit_Test_Private
```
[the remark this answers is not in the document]
nilo-ms: Correct. I copied the import from the signInController.

MSAL/test/unit/native_auth/public/state_machine/mfa/MFARequiredStateTests.swift
```diff
@@ -0,0 +1,187 @@
...
+    func test_submitInvalidChallenge_shouldReturnCorrectResponse() {
+        let exp = expectation(description: "mfa states")
+        let expectedError = MFASubmitChallengeError(type: .invalidChallenge, correlationId: correlationId)
+
+        let delegate = MFASubmitChallengeDelegateSpy(expectation: exp, expectedResult: nil, expectedError: expectedError)
+        
+        sut.submitChallenge(challenge: "", delegate: delegate)
+        wait(for: [exp], timeout: 1.0)
+    }
+
```
spetrescu84, line 186: nit: spacing

spetrescu84 (approved):

MSAL/test/unit/native_auth/public/state_machine/mfa/MFARequiredStateTests.swift
nilo-ms: Done

MSAL/test/unit/native_auth/controllers/MSALNativeAuthMFAControllerTests.swift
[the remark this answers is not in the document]
diegojerezba: I like that you added the tests that check the implementation of `MSALNativeAuthMFAControlling` in a separate file, I think that was a good idea to improve clarity. But the test I mention tests the State object.

Anyway, I don't have a strong opinion either.

diegojerezba (approved):

PR #2576, opened 2025-04-09
[iOS SDK] JIT Registration, Network implementation

spetrescu84:
## Proposed changes

Add the network layer along with the validation for it for JIT

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITContinueValidatedResponse.swift
```diff
@@ -0,0 +1,70 @@
...
+
+enum MSALNativeAuthJITContinueValidatedErrorType: Error {
+    case redirect
+    case unauthorizedClient(MSALNativeAuthJITContinueResponseError)
+    case invalidRequest(MSALNativeAuthJITContinueResponseError)
+    case invalidOOBCode(MSALNativeAuthJITContinueResponseError)
+    case unexpectedError(MSALNativeAuthJITContinueResponseError?)
+    case unsupportedChallengeType(MSALNativeAuthJITContinueResponseError)
+
+    func convertToRegisterStrongAuthSubmitChallengeError(correlationId: UUID) -> RegisterStrongAuthSubmitChallengeError {
```
spetrescu84, line 40: I am not sure this should be in as they are only used by the controllers, but the cases above are needed by the validator

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITIntrospectValidatedResponse.swift
```diff
@@ -0,0 +1,84 @@
...
+    case error(MSALNativeAuthJITIntrospectValidatedErrorType)
+}
+
+enum MSALNativeAuthJITIntrospectValidatedErrorType: Error {
+    case redirect
+    case expiredToken(MSALNativeAuthJITIntrospectResponseError)
+    case invalidRequest(MSALNativeAuthJITIntrospectResponseError)
+    case unexpectedError(MSALNativeAuthJITIntrospectResponseError?)
+
+    func convertToSignInStartError(correlationId: UUID) -> SignInStartError {
```
spetrescu84, line 38: I am not sure this should be in as they are only used by the controllers, but the cases above are needed by the validator

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
```diff
@@ -46,6 +46,7 @@ enum MSALNativeAuthErrorMessage {
     static let unexpectedResponseBody = "Unexpected response body received"
     static let unexpectedChallengeType = "Unexpected challenge type"
     static let refreshTokenMFARequiredError = "Multi-factor authentication is required, which can't be fulfilled as part of this flow. Please sign out and perform a new sign in operation. More information: "
+    static let strongAuthenticationRequiredError = "Strong authentication method is required, which can't be fulfilled as part of this flow. Please sign out and perform a new sign in operation. More information: "
```
nilo-ms, line 49: Where is this error message used?

MSAL/src/native_auth/network/errors/jit/MSALNativeAuthJITContinueOauth2ErrorCode.swift
```diff
@@ -0,0 +1,34 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+enum MSALNativeAuthJITContinueOauth2ErrorCode: String, Decodable, MSALNativeAuthUnknownCaseProtocol {
+    case invalidRequest = "invalid_request"
+    case expiredToken = "expired_token"
+    case invalidGrant = "invalid_grant"
+    case authorizationPending = "authorization_pending"
```
nilo-ms, line 31: Do we need to parse all these error codes? We should parse only the error codes that we need to handle in a different way than a generic error (invalid challenge or something similar). Everything else can be just use the default parser, which use SDK error type "generic_error" and parse the error codes and error description.

MSAL/src/native_auth/network/errors/jit/MSALNativeAuthJITChallengeResponseError.swift
```diff
@@ -0,0 +1,60 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+struct MSALNativeAuthJITChallengeResponseError: MSALNativeAuthResponseError {
+
+    let error: MSALNativeAuthJITChallengeOauth2ErrorCode
```
nilo-ms, line 29: Not an action item here, but I want to share some thoughts. It seems we have a lot of duplicated code in the response error code classes. Majority of them use the same structure. I think we can do something to group all of these classes in one single base class

MSAL/src/native_auth/network/responses/jit/MSALNativeAuthJITChallengeResponse.swift
```diff
@@ -0,0 +1,38 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+struct MSALNativeAuthJITChallengeResponse: Decodable, MSALNativeAuthResponseCorrelatable {
+
+    // MARK: - Variables
+    let continuationToken: String?
+    let challengeType: MSALNativeAuthInternalChallengeType
```
nilo-ms, line 31: This parsing will fail when "preverified" challenge type will be returned. In general, I don't think we should use an enum for "challenge_type" parsing. If a new challenge type is sent from the server, the parsing of the response will fail. I know that this is the approach we used in the past, but I think it needs to be improved. I suggest parsing this field as a string instead.

MSAL/src/native_auth/network/responses/jit/MSALNativeAuthJITContinueResponse.swift
```diff
@@ -0,0 +1,32 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+struct MSALNativeAuthJITContinueResponse: Decodable, MSALNativeAuthResponseCorrelatable {
+
+    // MARK: - Variables
+    let continuationToken: String?
+    var correlationId: UUID?
```
nilo-ms, line 31: It doesn't seem that the correlation ID is returned in the successful response

MSAL/src/native_auth/network/responses/jit/MSALNativeAuthJITIntrospectResponse.swift
```diff
@@ -0,0 +1,33 @@
...
+// THE SOFTWARE.  
+
+import Foundation
+
+struct MSALNativeAuthJITIntrospectResponse: Decodable, MSALNativeAuthResponseCorrelatable {
+
+    // MARK: - Variables
+    let continuationToken: String?
+    let methods: [MSALNativeAuthInternalAuthenticationMethod]?
+    var correlationId: UUID?
```
nilo-ms, line 32: Can you remind me if there was a specific reason we are adding correlation ID field here?

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
```diff
@@ -0,0 +1,187 @@
...
+            }
+    }
+
+    private func handleFailedJITContinueResult(error: MSALNativeAuthJITContinueResponseError) -> MSALNativeAuthJITContinueValidatedResponse {
+            switch error.error {
+            case .invalidRequest:
+                return .error(.invalidRequest(error))
+            case .unknown:
+                return .error(.unexpectedError(error))
+            case .invalidGrant:
```
nilo-ms, line 177: Are we sure that invalid grant always means invalid OOB? Shouldn't we check suberror or error code?

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITChallengeValidatedResponse.swift
```diff
@@ -0,0 +1,65 @@
...
+
+enum MSALNativeAuthJITChallengeValidatedErrorType: Error {
+    case redirect
+    case expiredToken(MSALNativeAuthJITChallengeResponseError)
+    case invalidToken(MSALNativeAuthJITChallengeResponseError)
+    case unauthorizedClient(MSALNativeAuthJITChallengeResponseError)
+    case invalidRequest(MSALNativeAuthJITChallengeResponseError)
+    case unexpectedError(MSALNativeAuthJITChallengeResponseError?)
+
+    func convertToRegisterStrongAuthChallengeError(correlationId: UUID) -> RegisterStrongAuthChallengeError {
```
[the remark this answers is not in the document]
nilo-ms: It's difficult to review those methods without knowing when they will be called in the controller class

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITContinueValidatedResponse.swift
```diff
@@ -0,0 +1,70 @@
...
+        switch self {
+        case .redirect:
+            return .init(type: .generalError, correlationId: correlationId)
```
nilo-ms, line 43: Why this redirect is not converted to browser required error?

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITContinueValidatedResponse.swift
```diff
@@ -0,0 +1,70 @@
...
+        case .unexpectedError(let apiError):
+            return .init(
+                type: .generalError,
+                message: apiError?.errorDescription,
+                correlationId: correlationId,
+                errorCodes: apiError?.errorCodes ?? [],
+                errorUri: apiError?.errorURI
+            )
+        case .unauthorizedClient(let apiError),
```
nilo-ms, line 52: Can you group this group with "unexpectedError" ?

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITChallengeValidatedResponse.swift
```diff
@@ -0,0 +1,65 @@
...
+        switch self {
+        case .redirect:
+            return .init(type: .generalError, correlationId: correlationId)
```
nilo-ms, line 43: Why are we converting redirect to general error?

MSAL/src/native_auth/network/responses/validator/sign_up/MSALNativeAuthSignUpResponseValidator.swift
```diff
@@ -257,7 +257,8 @@ final class MSALNativeAuthSignUpResponseValidator: MSALNativeAuthSignUpResponseV
             }
         case .unknown,
             .introspectRequired,
-            .mfaRequired:
+            .mfaRequired,
```
nilo-ms, line 260: Why are those errors handled in signUp validator while they should be returned only in /token endpoint?

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
spetrescu84: It should not be part of this PR. Will be used in the UserAccountResult.

MSAL/src/native_auth/network/errors/jit/MSALNativeAuthJITContinueOauth2ErrorCode.swift
spetrescu84: I think in other cases we parse all but return to general error. I will check.

MSAL/src/native_auth/network/responses/jit/MSALNativeAuthJITContinueResponse.swift
spetrescu84: By looking at other responses none of them have it in the body, so my understanding is that it's taken from the header, but will confirm

MSAL/src/native_auth/network/responses/jit/MSALNativeAuthJITIntrospectResponse.swift
spetrescu84: By looking at other responses none of them have it in the body, so my understanding is that it's taken from the header, but will confirm

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
spetrescu84: That's my understanding from the document, we can confirm.

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITContinueValidatedResponse.swift
spetrescu84: We need to check on this one as I don't see the error in the document, it might be overlooked.

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITChallengeValidatedResponse.swift
spetrescu84: We can confirm how we handle this one

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
```diff
@@ -0,0 +1,187 @@
...
+    func validateContinue(
+        context: MSIDRequestContext,
+        result: Result<MSALNativeAuthJITContinueResponse, Error>
+    ) -> MSALNativeAuthJITContinueValidatedResponse {
+        switch result {
+        case .success(let initiateResponse):
+            if let continuationToken = initiateResponse.continuationToken {
+                return .success(continuationToken: continuationToken)
+            }
+            MSALLogger.log(level: .error, context: context, format: "register/initiate: challengeType and continuation token empty")
```
Yuki-YuXin, line 112: nit: "register/continue"

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
```diff
@@ -0,0 +1,187 @@
...
+            return .error(.unexpectedError(.init(errorDescription: MSALNativeAuthErrorMessage.unexpectedResponseBody)))
+        case .failure(let responseError):
+            guard let initiateResponseError = responseError as? MSALNativeAuthJITContinueResponseError else {
+                MSALLogger.logPII(
+                    level: .error,
+                    context: context,
+                    format: "register/initiate: Unable to decode error response: \(MSALLogMask.maskPII(responseError))")
```
Yuki-YuXin, line 119: nit: "register/continue"

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITContinueValidatedResponse.swift
Yuki-YuXin: Personal thoughts: redirect will be triggered at the starting point of the sign in initiate

Yuki-YuXin (approved):

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
spetrescu84: Updated to use the suberror also

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITContinueValidatedResponse.swift
spetrescu84: Treating this one as redirect but in the next PR as I've removed the whole function.
It can be the case where the admin changes the settings on the tenant between API calls and this will return redirect at this point but not on initiate.

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITIntrospectValidatedResponse.swift
spetrescu84: Moved to other PR

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITChallengeValidatedResponse.swift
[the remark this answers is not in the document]
spetrescu84: Moved to other PR

MSAL/src/native_auth/network/errors/jit/MSALNativeAuthJITChallengeResponseError.swift
spetrescu84: After the other comments the Challenge one also got the subError so that makes them more different now.

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITChallengeValidatedResponse.swift
spetrescu84: Will change in other PR to be `browserRequired`

nilo-ms:
@spetrescu84 I was not able to find the changes to the "SignInValidator" class, where we parse the new response "registration required". Did you add this change in this PR?

spetrescu84:
> @spetrescu84 I was not able to find the changes to the "SignInValidator" class, where we parse the new response "registration required". Did you add this change in this PR?

@nilo-ms I pushed now the code, there is no way to decouple the response of the token from the one of the controller so I needed to change the controller also. It's actually in multiple places - SignInValidator and TokenValidator

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
```diff
@@ -0,0 +1,173 @@
...
+                level: .error,
+                context: context,
+                format: "register/challenge: Received unexpected challenge type: \(response.challengeType)")
+            return .error(.unexpectedError(.init(errorDescription: MSALNativeAuthErrorMessage.unexpectedChallengeType)))
+        }
+    }
+
+    private func handleFailedJITChallengeResult(
+        error: MSALNativeAuthJITChallengeResponseError) -> MSALNativeAuthJITChallengeValidatedResponse {
+            switch error.error {
```
nilo-ms, line 157: I think we need to parse the result when we receive error = "invalid_request" and error code is 901001.  We receive this error when the external dev submit an invalid "verification contact" (challenge target)

nilo-ms: We also need a dedicated validated response result for this case

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignInDelegateDispatchers.swift
```diff
@@ -77,6 +77,21 @@ final class SignInStartDelegateDispatcher: DelegateDispatcher<SignInStartDelegat
...
     }
 
+    func dispatchJITRequired(authMethods: [MSALAuthMethod], newState: RegisterStrongAuthState, correlationId: UUID) async {
+        if let onSignInJITRequired = delegate.onSignInStrongAuthMethodRegistration {
+            telemetryUpdate?(.success(()))
+            await onSignInJITRequired(authMethods, newState)
+        } else {
+            let error = SignInStartError(
+                type: .generalError,
+                message: requiredErrorMessage(for: "onSignInJITRequired"),
```
nilo-ms, line 87: I don't think this is the right name of the method "onSignInJITRequired". It should be:
```suggestion
                message: requiredErrorMessage(for: "onSignInStrongAuthMethodRegistration"),
```

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
spetrescu84: Added the error type invalid_request and doing the validation in the validator along with the `invalidVerificationContact` which in the controller will be dispatched to `onRegisterStrongAuthChallengeError(error: RegisterStrongAuthChallengeError, newState: RegisterStrongAuthState?)` 
I am thinking here, the continuation_token, is the old one? Or a new one comes back with the error?

MSAL/src/native_auth/network/responses/validator/jit/validated_response/MSALNativeAuthJITChallengeValidatedResponse.swift
```diff
@@ -0,0 +1,35 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+enum MSALNativeAuthJITChallengeValidatedResponse {
+    case codeRequired(continuationToken: String, sentTo: String, channelType: MSALNativeAuthChannelType, codeLength: Int)
```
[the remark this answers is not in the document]
spetrescu84: Added

MSAL/src/native_auth/public/state_machine/delegate_dispatcher/SignInDelegateDispatchers.swift
```diff
@@ -120,6 +135,21 @@ final class SignInPasswordRequiredDelegateDispatcher: DelegateDispatcher<SignInP
...
     }
+
+    func dispatchJITRequired(authMethods: [MSALAuthMethod], newState: RegisterStrongAuthState, correlationId: UUID) async {
+        if let onSignInJITRequired = delegate.onSignInStrongAuthMethodRegistration {
+            telemetryUpdate?(.success(()))
+            await onSignInJITRequired(authMethods, newState)
+        } else {
+            let error = PasswordRequiredError(
+                type: .generalError,
+                message: requiredErrorMessage(for: "onSignInPasswordRequiredJITRequired"),
```
nilo-ms, line 146: same here:
```suggestion
                message: requiredErrorMessage(for: "onSignInStrongAuthMethodRegistration"),
```

MSAL/src/native_auth/network/responses/validator/jit/MSALNativeAuthJITResponseValidator.swift
nilo-ms: The old continuation token is used

nilo-ms (approved):

MSAL/src/native_auth/network/errors/jit/MSALNativeAuthJITChallengeResponseError.swift
nilo-ms, line 29: Just for my understanding, what happen when we receive an error, for example, with "invalid_grant" as error field? Will the parsing of this object fail completely (meaning no fields are parsed), or will the error field be marked as "unknown"?
The second is ok, but the first one no

MSAL/test/integration/native_auth/common/Model.swift
```diff
@@ -90,6 +93,14 @@ enum MockAPIResponse: String {
     case ssprPollNotStarted = "SSPRPollNotStarted"
     case signUpContinueSuccess = "SignUpContinueSuccess"
     case invalidUsername = "InvalidUsername"
+    case registrationRequired = "RegistrationRequired"
+    case registrationIntrospectSuccess = "RegistraionIntrospectSuccess"
```
nilo-ms, line 97: I'd add a comment here saying that the typo is expected

MSAL/src/native_auth/network/errors/jit/MSALNativeAuthJITChallengeResponseError.swift
spetrescu84: We have this in the `MSALNativeAuthUnknownCaseProtocol`, if it fails it creates an `unknown` error

nilo-ms: Ok, thanks for clarifying

MSAL/test/integration/native_auth/common/Model.swift
spetrescu84: Done

PR #2635, opened 2025-05-29
Native auth: backward compatibility, add new capabilities parameter 

nilo-ms:
## Proposed changes

This PR contains:
- New public parameter "capabilities"
- Send this parameter in /initiate and /start endpoints

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [X] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/src/native_auth/client_application_wrapper/silent_token/MSALNativeAuthSilentTokenProvider.swift
+++ b/MSAL/src/native_auth/client_application_wrapper/silent_token/MSALNativeAuthSilentTokenProvider.swift
@@ -27,10 +27,8 @@ class MSALNativeAuthSilentTokenProvider: MSALNativeAuthSilentTokenProviding {
 
     private let application: MSALNativeAuthPublicClientApplication?
 
-    init(
-        configuration config: MSALPublicClientApplicationConfig,
-        challengeTypes: MSALNativeAuthChallengeTypes) throws {
-            self.application = try? MSALNativeAuthPublicClientApplication(configuration: config, challengeTypes: challengeTypes)
+    init(configuration: MSALNativeAuthPublicClientApplicationConfig) throws {
+        self.application = try? MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: configuration)
         }
 
     func acquireTokenSilent(parameters: MSALSilentTokenParameters,
--- a/MSAL/src/native_auth/client_application_wrapper/silent_token/MSALNativeAuthSilentTokenProvider.swift
+++ b/MSAL/src/native_auth/client_application_wrapper/silent_token/MSALNativeAuthSilentTokenProvider.swift
@@ -29,7 +29,7 @@ class MSALNativeAuthSilentTokenProvider: MSALNativeAuthSilentTokenProviding {
 
     init(configuration: MSALNativeAuthPublicClientApplicationConfig) throws {
         self.application = try? MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: configuration)
-        }
+    }
 
     func acquireTokenSilent(parameters: MSALSilentTokenParameters,
                             completionBlock: @escaping MSALNativeAuthSilentTokenResponse) {
--- a/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
+++ b/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
@@ -32,6 +32,7 @@ struct MSALNativeAuthInternalConfiguration {
     let clientId: String
     let authority: MSIDCIAMAuthority
     let challengeTypes: [MSALNativeAuthInternalChallengeType]
+    let capabilities: [MSALNativeAuthInternalCapability]?
     let redirectUri: String?
     var sliceConfig: MSALSliceConfig?
 
@@ -39,6 +40,7 @@ struct MSALNativeAuthInternalConfiguration {
         clientId: String,
         authority: MSALCIAMAuthority,
         challengeTypes: MSALNativeAuthChallengeTypes,
+        capabilities: MSALNativeAuthCapabilities?,
         redirectUri: String?) throws {
         self.clientId = clientId
         self.authority = try MSIDCIAMAuthority(
@@ -47,6 +49,7 @@ struct MSALNativeAuthInternalConfiguration {
             context: MSALNativeAuthRequestContext()
         )
         self.challengeTypes = MSALNativeAuthInternalConfiguration.getInternalChallengeTypes(challengeTypes)
+        self.capabilities = MSALNativeAuthInternalConfiguration.getInternalCapabilities(capabilities)
         self.redirectUri = redirectUri
     }
 
@@ -65,5 +68,21 @@ struct MSALNativeAuthInternalConfiguration {
 
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
+    }
+    
+    private static func getInternalCapabilities(
+        _ capabilities: MSALNativeAuthCapabilities?
+    ) -> [MSALNativeAuthInternalCapability]? {
+        guard let capabilities else { return nil }
+        var internalCapabilities: [MSALNativeAuthInternalCapability] = []
+        
+        if capabilities.contains(.mfaRequired) {
+            internalCapabilities.append(.mfaRequired)
+        }
+        
+        if capabilities.contains(.registrationRequired) {
+            internalCapabilities.append(.registrationRequired)
+        }
+        return internalCapabilities
     }
 }
--- a/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
+++ b/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
@@ -28,7 +28,7 @@ struct MSALNativeAuthInternalConfiguration {
     var challengeTypesString: String {
         return challengeTypes.map { $0.rawValue }.joined(separator: " ")
     }
-    
+
     var capabilitiesString: String? {
         return capabilities?.map { $0.rawValue }.joined(separator: " ")
     }
@@ -73,17 +73,17 @@ struct MSALNativeAuthInternalConfiguration {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
     }
-    
+
     private static func getInternalCapabilities(
         _ capabilities: MSALNativeAuthCapabilities?
     ) -> [MSALNativeAuthInternalCapability]? {
         guard let capabilities else { return nil }
         var internalCapabilities: [MSALNativeAuthInternalCapability] = []
-        
+
         if capabilities.contains(.mfaRequired) {
             internalCapabilities.append(.mfaRequired)
         }
-        
+
         if capabilities.contains(.registrationRequired) {
             internalCapabilities.append(.registrationRequired)
         }
--- a/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
+++ b/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
@@ -24,7 +24,7 @@
 
 @_implementationOnly import MSAL_Private
 
-struct MSALNativeAuthConfiguration {
+struct MSALNativeAuthInternalConfiguration {
     var challengeTypesString: String {
         return challengeTypes.map { $0.rawValue }.joined(separator: " ")
     }
@@ -38,7 +38,7 @@ struct MSALNativeAuthConfiguration {
     init(
         clientId: String,
         authority: MSALCIAMAuthority,
-        challengeTypes: [MSALNativeAuthInternalChallengeType],
+        challengeTypes: MSALNativeAuthChallengeTypes,
         redirectUri: String?) throws {
         self.clientId = clientId
         self.authority = try MSIDCIAMAuthority(
@@ -46,7 +46,24 @@ struct MSALNativeAuthConfiguration {
             validateFormat: false,
             context: MSALNativeAuthRequestContext()
         )
-        self.challengeTypes = challengeTypes
+        self.challengeTypes = MSALNativeAuthInternalConfiguration.getInternalChallengeTypes(challengeTypes)
         self.redirectUri = redirectUri
+    }
+
+    private static func getInternalChallengeTypes(
+        _ challengeTypes: MSALNativeAuthChallengeTypes
+    ) -> [MSALNativeAuthInternalChallengeType] {
+        var internalChallengeTypes = [MSALNativeAuthInternalChallengeType]()
+
+        if challengeTypes.contains(.OOB) {
+            internalChallengeTypes.append(.oob)
+        }
+
+        if challengeTypes.contains(.password) {
+            internalChallengeTypes.append(.password)
+        }
+
+        internalChallengeTypes.append(.redirect)
+        return internalChallengeTypes
     }
 }
--- a/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
+++ b/MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
@@ -28,6 +28,10 @@ struct MSALNativeAuthInternalConfiguration {
     var challengeTypesString: String {
         return challengeTypes.map { $0.rawValue }.joined(separator: " ")
     }
+    
+    var capabilitiesString: String? {
+        return capabilities?.map { $0.rawValue }.joined(separator: " ")
+    }
 
     let clientId: String
     let authority: MSIDCIAMAuthority
--- a/MSAL/src/native_auth/network/MSALNativeAuthInternalCapability.swift
+++ b/MSAL/src/native_auth/network/MSALNativeAuthInternalCapability.swift
@@ -1,0 +1,28 @@
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+enum MSALNativeAuthInternalCapability: String, Decodable {
+    case mfaRequired = "mfa_required"
+    case registrationRequired = "registration_required"
+}
--- a/MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
+++ b/MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
@@ -22,23 +22,24 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-#ifndef MSALNativeAuthChallengeTypes_h
-#define MSALNativeAuthChallengeTypes_h
+#ifndef MSALNativeAuthCapabilities_h
+#define MSALNativeAuthCapabilities_h
 
 #import <Foundation/Foundation.h>
 
-/// The set of challenge types that an application wishes to support for Native Auth operations.
+/// The set of capabilities that an application wishes to support for Native Auth operations.
 ///
 /// Valid options are:
-/// * OOB: The application can support asking a user to supply a verification code that is sent by email.
-/// * Password: The application can support asking a user to supply a password
+/// * MFARequired: The application can accommodate the challenge type specified by the user when MFA is required.
+/// * RegistrationRequired: The application can accommodate the challenge type specified by the user
+/// when registering a new strong authentication method.
 
-typedef NS_OPTIONS(NSInteger, MSALNativeAuthChallengeTypes) {
-    /// Specifies if the Challenge Type is OOB
-    MSALNativeAuthChallengeTypeOOB          = 1 << 0,
+typedef NS_OPTIONS(NSInteger, MSALNativeAuthCapabilities) {
+    /// Specifies that the challenge type are supported when MFA is required
+    MSALNativeAuthCapabilityMFARequired          = 1 << 0,
     
-    /// Specifies if the Challenge Type is Password
-    MSALNativeAuthChallengeTypePassword     = 1 << 1
+    /// Specifies that the challenge type are supported when the registration of a new strong authentication method is required
+    MSALNativeAuthCapabilityRegistrationRequired     = 1 << 1
 };
 
-#endif /* MSALNativeAuthChallengeTypes_h */
+#endif /* MSALNativeAuthCapabilities_h */
--- a/MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
+++ b/MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
@@ -32,8 +32,7 @@
 /// Valid options are:
 /// * MFARequired: The application can accommodate the challenge type specified by the user when MFA is required.
 /// * RegistrationRequired: The application can accommodate the challenge type specified by the user
-/// when registering a new strong authentication method.
-
+/// when registering a new strong authentication method is required.
 typedef NS_OPTIONS(NSInteger, MSALNativeAuthCapabilities) {
     /// Specifies that the challenge type are supported when MFA is required
     MSALNativeAuthCapabilityMFARequired          = 1 << 0,
--- a/MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
+++ b/MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
@@ -30,14 +30,14 @@
 /// The set of capabilities that an application wishes to support for Native Auth operations.
 ///
 /// Valid options are:
-/// * MFARequired: The application can accommodate the challenge type specified by the user when MFA is required.
-/// * RegistrationRequired: The application can accommodate the challenge type specified by the user
+/// * MFARequired: The application can accommodate the associated challenge type(s) specified by the user when MFA is required.
+/// * RegistrationRequired: The application can accommodate the associated challenge type(s) specified by the user
 /// when registering a new strong authentication method is required.
 typedef NS_OPTIONS(NSInteger, MSALNativeAuthCapabilities) {
-    /// Specifies that the challenge type are supported when MFA is required
+    /// Specifies that the associated challenge type(s) are supported when MFA is required
     MSALNativeAuthCapabilityMFARequired          = 1 << 0,
     
-    /// Specifies that the challenge type are supported when the registration of a new strong authentication method is required
+    /// Specifies that the associated challenge type(s) are supported when the registration of a new strong authentication method is required
     MSALNativeAuthCapabilityRegistrationRequired     = 1 << 1
 };
 
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
@@ -104,21 +104,4 @@ extension MSALNativeAuthPublicClientApplication {
             )
         )
     }
-
-    static func getInternalChallengeTypes(
-        _ challengeTypes: MSALNativeAuthChallengeTypes
-    ) -> [MSALNativeAuthInternalChallengeType] {
-        var internalChallengeTypes = [MSALNativeAuthInternalChallengeType]()
-
-        if challengeTypes.contains(.OOB) {
-            internalChallengeTypes.append(.oob)
-        }
-
-        if challengeTypes.contains(.password) {
-            internalChallengeTypes.append(.password)
-        }
-
-        internalChallengeTypes.append(.redirect)
-        return internalChallengeTypes
-    }
 }
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
@@ -121,21 +121,4 @@ extension MSALNativeAuthPublicClientApplication {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
     }
-
-    static func convertChallengeTypes(
-        _ internalChallengeTypes: [MSALNativeAuthInternalChallengeType]
-    ) -> MSALNativeAuthChallengeTypes {
-        var challenges: MSALNativeAuthChallengeTypes = []
-        for challenge in internalChallengeTypes {
-            switch challenge {
-            case .oob:
-                challenges.insert(.OOB)
-            case .password:
-                challenges.insert(.password)
-            default:
-                break
-            }
-        }
-        return challenges
-    }
 }
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -39,7 +39,7 @@ import Foundation
 ///             challengeTypes: [.OOB]
 ///          )
 ///         nativeAuth = try MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: config)
-///        print("Initialised Native Auth successfully.")
+///         print("Initialised Native Auth successfully.")
 ///     } catch {
 ///         print("Unable to initialize MSAL \(error)")
 ///     }
--- a/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
+++ b/MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
@@ -30,14 +30,15 @@ import Foundation
 /// to the initialiser method.
 ///
 /// For example:
-// TODO: change this comment
+
 /// <pre>
 ///     do {
-///         nativeAuth = try MSALNativeAuthPublicClientApplication(
+///         let config = try MSALNativeAuthPublicClientApplicationConfig(
 ///             clientId: "Enter_the_Application_Id_Here",
 ///             tenantSubdomain: "Enter_the_Tenant_Subdomain_Here",
 ///             challengeTypes: [.OOB]
-///        )
+///          )
+///         nativeAuth = try MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: config)
 ///        print("Initialised Native Auth successfully.")
 ///     } catch {
 ///         print("Unable to initialize MSAL \(error)")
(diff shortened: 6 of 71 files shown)


MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
```diff
@@ -0,0 +1,92 @@
...
+        self.capabilities = MSALNativeAuthInternalConfiguration.getInternalCapabilities(capabilities)
+        self.redirectUri = redirectUri
+    }
+
+    private static func getInternalChallengeTypes(
+        _ challengeTypes: MSALNativeAuthChallengeTypes
+    ) -> [MSALNativeAuthInternalChallengeType] {
+        var internalChallengeTypes = [MSALNativeAuthInternalChallengeType]()
+
+        if challengeTypes.contains(.OOB) {
```
Yuki-YuXin, line 65: Does here include lower and upper case handling?

MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
```diff
@@ -0,0 +1,92 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@_implementationOnly import MSAL_Private
+
+struct MSALNativeAuthInternalConfiguration {
```
Yuki-YuXin, line 27: Is there any specific reason to rename the existing `MSALNativeAuthConfiguration.swift` -> `MSALNativeAuthInternalConfiguration`?

Yuki-YuXin: Appears like from `native_auth/configuration/MSALNativeAuthConfiguration.swift` to  `native_auth/configuration/MSALNativeAuthInternalConfiguration.swift` and a new folder `native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift`

MSAL/src/native_auth/client_application_wrapper/silent_token/MSALNativeAuthSilentTokenProvider.swift
```diff
@@ -27,10 +27,8 @@ class MSALNativeAuthSilentTokenProvider: MSALNativeAuthSilentTokenProviding {
 
     private let application: MSALNativeAuthPublicClientApplication?
 
-    init(
-        configuration config: MSALPublicClientApplicationConfig,
-        challengeTypes: MSALNativeAuthChallengeTypes) throws {
-            self.application = try? MSALNativeAuthPublicClientApplication(configuration: config, challengeTypes: challengeTypes)
+    init(configuration: MSALNativeAuthPublicClientApplicationConfig) throws {
+        self.application = try? MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: configuration)
         }
```
spetrescu84: nit: I know this is not part of this PR, indentation

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
```diff
@@ -22,31 +22,24 @@
...
-    let authority: MSIDCIAMAuthority
-    let challengeTypes: [MSALNativeAuthInternalChallengeType]
-    let redirectUri: String?
-    var sliceConfig: MSALSliceConfig?
+/// The set of capabilities that an application wishes to support for Native Auth operations.
+///
+/// Valid options are:
+/// * MFARequired: The application can accommodate the challenge type specified by the user when MFA is required.
+/// * RegistrationRequired: The application can accommodate the challenge type specified by the user
+/// when registering a new strong authentication method.
```
spetrescu84: Nit: 
`/// when registering a new strong authentication method.`
Imo should be:
`/// when registering a new strong authentication method is required.`

MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
spetrescu84: Based on the naming of the new capabilities class `MSALNativeAuthInternalCapability` we can rename this one to match.

MSAL/src/native_auth/network/MSALNativeAuthInternalCapability.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+enum MSALNativeAuthInternalCapability: String, Decodable {
```
spetrescu84, line 25: I would have put this one in the configuration folder, same as `MSALNativeAuthInternalConfiguration`. Not sure how much it relates to network

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
```diff
@@ -22,31 +22,24 @@
...
-            url: authority.url,
-            validateFormat: false,
-            context: MSALNativeAuthRequestContext()
-        )
-        self.challengeTypes = challengeTypes
-        self.redirectUri = redirectUri
-    }
-}
+typedef NS_OPTIONS(NSInteger, MSALNativeAuthCapabilities) {
+    /// Specifies that the challenge type are supported when MFA is required
```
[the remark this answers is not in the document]
spetrescu84: Also should be `Specifies that the challenge type are supported when MFA is required`

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
```diff
@@ -22,31 +22,24 @@
...
+    MSALNativeAuthCapabilityMFARequired          = 1 << 0,
+    
+    /// Specifies that the challenge type are supported when the registration of a new strong authentication method is required
```
spetrescu84: nit: this one should be: `Specifies that the capability type is supported when the registration of a new strong authentication method is required`

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -33,11 +33,12 @@ import Foundation
 
 /// <pre>
 ///     do {
-///         nativeAuth = try MSALNativeAuthPublicClientApplication(
+///         let config = try MSALNativeAuthPublicClientApplicationConfig(
 ///             clientId: "Enter_the_Application_Id_Here",
 ///             tenantSubdomain: "Enter_the_Tenant_Subdomain_Here",
 ///             challengeTypes: [.OOB]
-///        )
+///          )
```
spetrescu84, line 40: nit: spacing here and on the next line

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -50,99 +51,96 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
...
     lazy var cacheAccessor: MSALNativeAuthCacheAccessor = {
         return cacheAccessorFactory.makeCacheAccessor(tokenCache: tokenCache, accountMetadataCache: accountMetadataCache)
     }()
 
-    /// Initialize a MSALNativePublicClientApplication with a given configuration and challenge types
+    /// Initialize a MSALNativePublicClientApplication with a given configuration
     /// - Parameters:
-    ///   - config: Configuration for PublicClientApplication
-    ///   - challengeTypes: The set of capabilities that this application can support as an ``MSALNativeAuthChallengeTypes`` optionset
+    ///   - nativeAuthConfiguration: Configuration for native auth PublicClientApplication
```
spetrescu84, line 61: This does look more in line with the caller of `MSALPublicClientApplication` when initiated from Swift 
Screenshot 2025-06-03 at 10 27 08

MSAL/src/native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift
```diff
@@ -0,0 +1,92 @@
...
+// THE SOFTWARE.  
+import Foundation
+
+@objcMembers
+public final class MSALNativeAuthPublicClientApplicationConfig: MSALPublicClientApplicationConfig {
+
+    let challengeTypes: MSALNativeAuthChallengeTypes
+
+    /** The set of capabilities that this application can support as an ``MSALNativeAuthCapabilities`` optionset */
+    public var capabilities: MSALNativeAuthCapabilities?
```
spetrescu84, line 32: The plan here, to avoid breaking changes for the future is that everything that gets added new is like this one, an optional, right? 
The other option I can think of is a parameters class that gets passed to this config

MSAL/src/native_auth/public/state_machine/result/MSALNativeAuthUserAccountResult+Internal.swift
```diff
@@ -31,27 +31,31 @@ extension MSALNativeAuthUserAccountResult {
                                 claimsRequest: MSALClaimsRequest?,
                                 correlationId: UUID?,
                                 delegate: CredentialsDelegate) {
-
-        let params = MSALSilentTokenParameters(scopes: scopes, account: account)
         let context = MSALNativeAuthRequestContext(correlationId: correlationId)
+        guard let authority = try? MSALCIAMAuthority(url: configuration.authority.url) else {
```
spetrescu84, line 35: I think we didn't do this before because there was no way to have an invalid authority here as it's the same one as it was when the initial PublicClientApplication was created.

MSAL/test/integration/native_auth/common/MSALNativeAuthIntegrationBaseTests.swift
```diff
@@ -31,12 +31,13 @@ class MSALNativeAuthIntegrationBaseTests: XCTestCase {
...
     let mockAPIHandler = MockAPIHandler()
     let correlationId = UUID()
-    let config: MSALNativeAuthConfiguration = try! MSALNativeAuthConfiguration(
+    let config: MSALNativeAuthInternalConfiguration = try! MSALNativeAuthInternalConfiguration(
         clientId: UUID().uuidString,
         authority: MSALCIAMAuthority(url:
                                         URL(string: (ProcessInfo.processInfo.environment["authorityURL"] ?? "<mock api url not set>") + "/testTenant")!),
                                                                                
-        challengeTypes: [.password, .oob, .redirect],
+        challengeTypes: [.password, .OOB],
```
spetrescu84, line 39: nit: `.redirect` was never actually needed because it's added inside the sdk right?

MSAL/test/integration/native_auth/end_to_end/reset_password/ResetPasswordDelegateSpies.swift
```diff
@@ -140,7 +140,7 @@ class ResetPasswordResendCodeDelegateSpy: ResetPasswordResendCodeDelegate {
         self.error = error
     }
 
-    func onResetPasswordResendCodeCodeRequired(newState: ResetPasswordCodeRequiredState, sentTo: String, channelTargetType: MSALNativeAuthChannelType, codeLength: Int) {
+    func onResetPasswordResendCodeRequired(newState: ResetPasswordCodeRequiredState, sentTo: String, channelTargetType: MSALNativeAuthChannelType, codeLength: Int) {
```
spetrescu84, line 143: Good catch, yeah, we didn't actually ever ran this test e2e

spetrescu84: nit: maybe change the name of the `onResetPasswordResendCodeCodeRequiredCalled` variable also

MSAL/test/unit/native_auth/mock/MSALNativeAuthConfigStubs.swift
```diff
@@ -33,13 +33,15 @@ enum ErrorMock: Error {
...
 
-    static var configuration: MSALNativeAuthConfiguration {
+    static var configuration: MSALNativeAuthInternalConfiguration {
         try! .init(
             clientId: DEFAULT_TEST_CLIENT_ID,
             authority: try! .init(
                 url: URL(string: DEFAULT_TEST_AUTHORITY)!
             ),
-            challengeTypes: [.redirect], redirectUri: nil
+            challengeTypes: [.password],
```
spetrescu84, line 42: Should this one be empty?

MSAL/test/unit/native_auth/network/MSALNativeAuthInternalConfigurationTest.swift
```diff
@@ -0,0 +1,56 @@
...
+    func test_challengeTypeRedirect_isAddedAutomatically() {
+        let sut = try? MSALNativeAuthInternalConfiguration(
+            clientId: DEFAULT_TEST_CLIENT_ID,
+            authority: try! .init(
+                url: URL(string: DEFAULT_TEST_AUTHORITY)!
+            ),
+            challengeTypes: [.password],
+            capabilities: nil,
+            redirectUri: nil)
+        XCTAssertEqual(sut?.challengeTypes, [.password, .redirect])
```
spetrescu84, line 41: We should also check that `challengeTypesString` is equal to "password redirect" to prevent unwanted changes to that string.

MSAL/test/unit/native_auth/network/MSALNativeAuthInternalConfigurationTest.swift
```diff
@@ -0,0 +1,56 @@
...
+    func test_capabilities_AreAddedCorrectlyToTheConfig() {
+        let sut = try? MSALNativeAuthInternalConfiguration(
+            clientId: DEFAULT_TEST_CLIENT_ID,
+            authority: try! .init(
+                url: URL(string: DEFAULT_TEST_AUTHORITY)!
+            ),
+            challengeTypes: [.password],
+            capabilities: [.mfaRequired, .registrationRequired],
+            redirectUri: nil)
+        XCTAssertEqual(sut?.capabilities, [.mfaRequired, .registrationRequired])
```
spetrescu84, line 54: We should also check that `capabilitiesString` is equal to "mfa_required registartion_required" to prevent unwanted changes to that string.

MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
nilo-ms: I renamed this class because I added a new class "MSALNativeAuthPublicClientApplicationConfig". To make things clearer I renamed this existing internal class as "Internal", so also the name clarify that this is not public

MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
nilo-ms: "challengeTypes" is not a string, but a "NS_Options" (a sort of enum) so we don't need to check for upper or lower case

MSAL/src/native_auth/network/MSALNativeAuthInternalCapability.swift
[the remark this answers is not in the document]
nilo-ms: It can technically still be returned. I don't see big of advantages to remove it, so I preferred to keep it

nilo-ms: It is related to network because this enum is used to parse the value from the HTTP response

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -50,99 +51,96 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
...
+            nativeAuthConfiguration.bypassRedirectURIValidation = true
         }
 
-        try super.init(configuration: config)
+        try super.init(configuration: nativeAuthConfiguration)
+    }
+
+    /// Initialize a MSALNativePublicClientApplication with a given configuration and challenge types
+    /// - Parameters:
+    ///   - config: Configuration for PublicClientApplication
```
Yuki-YuXin, line 92:
```
    /// - Parameters:
    ///   - nativeAuthConfiguration: Configuration for native auth PublicClientApplication
```

Maybe keep alignment with the comments above

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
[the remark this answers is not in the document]
nilo-ms: Sorry, I'm not getting what is the suggestion here. I think the current comment, that is "Specifies that the challenge type are supported when MFA is required", clarifies what "mfaRequired" capability is

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
nilo-ms: I don't think this is what the "registration required" capabilities is. Capability is strictly related to "challenge type". Registration required, means that the "challenge type" used during the SDK initialisation are supported also when a registration of a new strong authentication method is required.

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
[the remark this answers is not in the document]
spetrescu84: Sorry, wrote it wrong, I meant `Specifies that the **capability** is supported when MFA is required` because we are talking about capabilities here not challenge types.
Maybe even:
`Specifies that MFA Required **capability** is supported` 
if that makes more sense.

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
spetrescu84: Sorry, is this not connected to this value in the config?
```
 /** The set of capabilities that this application can support as an ``MSALNativeAuthCapabilities`` optionset */
public var capabilities: MSALNativeAuthCapabilities?
```
If so, why do you refer to it as the challenge type?
The developer doesn't have to add the challenge_type "mfa_required" unless I'm mistaken somehow?

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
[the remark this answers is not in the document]
nilo-ms: Thanks, removed

nilo-ms: Thanks, comment updated

MSAL/src/native_auth/client_application_wrapper/silent_token/MSALNativeAuthSilentTokenProvider.swift
nilo-ms: indentation fixed, thanks

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
nilo-ms: Those two looked ok, but I fixed indentation at line 42

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
nilo-ms: I had to use "nativeAuthConfiguration"because I can't use "configuration". This is because Objective-C use only parameter name to distinguish methods of the same class and not parameter type

MSAL/src/native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift
nilo-ms: Correct, same as other MSALPublicClientApplication init.
Using a parameter class to initialise a configuration class seems a repetition because this configuration acts as parameter class itself

MSAL/src/native_auth/public/state_machine/result/MSALNativeAuthUserAccountResult+Internal.swift
nilo-ms: We didn't do this before because we used to use the initialiser where the authority was optional. Now, we use the init of the native auth configuration where the authority is not optional. Anyway, this guard should always be true

MSAL/test/integration/native_auth/common/MSALNativeAuthIntegrationBaseTests.swift
nilo-ms: Correct. 
Before challenge type needed here was the internal type and the PublicClientApplication used to convert public challenge type to internal ones. Now, this conversion is done internally in the configuration class. Where, I think, it's more accurate instead of having internal static methods in the PublicClientApplication class

MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
```diff
@@ -0,0 +1,92 @@
...
+    var challengeTypesString: String {
+        return challengeTypes.map { $0.rawValue }.joined(separator: " ")
+    }
+
+    var capabilitiesString: String? {
```
Yuki-YuXin, line 32: How about Pre-compute string properties
```
struct MSALNativeAuthInternalConfiguration {
    let clientId: String
    let authority: MSIDCIAMAuthority
    let challengeTypes: [MSALNativeAuthInternalChallengeType]
    let capabilities: [MSALNativeAuthInternalCapability]?
    let redirectUri: String?
    let sliceConfig: MSALSliceConfig?
    
    // Pre-computed strings for performance
    let challengeTypesString: String
    let capabilitiesString: String?
    
    init(
        clientId: String,
        authority: MSALCIAMAuthority,
        challengeTypes: MSALNativeAuthChallengeTypes,
        capabilities: MSALNativeAuthCapabilities?,
        redirectUri: String?,
        sliceConfig: MSALSliceConfig? = nil) throws {
       
        self.clientId = clientId
        self.authority = try MSIDCIAMAuthority(
            url: authority.url,
            validateFormat: false,
            context: MSALNativeAuthRequestContext()
        )
        
        let internalChallengeTypes = Self.getInternalChallengeTypes(challengeTypes)
        let internalCapabilities = Self.getInternalCapabilities(capabilities)
        
        self.challengeTypes = internalChallengeTypes
        self.capabilities = internalCapabilities
        self.redirectUri = redirectUri
        self.sliceConfig = sliceConfig
        
        self.challengeTypesString = internalChallengeTypes.map { $0.rawValue }.joined(separator: " ")
        self.capabilitiesString = internalCapabilities?.map { $0.rawValue }.joined(separator: " ")
    }
}
```

MSAL/test/integration/native_auth/end_to_end/reset_password/ResetPasswordDelegateSpies.swift
nilo-ms: Renamed, thanks

MSAL/test/unit/native_auth/network/MSALNativeAuthInternalConfigurationTest.swift
nilo-ms: This specific tests is covered in another test class.
As a configuration class it should not care in which string the capabilities are translated. It should only care that the correct capabilities are added to the config.
But, in the "parameters" class tests I'm testing that the expected string is added when certains "capabilities" to the request body.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
```diff
@@ -121,21 +121,4 @@ extension MSALNativeAuthPublicClientApplication {
         internalChallengeTypes.append(.redirect)
         return internalChallengeTypes
```
Yuki-YuXin: Still need to keep `static func getInternalChallengeTypes` here?

MSAL/src/native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift
```diff
@@ -0,0 +1,92 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.  
+import Foundation
+
+@objcMembers
+public final class MSALNativeAuthPublicClientApplicationConfig: MSALPublicClientApplicationConfig {
+
+    let challengeTypes: MSALNativeAuthChallengeTypes
```
Yuki-YuXin, line 29: `challengeTypes` is immutable (`let`) but `capabilities` is mutable (`var`). No clear reason why `capabilities` needs to be mutable while `challengeTypes` doesn't. 

How about adding a new initializer that includes capabilities
```
public init(clientId: String, 
           authority: MSALCIAMAuthority, 
           challengeTypes: MSALNativeAuthChallengeTypes,
           capabilities: MSALNativeAuthCapabilities? = nil) {
    self.challengeTypes = challengeTypes
    self.capabilities = capabilities
    super.init(clientId: clientId, redirectUri: nil, authority: authority, 
              nestedAuthBrokerClientId: nil, nestedAuthBrokerRedirectUri: nil)
}
```

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
Yuki-YuXin: Appears like this internal file doesn't exist now with converting methods moved to `MSALNativeAuthPublicClientApplicationConfig`

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -99,27 +100,22 @@ class MSALNativeAuthEndToEndBaseTestCase: XCTestCase {
                 format: customAuthorityURLFormat
             )
             
-            let authority = try? MSALCIAMAuthority(
+            let authority = try! MSALCIAMAuthority(
```
Yuki-YuXin, line 103: Double force unwrapping (`try!` and `!`) here.
How about replacing with proper error handling

```
do {
    guard let url = URL(string: customSubdomain) else {
        XCTFail("Invalid authority URL format: \(customSubdomain)")
        return nil
    }
    
    let authority = try MSALCIAMAuthority(url: url, validateFormat: false)
    let configuration = MSALNativeAuthPublicClientApplicationConfig(
        clientId: clientId, 
        authority: authority, 
        ...
    )
    configuration.capabilities = capabilities
    
    return try MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: configuration)
} catch {
    XCTFail("Failed to create MSAL authority: \(error)")
    return nil
}
```

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -99,27 +100,22 @@ class MSALNativeAuthEndToEndBaseTestCase: XCTestCase {
...
-                authority: authority
-            )
-
+            let configuration = MSALNativeAuthPublicClientApplicationConfig(clientId: clientId, authority: authority, challengeTypes: challengeTypes)
+            configuration.capabilities = capabilities
+           
             return try? MSALNativeAuthPublicClientApplication(
-                configuration: configuration,
-                challengeTypes: challengeTypes
+                nativeAuthConfiguration: configuration
```
Yuki-YuXin, line 112: Using `try?` silently swallows initialization errors

How about
```
do {
    return try MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: configuration)
} catch {
    XCTFail("Failed to create MSALNativeAuthPublicClientApplication: \(error)")
    return nil
}
```

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
```diff
@@ -99,27 +100,22 @@ class MSALNativeAuthEndToEndBaseTestCase: XCTestCase {
...
             )
             
-            let configuration = MSALPublicClientApplicationConfig(
-                clientId: clientId,
-                redirectUri: nil,
-                authority: authority
-            )
-
+            let configuration = MSALNativeAuthPublicClientApplicationConfig(clientId: clientId, authority: authority, challengeTypes: challengeTypes)
+            configuration.capabilities = capabilities
```
Yuki-YuXin, line 109:
```
let configuration = MSALNativeAuthPublicClientApplicationConfig(clientId: clientId, authority: authority, challengeTypes: challengeTypes)
configuration.capabilities = capabilities
```
Two-step configuration setup is error-prone

How about using single-step initialization when possible it supports.
```
let configuration = MSALNativeAuthPublicClientApplicationConfig(
    clientId: clientId,
    authority: authority,
    challengeTypes: challengeTypes,
    capabilities: capabilities 
)

```

MSAL/src/native_auth/configuration/MSALNativeAuthInternalConfiguration.swift
nilo-ms: Yes, it is another approach, better for perfomance. I used the same existing approach we had for challenge types. I don't think performance is a concern here, so I don't think one approach is much better than the other one

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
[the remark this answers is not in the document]
nilo-ms: I think there is a misunderstanding about what a "capability" is. Let's discuss this offline.

MSAL/test/unit/native_auth/network/MSALNativeAuthRequestConfiguratorTests.swift
```diff
@@ -489,7 +489,7 @@ final class MSALNativeAuthRequestConfiguratorTests: XCTestCase {
     }
 
     func test_refreshToken_getsConfiguredSuccessfully() throws {
-        XCTAssertNoThrow(config = try .init(clientId: DEFAULT_TEST_CLIENT_ID, authority: MSALCIAMAuthority(url: baseUrl), challengeTypes: [.password], redirectUri: nil))
+        XCTAssertNoThrow(config = try .init(clientId: DEFAULT_TEST_CLIENT_ID, authority: MSALCIAMAuthority(url: baseUrl), challengeTypes: [.password], capabilities: nil, redirectUri: nil))
```
Yuki-YuXin, line 492: A quick question, could the init ( capabilities ) set a default value nil?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication+Internal.swift
nilo-ms: Good catch, forgot to remove this method from here.

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
nilo-ms: This comment is aligned with the comment at line 59, right?

MSAL/src/native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift
nilo-ms: This explained in the technical design at page 4. If this is not clear, we can discuss this offline

MSAL/test/unit/native_auth/mock/MSALNativeAuthConfigStubs.swift
nilo-ms: No particular reason why this should be empty. We don't need to test that the redirect is added automatically i this case. This is just a stub with a value for the challenge types

MSAL/test/unit/native_auth/network/MSALNativeAuthInternalConfigurationTest.swift
nilo-ms: Answered in another comment

MSAL/src/native_auth/network/MSALNativeAuthInternalCapability.swift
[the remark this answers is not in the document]
Yuki-YuXin: I remembered that this was the legacy name for OOB and realized that it had only been used in test now.

MSAL/test/unit/native_auth/network/MSALNativeAuthInternalConfigurationTest.swift
spetrescu84: All right, I saw that test there, but figured we should test it in the class. I'm ok either way.

mustafamizrak (approved):

MSAL/src/native_auth/public/MSALNativeAuthCapabilities.h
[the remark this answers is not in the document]
spetrescu84: sorted offline

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: We used to have single step initialization, but this approach is not extensible for Objective-C. Adding a new optional method will result in addition of a new constructor and deprecate the previous one. So, we moved to multiple steps configuration.

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: It's another approach that makes the test longer (do/catch is at least 4 lines), but we can print the error. In case the initialisation fails, the test will fail. Then the dev can debug. I personally prefer the current approach for testing

MSAL/test/unit/native_auth/network/MSALNativeAuthRequestConfiguratorTests.swift
nilo-ms: Yes, it can, but why?

MSAL/test/integration/native_auth/end_to_end/MSALNativeAuthEndToEndBaseTestCase.swift
nilo-ms: The code you are proposing is not applicable here. We use different config initialiser based on the value of "customAuthorityURLFormat". 
We never use force unwrap "!" in code, but it is acceptable in testing where the code is never expected to fail. See the "!" at line 104 for example. The authority initialisation will not fail if the URL is not nil and valid, and that is checked at line 96. So, It's acceptable here. TO remove the "!" I'd need to add an inner if let and will make the test less readable

MSAL/src/native_auth/network/MSALNativeAuthInternalCapability.swift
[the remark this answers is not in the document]
nilo-ms: Even if this is out of scope for this PR, I removed OTP as challenge type value.

MSAL/src/native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift
Yuki-YuXin: The technical design at page 4 claimed that challengeType is mandatory while capabilities is optional, slightly different from the immutable and mutable.

MSAL/test/unit/native_auth/network/MSALNativeAuthRequestConfiguratorTests.swift
Yuki-YuXin: Since the  capabilities is optional, set a default value prevent the changes of the existing init in the code.

nilo-ms: I didn't add a default value because I wanted that the compiler highlighted where the configuration was initialised, so I could make sure that the correct value of the capabilities value was used.

MSAL/src/native_auth/public/configuration/MSALNativeAuthPublicClientApplicationConfig.swift
nilo-ms: If I make the capabilities not mutable ("let"), I'd need to add this parameter as part of the initialiser parameter list. We decided to do not include optional parameters as part of the init parameter list to make the SDK more extensible

PR #2687, opened 2025-08-22
[iOS SDK] Update Email OTP MFA to match EC implementation

spetrescu84:
## Proposed changes

This PR updates the SDK to match the latest flow from EC.

In this new flow, the developer must always supply an existing Method Id to the `/oauth2/v2.0/challenge` endpoint which means once the `.strongAuthRequired` error is received from any endpoint, the `/oauth2/v2.0/introspect` endpoint needs to be called to retrieve the methods which are automatically returned to be used by the developer.

Furthermore whenever calling the `/token` endpoint is called with an MFA Email OTP code, the grant type should be `mfa_oob`

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


MSAL/MSAL.xcodeproj/project.pbxproj
```diff
@@ -288,18 +288,14 @@
 		289E156D2948EB8A006104D9 /* MSALNativeAuthCacheAccessor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 289E156C2948EB8A006104D9 /* MSALNativeAuthCacheAccessor.swift */; };
```
nilo-ms, line 288: @spetrescu84 can you please update the changelog file too?

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -438,54 +448,40 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
                     self?.stopTelemetryEvent(event, context: context, delegateDispatcherResult: result)
                 })
         case .introspectRequired:
```
nilo-ms: I think this type of result can't be returned by the API, so we can get rid of it

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
```diff
@@ -788,15 +744,23 @@ final class MSALNativeAuthSignInController: MSALNativeAuthTokenController, MSALN
...
-                controller: self,
-                scopes: scopes ?? [],
-                claimsRequestJson: claimsRequestJson,
-                continuationToken: continuationToken,
-                correlationId: telemetryInfo.context.correlationId()
-            )
-            MSALNativeAuthLogger.log(level: .info, context: telemetryInfo.context, format: "Multi factor authentication required")
-            return .awaitingMFA(state)
+            MSALNativeAuthLogger.log(level: .info, context: telemetryInfo.context, format: "Strong authentication required.")
+            let response = await performAndValidateIntrospectRequest(continuationToken: continuationToken, context: telemetryInfo.context)
```
nilo-ms, line 711: it seems this code is very similar to code at line 280 and 450. Do you think can we create a private func and reuse it?

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
spetrescu84: I wasn't sure on this one, removed

MSAL/MSAL.xcodeproj/project.pbxproj
spetrescu84: Updated

MSAL/src/native_auth/controllers/sign_in/MSALNativeAuthSignInController.swift
spetrescu84: It is indeed and I tried to reuse it but the problem is that the `handleTokenResponse` (line 280) returns a `HandleTokenResult` and `submitPassword` (line 450) returns a `SignInSubmitPasswordControllerResponse`. Neither of these are compatible with each other that's why:
-Line 280 returns 
```
return .init(.awaitingMFA(authMethods: authMethods.map({$0.toPublicAuthMethod()}),
                                          newState: newState),
                             correlationId: context.correlationId(),
                             telemetryUpdate: { [weak self] result in
                             })
```
-Line 450 returns `return .awaitingMFA(authMethods.map({$0.toPublicAuthMethod()}), newState)`
Each of them also return different errors.
I don't see a way to homogenise these.
The same issue happened with `let jitIntrospectResponse = await jitController.getJITAuthMethods` below

nilo-ms: We could return the same response as the introspect request, but it doesn't seem worth the refactoring to me.

nilo-ms (approved):

antrix1989 (approved):

PR #2862, opened 2026-02-07
[minor][Feature]: Provide public api to set custom headers for CIAM requests

antrix1989:
## Proposed changes

See proposal [here](https://microsoft.sharepoint-df.com/:w:/t/AuthClientDesignDocs/cQoQTWJ1PJnKQ4eXyhwj3D7eEgUDJF9x3grLCArHX5yQ9dxCMw?isSPOFile=1&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Csedemche%40microsoft.com&wdExp=TEAMS-TREATMENT&web=1&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI1MC8yNjAyMjcwNDIwOSJ9).
Core PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/1723

Proposal summary:

Adopt a narrowly scoped extensibility model that enables per‑request, headers‑only modification with controlled naming and merge semantics. The SDK will allow applications to attach custom `x‑*` headers on a per‑request basis, injected as late as possible in the networking pipeline, while preserving MSAL ownership of request structure and security‑critical invariants. The following headers sould not be allowed: 

- “x-ms-” 
- “x-client-” 
- “x-broker-” 
- x-app- 

This approach intentionally avoids global configuration and full request mutation, balancing CIAM fraud‑signal requirements (exact header names, precise scoping) with MSAL’s responsibility to protect authentication, telemetry, and correlation semantics. 

Steps to integrate:
1. Implement MSALNativeAuthRequestInterceptor protocol.
2. Register requestInterceptor.
3. In `addAdditionalHeaderFields`, inspect the outgoing request URL and use it to apply headers conditionally per endpoint.
4. **Always invoke completionBlock**, either with the nil or with the list of headers.
5. All custom header field names must start with the `"x-"` prefix.
6. The prefixes `"x-ms-"`, `"x-client-"`, `"x-broker-"`, and `"x-app-"` are reserved and must not be used. If provided, the SDK will ignore them.

Example:

  ```
extension EmailAndPasswordViewController: MSALNativeAuthRequestInterceptor
{
    /// Called before each native auth network request, allowing you to inject custom HTTP header fields.
    ///
    /// - Important: `completionBlock` **must always be called**, regardless of whether additional headers are needed.
    ///   - If no additional headers are required, call `completionBlock(nil)`.
    ///   - If additional headers are needed, inspect `requestUrl` to determine the request path and
    ///     call `completionBlock` with a `[String: String]` dictionary of headers to inject.
    ///
    /// - Note: All custom header field names must start with the `"x-"` prefix.
    ///   The prefixes `"x-ms-"`, `"x-client-"`, `"x-broker-"`, and `"x-app-"` are reserved and must not be used.
    ///
    /// - Parameters:
    ///   - requestUrl: The URL of the outgoing request. Use this to conditionally apply headers per endpoint.
    ///   - completionBlock: Must be called with a header dictionary, or `nil` if no extra headers are needed.
    func addAdditionalHeaderFields(_ requestUrl: URL?, completionBlock: @escaping MSALNativeAuthRequestInterceptorAddHeaderCompletionBlock)
    {
        if requestUrl?.absoluteString.contains("oauth2/v2.0/initiate") == true {
            completionBlock(
                ["value_1": "customer_header_1", // Will be ignored: doesn't start with "x-"
                 "x-client-header": "customer_header_2", // Will be ignored: starts with reserved prefix "x-client-"
                 "X-acf-sensor-data": "AKAMAI Sensor Data" // Will be added to the network request.
                ])
            return;
        }
        
        completionBlock(nil)
    }
}

class EmailAndPasswordViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()

        do {
            let config = try MSALNativeAuthPublicClientApplicationConfig(
                clientId: Configuration.clientId,
                tenantSubdomain: Configuration.tenantSubdomain,
                challengeTypes: [.OOB, .password]
            )
            
            config.requestInterceptor = self
            
            nativeAuth = try MSALNativeAuthPublicClientApplication(nativeAuthConfiguration: config)
        } catch {
            print("Unable to initialize MSAL \(error)")
            showResultText("Unable to initialize MSAL: \(error.localizedDescription)")
        }
    }

...
}
```

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### CODE REVIEW

**Review by mustafamizrak — APPROVED** [review]

File: MSAL/src/native_auth/network/MSALNativeAuthRequestConfigurator.swift [hunk 1]
```diff
@@ -320,6 +320,26 @@ class MSALNativeAuthRequestConfigurator: MSIDAADRequestConfigurator {
             )
             throw MSALNativeAuthInternalError.invalidRequest
         }
+        
+        if let interceptor = MSALNativeGlobalConfig.httpConfig.requestInterceptor {
```
shenj: I have two questions:
1. Can we use the existing `MSALNativeAuthPublicClientApplicationConfig` to let developers to configure the interceptor?
2. I am not familiar with the iOS development, so I am wondering how this configurator to access the `MSALNativeGlobalConfig` without passing it in.

antrix1989: 1. MSALNativeAuthPublicClientApplicationConfig is used to configure public client application. This is a new config that allows to configure network stack globally.
2. It is static and can be accessed anywhere in the code, i.e: MSALNativeGlobalConfig.httpConfig.requestInterceptor

shenj: I am wondering why we cannot reuse the `MSALNativeAuthPublicClientApplicationConfig` to allow users to configure the interceptor there. IMHO, I think it is better to use this existing public configuration since:
1. The dependency is clear. A developer can see exactly how the SDK is behaving by looking at this public configuration object passed at initialization stage. But if we use that static global http config, it is not immediately clear to a new developer that they need to touch a static class to inject the custom headers.
2. I am not sure whether iOS SDK allows multiple instances of public client application. If so, if we use the static global config, this setting will be not instance isolation.
3. I am not familiar with iOS development, but do you think this static global configure can be thread safety? Is it possible it may run into race conditions?
4. For testing, the settings in static global configuration may impact multiple unit tests if they are run in parallel.

mustafamizrak (commented):
How did you test this?

mustafamizrak:
This is app level setting up customer header modification, right? A customer can set for different apps with different headers and must do for all apps?

MSAL/src/native_auth/network/MSALNativeAuthRequestConfigurator.swift
antrix1989: Moved it to MSALNativeAuthPublicClientApplicationConfig

MSAL/src/native_auth/network/MSALNativeAuthRequestConfigurator.swift
```diff
@@ -320,6 +320,26 @@ class MSALNativeAuthRequestConfigurator: MSIDAADRequestConfigurator {
...
+
+    private let interceptor: MSALNativeAuthRequestInterceptor
+
+    init(interceptor: MSALNativeAuthRequestInterceptor) {
+        self.interceptor = interceptor
+    }
+
+    func addAdditionalHeaderFields(for requestUrl: URL?, with completionBlock: @escaping MSIDHttpRequestInterceptorAddHeaderCompletionBlock) {
+        interceptor.addAdditionalHeaderFields(requestUrl) { additionalHeaders in
+            completionBlock(additionalHeaders)
```
kaisong1990, line 342: Do we need error handler as well or pass back error in case?

kaisong1990 (approved):
minor question, rest looks god!

MSAL/src/native_auth/network/MSALNativeAuthRequestConfigurator.swift
antrix1989: This is just bridge from obj c to swift code, we don't need to have any handling here.

PR #3039, opened 2026-07-15
Add SSPR V2

antrix1989:
## PR Checklist (must be completed before review)

- [ ] All tests pass locally
- [ ] PR size is <= 500 LOC per PR Size Check policy
- [ ] PR is independently mergeable (no hidden dependencies)
- [ ] Appropriate reviewers are assigned
- [ ] PR reviewed by code owner (required if Copilot-generated)
- [ ] SME or Senior IC assigned where required

## PR Title Format

**Required Format:** `[Keyword1] [Keyword2]: Description`

- **Keyword1:** `major`, `minor`, or `patch` (case-insensitive)
- **Keyword2:** `feature`, `bugfix`, `engg`, or `tests` (case-insensitive)

**Examples:**
- `[MAJOR] [Feature]: new API`
- `[minor] [bugfix]: fix crash`
- `[PATCH] [tests]: add coverage`

## Proposed changes

Merge SSPR v2 branch into dev

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High – Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)
- [ ] Medium – Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)
- [ ] Small – No issues are expected. (Example: Very small bug fixes, string changes, or configuration settings changes)

PR #3052, opened 2026-07-21
Add SSPR v2 e2e basic tests

antrix1989:
## Summary
Adds `MSALNativeAuthResetPasswordV2EndToEndTests` — a new end-to-end suite covering the server-driven **V2 self-service password reset** flow (`resetPasswordV2`) via the unified delegate.

Branch is a direct descendant of `sedemche/native-auth-v2_release`, so the diff is exactly **one new test file + its project wiring** (no conflicts). It compiles against the V2 API already present in the base.

## Test coverage
- **Happy path** — `resetPasswordV2` → `onCodeRequired` → `submitCode` → `onNewPasswordRequired` → `submitNewPassword` → `onFlowCompleted` (asserts `scenario == .passwordReset`).
- **Invalid password** — `submitNewPassword` with a non-compliant password → `onFlowError` with `error.isInvalidPassword`.
- **Resend code** — `onCodeRequired` → `resendCode` → second `onCodeRequired`; retrieves two codes.
- **Email not found** — unknown username → `onFlowError` with `error.isUserNotFound`.

OTP retrieval reuses the existing mail.tm code retriever already in the base branch.

## Notes
- All tests are `XCTSkip`-gated until a V2 test tenant is enabled (keeps CI green; runnable on demand), consistent with the V1 SSPR suite.
- Wired into both **MSAL iOS Native Auth E2E Tests** and **MSAL Mac Native Auth E2E Tests** targets.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,318 @@
...
+    // Number of times a submitted OTP is retried when the server reports it as invalid (races with
+    // email delivery / an older code being read before the fresh one arrives).
+    private let codeRetryCount = 3
+
+    // SSPR – happy path: start → code required → new password required → flow completed (tokens).
+    func test_resetPasswordV2_succeeds() async throws {
+        throw XCTSkip("Skip until we can reliably test SSPR")
+        
+        guard let sut = initialisePublicClientApplication(),
+              let username = retrieveUsernameForResetPassword()
```
kaisong1990, line 47: Every test in this suite exits through `XCTSkip` before invoking the API, so this adds no executable regression coverage. Please enable the supported scenarios before treating this as E2E coverage.

antrix1989: this is expected, these test will be enable after test teanant is allowlisted on the server side.

jasoncoolmax (dismissed):

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,318 @@
...
+
+/// End-to-end tests for the V2 (server-driven) self-service-password-reset flow exposed through
+/// `MSALNativeAuthPublicClientApplication.resetPasswordV2(parameters:delegate:)`.
+///
+/// Unlike the V1 suite — which uses a different delegate protocol per step — V2 drives the whole
+/// flow through a single unified delegate. The SDK reports each server-requested step
+/// (`onCodeRequired`, `onNewPasswordRequired`) and the two terminal callbacks
+/// (`onFlowCompleted`, `onFlowError`), and the app continues by calling methods on the
+/// `MSALNativeAuthState` it is handed.
+final class MSALNativeAuthResetPasswordV2EndToEndTests: MSALNativeAuthEndToEndBaseTestCase {
```
spetrescu84, line 37: We should have a test where the code sent is invalid and we get back the correct error and state and only after we submit again correctly. It can be standalone test or a branch on an existing one

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,318 @@
...
+                                                                         retries: codeRetryCount)
+        else {
+            return
+        }
+
+        // Now submit an invalid password...
+        let flowErrorExp = expectation(description: "reset password flow error")
+        delegate.reset(expectation: flowErrorExp)
+
+        newPasswordRequiredState.submitNewPassword("INVALID_PASSWORD", delegate: delegate)
```
kaisong1990, line 133: `INVALID_PASSWORD` only fails the Entra default policy because it hits 2 of the 4 character classes - it is 16 chars, so it passes on length. If the tenant policy is ever relaxed this test silently exercises the success path and then fails on `onFlowErrorCalled` with a misleading message. The V1 suite uses `\"1\"` for the same case; please use a value that can't be complex enough.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,318 @@
...
+        self.expectation = expectation
+    }
+
+    func reset(expectation: XCTestExpectation) {
+        self.expectation = expectation
+        onCodeRequiredCalled = false
+        onNewPasswordRequiredCalled = false
+        onFlowCompletedCalled = false
+        onFlowErrorCalled = false
+        error = nil
```
kaisong1990, line 277: `reset` clears the called flags and `error` but leaves `codeRequiredState`, `newPasswordRequiredState` and `result` from the previous step. Unlike V1, this spy is reused across every step of the flow, so a future test that reads one of those after a step that didn't set it will get stale data instead of nil. Please clear them here too.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,318 @@
...
+        let flowCompletedExp = expectation(description: "reset password flow completed")
+        delegate.reset(expectation: flowCompletedExp)
+
+        let uniquePassword = generateRandomPassword()
+        newPasswordRequiredState.submitNewPassword(uniquePassword, delegate: delegate)
+
+        await fulfillment(of: [flowCompletedExp])
+        XCTAssertTrue(delegate.onFlowCompletedCalled)
+        XCTAssertEqual(delegate.scenario, .passwordReset)
+        XCTAssertNotNil(delegate.result)
```
kaisong1990, line 91: `XCTAssertNotNil` on the result is weak for the one V2 behaviour that differs from V1 - SSPR now ends with tokens. Please also assert `delegate.result?.account.username` matches `username` so the test would catch tokens issued for the wrong account.

antrix1989:
@copilot resolve the merge conflicts in this pull request

antrix1989:
@copilot Please update this existing PR directly. Do not open a child PR.

Rewrite the existing head branch `sedemche/native-auth-v2-sspr-e2e-v2release` so it is based directly on the current `dev` branch, then change this PR's base branch to `dev`.

The final `dev...HEAD` diff must contain only:

1. `MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift`
2. The minimal entries in `MSAL/MSAL.xcodeproj/project.pbxproj` required to add that test file to the existing `reset_password` group and both Native Auth E2E source targets:
   - one `PBXFileReference`
   - two `PBXBuildFile` entries
   - one group child entry
   - one Sources entry for each E2E target

Everything else must come unchanged from current `dev`. Do not retain ancestry-only changes from `sedemche/native-auth-v2_release`, PR #3039, unrelated files, submodule changes, or other project-file differences.

Implementation requirements:

- Work on this PR's existing branch and commit directly to it.
- Capture the current remote head before rewriting and force-push with `--force-with-lease`, never an unconditional force push.
- Verify `git diff --name-only dev...HEAD` shows exactly the test file and `MSAL/MSAL.xcodeproj/project.pbxproj`.
- Verify the project-file diff contains only the six test-wiring entries described above.
- Follow `.github/copilot-instructions.md` and `.clinerules/AGENTS.md`.
- Validate through the repository-supported `build.py` workflow with `MSAL.xcworkspace`; do not invoke raw `xcodebuild`.
- After pushing, confirm this PR targets `dev` and GitHub shows only the intended two-file diff.

If repository protection prevents the force-push or base-branch update, do not create another PR. Report the exact blocker and leave the prepared commit available on this PR branch if possible.

antrix1989:
@copilot please follow up on the unresolved review feedback below on this existing PR. Do not open a child PR and do not resolve reviewer threads.

The current head is `36747e39`. Verify each item against the current code before editing. The feedback appears to already be implemented, so avoid code churn unless you find a real gap.

## Suggestions to verify and address

1. [Invalid code followed by valid code](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/3052#discussion_r3646046323)
   - Current evidence: `test_resetPasswordV2_invalidCode_thenValidCode_succeeds` submits `"1"`, asserts `isInvalidCode` and `.passwordReset`, then submits the retrieved valid code and asserts `onNewPasswordRequired`.
   - Action: verify the test fully satisfies the review request. If it does, make no code change and reply on the thread with the test name and behavior. If not, fix it directly on this PR branch and reply with the commit.

2. [Use a deterministically invalid password](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/3052#discussion_r3691922856)
   - Current evidence: `test_resetPasswordV2_passwordComplexity_error` now calls `submitNewPassword("1", delegate:)`, matching the V1 pattern requested by the reviewer.
   - Action: verify the current value satisfies the request. This thread is outdated but unresolved; make no code change if already correct and reply with the current line.

3. [Clear stale state in the reused delegate spy](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/3052#discussion_r3691922864)
   - Current evidence: `ResetPasswordV2DelegateSpy.reset(expectation:)` clears `codeRequiredState`, `newPasswordRequiredState`, `result`, `error`, `scenario`, `sentTo`, `channelTargetType`, and `codeLength`, in addition to all callback flags.
   - Action: verify all reusable state is cleared. If satisfied, make no code change and reply on the thread with the relevant method. If anything is missing, fix it directly on this PR branch and reply with the commit.

## Constraints

- Work only on `sedemche/native-auth-v2-sspr-e2e-v2release`.
- Keep the PR diff limited to the existing test file and minimal Xcode project wiring.
- Leave already-resolved threads unchanged.
- Do not resolve threads on the reviewers' behalf.
- If code changes are needed, follow `.github/copilot-instructions.md` and `.clinerules/AGENTS.md`, and validate with the repository-supported `build.py` workflow rather than raw `xcodebuild`.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed in `ResetPasswordV2DelegateSpy.reset(expectation:)`. It now clears `codeRequiredState`, `newPasswordRequiredState`, `result`, `error`, `scenario`, `sentTo`, `channelTargetType`, and `codeLength`, along with all callback flags, so no state can leak between flow steps.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed in `test_resetPasswordV2_invalidCode_thenValidCode_succeeds`. The test first submits `"1"`, verifies `isInvalidCode` and the `.passwordReset` scenario, then retrieves and submits the valid OTP and verifies the flow reaches `onNewPasswordRequired` with a non-nil next state.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed. `test_resetPasswordV2_passwordComplexity_error` now submits `"1"`, matching the V1 test pattern and ensuring the value deterministically fails password complexity requirements. This thread points to the previous version of the line.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: it is there, see `test_resetPasswordV2_invalidCode_thenValidCode_succeeds`

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,375 @@
...
+            return await retrieveAndSubmitCode(delegate: delegate,
+                                               codeRequiredState: codeRequiredState,
+                                               username: username,
+                                               retries: retries - 1)
+        }
+
+        guard delegate.onNewPasswordRequiredCalled, let newPasswordRequiredState = delegate.newPasswordRequiredState else {
+            XCTFail("onNewPasswordRequired not called")
+            return nil
+        }
```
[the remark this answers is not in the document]
antrix1989: Addressed locally. `retrieveAndSubmitCode` now handles the exhausted invalid-code retry case explicitly: when `isInvalidCode` remains true and `retries == 0`, it fails with `OTP remained invalid after all retry attempts were exhausted` and returns before the generic `onNewPasswordRequired` guard. This change is pending the next branch update.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,375 @@
...
+        onFlowErrorCalled = false
+        codeRequiredState = nil
+        newPasswordRequiredState = nil
+        result = nil
+        error = nil
+        scenario = nil
+        sentTo = nil
+        channelTargetType = nil
+        codeLength = nil
+    }
```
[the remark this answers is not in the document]
antrix1989: Addressed locally. `ResetPasswordV2DelegateSpy` is now `@MainActor`, and each test plus the two private helpers that initialize, reset, and read the spy are also main-actor isolated. This keeps callback mutations and test-side state access on the same actor. This change is pending the next branch update.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,453 @@
...
+    // email delivery / an older code being read before the fresh one arrives).
+    private let codeRetryCount = 3
+
+    // SSPR – happy path: start → code required → new password required → sign in → flow completed (tokens).
+    @MainActor
+    func test_resetPasswordV2_succeeds() async throws {
+//        throw XCTSkip("SSPRV2 requires a test slice. Disable this test until api is in prod.")
+        
+        guard let sut = initialisePublicClientApplication(customAuthorityURLFormat: .tenantSubdomainTenantId),
+              let username = retrieveUsernameForResetPassword()
```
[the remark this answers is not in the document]
antrix1989: fixed

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,453 @@
...
+    // code, and returns the resulting new-password-required state.
+    @MainActor
+    private func retrieveAndSubmitCode(delegate: ResetPasswordV2DelegateSpy,
+                                       codeRequiredState: MSALNativeAuthCodeRequiredState,
+                                       username: String,
+                                       retries: Int) async throws -> MSALNativeAuthNewPasswordRequiredState? {
+        let newPasswordRequiredExp = expectation(description: "new password required")
+        delegate.reset(expectation: newPasswordRequiredExp)
+
+        guard let code = await retrieveCodeFor(email: username) else {
```
kaisong1990, line 290: This retry can never succeed. `attemptReadOtpCode` sets `lastCheckedTime = messageTime` once it returns a code, so the next `retrieveCodeFor` only matches a strictly newer message - and submitting a bad code doesn't make the server send one. The retry just polls through the progressive delays and then fails on `XCTFail("OTP code not retrieved from email")` instead of the intended "OTP remained invalid" message. Either `markEmailCheckpoint()` + `resendCode` before retrying, or drop `codeRetryCount` and let the first invalid code fail the test.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,453 @@
...
+/// on the `MSALNativeAuthState` it is handed.
+final class MSALNativeAuthResetPasswordV2EndToEndTests: MSALNativeAuthEndToEndBaseTestCase {
+    // Number of times a submitted OTP is retried when the server reports it as invalid (races with
+    // email delivery / an older code being read before the fresh one arrives).
+    private let codeRetryCount = 3
+
+    // SSPR – happy path: start → code required → new password required → sign in → flow completed (tokens).
+    @MainActor
+    func test_resetPasswordV2_succeeds() async throws {
+        throw XCTSkip("SSPRV2 requires a test slice. Disable this test until api is in prod.")
```
kaisong1990, line 42: Same skip repeated in all five tests. Put it in `setUpWithError()` so there's one place to delete when the slice goes to prod - right now every body below it is unreachable and it's easy to enable one test and miss the rest.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,453 @@
...
+        try skipIfEmailOTPThrottled(delegate.error)
+        XCTAssertTrue(delegate.onCodeRequiredCalled, "onCodeRequired should be called again after resend")
+
+        // Now get code2...
+        guard let code2 = await retrieveCodeFor(email: username) else {
+            XCTFail("OTP code could not be retrieved")
+            return
+        }
+
+        XCTAssertNotEqual(code1, code2, "Resent code should be different from the original code")
```
kaisong1990: This only proves two different strings arrived in the mailbox, not that the resent code is usable. Please submit `code2` and assert `onNewPasswordRequired` - otherwise a resend that invalidates the state still passes.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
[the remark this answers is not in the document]
antrix1989: fixed

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,453 @@
...
+    private(set) var result: MSALNativeAuthUserAccountResult?
+    private(set) var error: MSALNativeAuthFlowError?
+    private(set) var scenario: MSALNativeAuthFlowScenario?
+    private(set) var sentTo: String?
+    private(set) var channelTargetType: MSALNativeAuthChannelType?
+    private(set) var codeLength: Int?
+
+    init(expectation: XCTestExpectation) {
+        self.expectation = expectation
+    }
```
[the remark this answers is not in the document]
antrix1989: fixed

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: we will need to remove "setUpWithError" as well once we enable these test. Will keep as it.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: we don't need to prove that resent code is usable. Will keep as it is.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. I removed `codeRetryCount`, the `retries` parameter, and the recursive polling path. `retrieveAndSubmitCode` now submits the retrieved OTP once and fails immediately with a specific diagnostic if the server rejects it as invalid. This will be visible on the PR with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+/// flow through a single unified delegate. The SDK reports each server-requested step
+/// (`onCodeRequired`, `onNewPasswordRequired`, `onSignInAfterResetPasswordRequired`) and the two
+/// terminal callbacks (`onFlowCompleted`, `onFlowError`), and the app continues by calling methods
+/// on the `MSALNativeAuthState` it is handed.
+final class MSALNativeAuthResetPasswordV2EndToEndTests: MSALNativeAuthEndToEndBaseTestCase {
+
+    // SSPR – happy path: start → code required → new password required → sign in → flow completed (tokens).
+    @MainActor
+    func test_resetPasswordV2_succeeds() async throws {
+        throw XCTSkip("SSPRV2 requires a test slice. Disable this test until api is in prod.")
```
josephpab, line 42: Every test in this file opens with an unconditional `throw XCTSkip(...)`, so the whole body below it is unreachable. Swift emits `code after 'throw' will never be executed` for all five tests, which means this PR adds five permanent warnings to both E2E targets — and none of this code has ever executed against a real server.

I'd prefer gating on config so the suite self-enables when the slice ships, rather than needing a follow-up PR to delete five `throw` lines:

```swift
try XCTSkipUnless(isSSPRV2Enabled, "SSPRV2 requires a test slice. Enable once the API is in prod.")
```

backed by a key in `conf.json`, read the same way as the other native-auth config lookups in `MSALNativeAuthEndToEndBaseTestCase`. That keeps CI green, removes the dead-code warnings, and makes the suite runnable by flipping config instead of editing source.

If we keep the unconditional skip, please file a tracking issue and reference it in the skip message — right now nothing ties this back to when the slice is expected.

(Same applies to lines 87, 148, 196 and 250. Those also leave trailing-whitespace-only lines at 43, 88, 149 and 251.)

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+        try skipIfEmailOTPThrottled(delegate.error)
+
+        guard delegate.onCodeRequiredCalled, let codeRequiredState = delegate.codeRequiredState else {
+            XCTFail("onCodeRequired not called")
+            return
+        }
+
+        XCTAssertEqual(delegate.channelTargetType?.isEmailType, true)
+        XCTAssertFalse(delegate.sentTo?.isEmpty ?? true)
+        XCTAssertNotNil(delegate.codeLength)
```
josephpab: This assertion can't fail. `MSALNativeAuthCodeRequiredState.codeLength` is a non-optional `Int`, and the spy only assigns `codeLength` inside `onCodeRequired`, which the guard above already proved fired. So it's always non-nil here.

Assert something meaningful instead:

```swift
XCTAssertEqual(delegate.codeLength, 8)
```

or at minimum `XCTAssertGreaterThan(delegate.codeLength ?? 0, 0)`.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+        // Now submit an invalid password...
+        let flowErrorExp = expectation(description: "reset password flow error")
+        delegate.reset(expectation: flowErrorExp)
+
+        newPasswordRequiredState.submitNewPassword("1", delegate: delegate)
+
+        await fulfillment(of: [flowErrorExp])
+        try skipIfEmailOTPThrottled(delegate.error)
+        XCTAssertTrue(delegate.onFlowErrorCalled)
+        XCTAssertEqual(delegate.error?.isInvalidPassword, true)
```
josephpab, line 190: This test asserts the error classification but not the scenario, whereas `test_resetPasswordV2_invalidCode_thenValidCode_succeeds` does assert `XCTAssertEqual(delegate.scenario, .passwordReset)` on its error path (line 120).

Since scenario tagging on the unified delegate is one of the main behavioural differences in V2, it's worth asserting consistently on every terminal callback. Same gap at line 268 in `test_resetPasswordV2_emailNotFound_error`.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+        // Resend code
+        let resendCodeRequiredExp = expectation(description: "code required again")
+        delegate.reset(expectation: resendCodeRequiredExp)
+
+        markEmailCheckpoint()
+        codeRequiredState.resendCode(delegate: delegate)
+
+        await fulfillment(of: [resendCodeRequiredExp])
+        try skipIfEmailOTPThrottled(delegate.error)
+        XCTAssertTrue(delegate.onCodeRequiredCalled, "onCodeRequired should be called again after resend")
```
josephpab: After a successful resend the dispatcher hands back a **new** `MSALNativeAuthCodeRequiredState` via `onCodeRequired`, but this test only checks the boolean flag and then drops the state on the floor — `delegate.reset(...)` on line 229 already nil'd `codeRequiredState`, and the local variable still points at the pre-resend state.

So the test proves an email was re-sent, but not that the resent state is usable. Worth capturing and exercising it:

```swift
guard let resentState = delegate.codeRequiredState else {
    XCTFail("resend did not produce a new code-required state")
    return
}
```

and then submitting `code2` against `resentState` to confirm the flow can actually continue. Otherwise a regression where resend returns an unusable continuation would still pass this test.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+
+        // Now get code2...
+        guard let code2 = await retrieveCodeFor(email: username) else {
+            XCTFail("OTP code could not be retrieved")
+            return
+        }
+
+        XCTAssertNotEqual(code1, code2, "Resent code should be different from the original code")
```
josephpab: This is a flake risk. Entra commonly re-issues the **same** OTP on resend while the original code is still inside its validity window — resend guarantees a new email, not a new secret. If the service behaves that way this test fails for a reason that isn't a bug.

The durable assertions are (a) a second `onCodeRequired` fired, and (b) the code retrieved after the resend still verifies. I'd drop the inequality check, or downgrade it to a non-fatal log.

Related: the V1 equivalent (`test_resetPassword_resendCode_succeeds`) is skipped outright because repeated resends trip `AADSTS701014` throttling. Good that `skipIfEmailOTPThrottled` is wired in here — but expect this test to skip more often than it runs once the slice is live.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+        let flowErrorExp = expectation(description: "reset password user not found")
+        let delegate = ResetPasswordV2DelegateSpy(expectation: flowErrorExp)
+
+        let unknownUsername = UUID().uuidString + "@contoso.com"
+
+        let parameters = MSALNativeAuthResetPasswordParameters(username: unknownUsername)
+        sut.resetPasswordV2(parameters: parameters, delegate: delegate)
+
+        await fulfillment(of: [flowErrorExp])
+        try skipIfEmailOTPThrottled(delegate.error)
```
josephpab: This throttle check can never fire. No OTP is requested for an unknown username — the flow fails at start with `userNotFound` — so `skipIfEmailOTPThrottled` has nothing to classify. Safe to drop for clarity.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+            XCTFail("OTP code not retrieved from email")
+            return nil
+        }
+
+        codeRequiredState.submitCode(code, delegate: delegate)
+
+        await fulfillment(of: [newPasswordRequiredExp])
+        try skipIfEmailOTPThrottled(delegate.error)
+
+        if delegate.onFlowErrorCalled, delegate.error?.isInvalidCode == true {
```
josephpab: This narrows the diagnostic to `isInvalidCode` only. If `onFlowError` fires with anything else (`notImplemented` because the spy is missing a conformance, `invalidChallenge`, an unexpected server error), you fall through to the guard below and get the generic `"onNewPasswordRequired not called"` message, which hides the real cause — and this suite is skip-gated, so whoever first enables it will be debugging blind.

Suggest reporting the actual error instead:

```swift
if delegate.onFlowErrorCalled {
    XCTFail("submitCode failed: \(String(describing: delegate.error))")
    return nil
}
```

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+    @MainActor
+    private func submitNewPasswordAndSignIn(
+        delegate: ResetPasswordV2DelegateSpy,
+        newPasswordRequiredState: MSALNativeAuthNewPasswordRequiredState,
+        username: String
+    ) async throws {
+        let signInAfterResetPasswordRequiredExp = expectation(description: "sign in after reset password required")
+        delegate.reset(expectation: signInAfterResetPasswordRequiredExp)
+
+        let uniquePassword = generateRandomPassword()
```
josephpab, line 324: This permanently rotates the password on the shared test account, and I think that's a real problem once these tests are enabled.

`retrieveUsernameForResetPassword()` returns a single fixed account from `conf.json` (`reset_password_username` / `reset_password_username_macos`). Two tests in this new class reach this helper (`test_resetPasswordV2_succeeds` and `test_resetPasswordV2_invalidCode_thenValidCode_succeeds`), and the existing V1 class `MSALNativeAuthResetPasswordEndToEndTests` uses the *same* account in five more tests.

The E2E schemes set `parallelizable = "YES"`, and XCTest parallelization distributes **test classes** across separate processes. So the V1 and V2 reset-password classes can run concurrently against one account and one mail.tm mailbox. Two concrete failure modes:

1. **OTP cross-talk.** The per-test-case `codeRetriever` checkpoint isolates the client object, not the mailbox. Two classes polling the same inbox after their own checkpoints can each pick up the other's code, producing spurious `invalidCode` failures.
2. **Password races.** One class rotates the password while the other is mid-flow.

V1 alone was safe because its tests share a process and run serially. Adding a second class that targets the same account is what introduces the cross-process contention.

Cheapest fix is a dedicated `reset_password_v2_username` (plus macOS variant) in `conf.json` so the V2 suite gets its own account and mailbox. Alternatively mark the V2 class non-parallelizable. Worth resolving before the skips are lifted, otherwise this lands as intermittent CI noise that's painful to diagnose.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+
+        let signInParameters = MSALNativeAuthSignInAfterResetPasswordParameters()
+        signInAfterResetPasswordState.signIn(parameters: signInParameters, delegate: delegate)
+
+        await fulfillment(of: [flowCompletedExp])
+        try skipIfEmailOTPThrottled(delegate.error)
+        XCTAssertTrue(delegate.onFlowCompletedCalled)
+        XCTAssertEqual(delegate.scenario, .passwordReset)
+        XCTAssertNotNil(delegate.result)
+        XCTAssertEqual(delegate.result?.account.username, username)
```
josephpab: Exact-match on username is fragile here. Entra frequently returns a normalized UPN (casing in particular) that differs from the string we sent, so this can fail even on a fully successful flow.

Safer:

```swift
XCTAssertEqual(delegate.result?.account.username?.lowercased(), username.lowercased())
```

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+    }
+}
+
+/// Unified spy delegate for the V2 reset-password flow. A single instance receives every callback
+/// across the flow; `reset(expectation:)` swaps in a fresh expectation and clears the per-step flags
+/// before each continuation call.
+@MainActor
+class ResetPasswordV2DelegateSpy: MSALNativeAuthCodeRequiredDelegate,
```
josephpab: A few things on the spy:

1. **Not `final`, not `private`, and file-scoped.** It's visible across the whole E2E test target, so a future `SignUpV2`/`SignInV2` suite that wants its own unified spy has to work around this name. The existing V2 spies in `MSALNativeAuthFlowResponseDispatcherTests.swift` are declared `private` and `final`.
2. **Doesn't inherit `NSObject`.** It compiles (Swift synthesizes the `@objc` witnesses), but every other conformance to these `@objc` protocols in the codebase goes through a `BaseDelegateSpy: NSObject` base. Worth matching for consistency, and it insulates us if the dispatcher ever switches to `responds(to:)`-style checks rather than `as?` casts.
3. **Placement.** V1's spies live in the shared `ResetPasswordDelegateSpies.swift`. If this spy is intended to be reusable across V2 suites, it belongs in a shared file; if not, make it `private` so it clearly isn't.

Minimum ask: `private final class`.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
```diff
@@ -0,0 +1,437 @@
...
+        signInAfterResetPasswordState = nil
+        result = nil
+        error = nil
+        scenario = nil
+        sentTo = nil
+        channelTargetType = nil
+        codeLength = nil
+    }
+
+    @MainActor
```
josephpab: `@MainActor` is redundant on each method — the class is already annotated `@MainActor` at line 343, so all members inherit it. Same at 399, 408, 420 and 429. Removing them reduces noise and makes it clearer that isolation is class-wide.

MSAL/MSAL.xcodeproj/project.pbxproj
```diff
@@ -2942,6 +2944,7 @@
 		DEFB46EC2A52BA3700DBC006 /* MSALNativeAuthSignOutEndToEndTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MSALNativeAuthSignOutEndToEndTests.swift; sourceTree = "<group>"; };
 		DEFB46F12A52C11400DBC006 /* MSALNativeAuthResetPasswordEndToEndTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MSALNativeAuthResetPasswordEndToEndTests.swift; sourceTree = "<group>"; };
 		DEFB46F32A52C28100DBC006 /* ResetPasswordDelegateSpies.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ResetPasswordDelegateSpies.swift; sourceTree = "<group>"; };
+		DA3D00FD456DC3F9BF81C82E /* MSALNativeAuthResetPasswordV2EndToEndTests.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = MSALNativeAuthResetPasswordV2EndToEndTests.swift; sourceTree = "<group>"; };
```
josephpab, line 2947: `includeInIndex = 1` here is inconsistent with every neighbouring `PBXFileReference` in this section (the V1 test file and `ResetPasswordDelegateSpies.swift` both omit it). It's harmless, but it's a tell that the pbxproj was hand-edited rather than written by Xcode.

Worth dropping it so the entry matches its siblings and future Xcode-driven edits don't produce churn on this line.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: this is expected

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: we don't need to assert scenario in this tests

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. `retrieveAndSubmitCode` now treats every `onFlowError` as terminal and reports the SDK error description plus correlation ID before returning, while retaining the earlier throttle skip. This will be visible with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: changed to this: XCTAssertGreaterThan(delegate.codeLength, 0)

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. I removed the assumption that a resend must produce a different secret. The durable check now submits the OTP from the resent email through the new continuation state and verifies it advances successfully. Throttle skipping remains in place. This will be visible with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. I removed `skipIfEmailOTPThrottled` from the unknown-user path because that flow terminates before requesting an OTP. This will be visible with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. The final username assertion now lowercases both the returned account username and the configured username before comparing them. This will be visible with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. `ResetPasswordV2DelegateSpy` is now a file-private `final` class inheriting from `NSObject`, with `super.init()` called after initializing its stored expectation. This will be visible with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. I kept `@MainActor` on the spy class and removed the redundant annotation from each callback method. This will be visible with the next push.

MSAL/MSAL.xcodeproj/project.pbxproj
antrix1989: I left this harmless attribute unchanged to avoid unrelated project-file churn. Removing the single attribute while Xcode had the project loaded caused it to rewrite and reorder roughly 120 unrelated project entries; that generated rewrite was reverted. The functional test wiring remains limited to the existing file reference, group entry, and two source-target entries.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally. The resend test now captures the new `codeRequiredState`, retrieves the resent OTP, submits it through that returned state, and verifies the flow reaches `onNewPasswordRequired`. This will be visible with the next push.

MSAL/test/integration/native_auth/end_to_end/reset_password/MSALNativeAuthResetPasswordV2EndToEndTests.swift
antrix1989: Addressed locally by disabling parallel execution for the Native Auth E2E test targets in the shared iOS and macOS E2E schemes, plus the E2E target reference in `unit-test-host`. The separate Native Auth Integration Tests target remains parallelizable. This serializes the V1/V2 reset-password classes that share the account and mailbox. This will be visible with the next push.

josephpab (approved):

PR #3057, opened 2026-07-23
Implement V2 (HAL-based) SSPR flow for native auth

spetrescu84:
## Overview

This PR implements the **V2 (server-driven / HAL-based) self-service password reset (SSPR)** flow for native auth on iOS and macOS.

Unlike the V1 client-driven flow, V2 is driven by the server's HAL responses: the client follows `_links` relations returned in each response to decide the next network call, rather than hard-coding the sequence of endpoints.

## What's included

**Flow orchestration**
- `MSALNativeAuthFlowController` — orchestrates the SSPR flow (start → challenge → verify code → update password → completion) by mapping validated responses onto the next required state.
- `MSALNativeAuthFlowResponseDispatcher` — routes validated responses to the appropriate public state / delegate.
- `MSALNativeAuthFlowContinuationState`, `MSALNativeAuthFlowInternalState`, `MSALNativeAuthFlowControllerResponse`, `MSALNativeAuthFlowResult` — internal state plumbing carrying continuation tokens and resolved HAL links between layers.
- `MSALNativeAuthFlowControlling` — controller protocol seam.

**HAL networking**
- `HALResource`, `MSALNativeAuthHALResponse`, `MSALNativeAuthV2HALAction`, `MSALNativeAuthV2HALResponseSerializer` — HAL response model and serialization.
- `MSALNativeAuthV2ResponseValidator` / `MSALNativeAuthV2ValidatedResponses` — validate HAL responses into typed, actionable outcomes (code required, update required, poll in progress, ready to complete, error).
- `MSALNativeAuthV2ResponseErrorHandler`, `MSALNativeAuthV2HrefURLResolver`, `MSALNativeAuthV2LinkRelation`, `MSALNativeAuthV2Endpoint` — error handling and `_links` resolution.

**Request building (parameters-driven, mirroring V1)**
- `MSALNativeAuthV2RequestProvider` / `MSALNativeAuthV2RequestConfigurator` — build requests from typed parameter objects.
- Typed parameter classes: `MSALNativeAuthV2EntryParameters`, `MSALNativeAuthV2HrefParameters`, `MSALNativeAuthV2AuthorizeChallengeStart/ContinueParameters`, `MSALNativeAuthV2TokenParameters`, `MSALNativeAuthV2RequestTarget`, `MSALNativeAuthV2Requestable`.
- `MSALNativeAuthV2RequestBody` / `MSALNativeAuthV2RequestBodyKey` — typed request body.

**Token handling & telemetry**
- `MSALNativeAuthTokenCacher`, `MSALNativeAuthTokenRequestHandling` — extracted token request/caching seam reused by the V2 completion step.
- Telemetry api ids and operation types for the SSPR V2 operations.

**Tests**
- Unit tests for the flow controller, dispatcher, response validator, request provider, parameters, HAL serializer, href resolver, error handler, and flow error mapping.

## Notes
- Public API surface for native auth is unchanged in shape; V2 plugs in behind the existing entry points.
- Built for iOS and macOS.


MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowContinuationState.swift
```diff
@@ -0,0 +1,75 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+/// Typed key for a continuation state's link map: either a flow `_links` relation or a
+/// per-auth-method action link (keyed by method id).
+enum MSALNativeAuthV2LinkKey: Hashable {
```
spetrescu84, line 29: I've kept this in even though it will only be used on signup/signin to clarify there's 2 types of links

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowResult.swift
```diff
@@ -0,0 +1,54 @@
...
+import Foundation
+
+/// Identifies which V2 flow a ``MSALNativeAuthFlowInternalState`` belongs to.
+extension MSALNativeAuthFlowScenario {
+
+    /// The server-driven flows the SDK follows when resolving `authorize-challenge` links.
+    static let authorizeChallengeFlows: [MSALNativeAuthFlowScenario] = [.signUp, .signIn, .passwordReset, .unknown]
+
+    /// The `authorize-challenge` link relation this flow follows.
+    var link: String {
```
spetrescu84, line 34: I've kept these on to clarify why this exists. This is to connect the flow to the links that should be present in the authorize-challenge

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -35,6 +35,16 @@ class MSALNativeAuthBaseController {
         self.clientId = clientId
     }
 
+    func joinScopes(_ scopes: [String]?) -> [String] {
```
spetrescu84, line 38: I've extracted this here as now the scopes are passed when ResetPassword is called on the `publicClientApplication`

MSAL/src/native_auth/controllers/MSALNativeAuthBaseController.swift
```diff
@@ -153,35 +163,42 @@ class MSALNativeAuthBaseController {
         _ request: MSIDHttpRequest,
         context: MSALNativeAuthRequestContext
     ) async -> Result<T, Error> {
-        return await withCheckedContinuation { continuation in
```
spetrescu84, line 156: This is due to a new type of warning that says you can't capture specific things in a `async` closure

MSAL/src/native_auth/controllers/MSALNativeAuthTokenCacher.swift
```diff
@@ -0,0 +1,102 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@_implementationOnly import MSAL_Private
+
+/// Token-response cache persistence.
+final class MSALNativeAuthTokenCacher {
```
spetrescu84: I've extracted this so both v1 and v2 can use it.
The `/token` call is now different for v2 because it uses a `code`

MSAL/src/native_auth/controllers/MSALNativeAuthTokenController.swift
```diff
@@ -61,16 +63,6 @@ class MSALNativeAuthTokenController: MSALNativeAuthBaseController {
             )
         }
 
-    func joinScopes(_ scopes: [String]?) -> [String] {
```
spetrescu84, line 64: Extracted to base controller

MSAL/src/native_auth/controllers/MSALNativeAuthTokenController.swift
```diff
@@ -132,104 +124,12 @@ class MSALNativeAuthTokenController: MSALNativeAuthBaseController {
...
-                MSALNativeAuthLogger.log(level: .warning,
-                               context: context,
-                               format: "Error creating MSIDAccountIdentifier out of MSALAccount (ignoring)")
-            }
-        } catch {
-            MSALNativeAuthLogger.log(level: .warning, context: context, format: "Error clearing previous account (ignoring)")
-        }
-    }
-
-    private func performTokenRequest(
```
spetrescu84, line 209: Extracted to MSALNativeAuthTokenRequestHandling to be used by both v1 and v2

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
```diff
@@ -40,7 +40,6 @@ enum MSALNativeAuthErrorMessage {
     static let generalError = "General error"
     static let invalidCode = "Invalid code"
     static let delegateNotImplementedV2 = "Delegate %@ is not implemented"
-    static let invalidContinuationToken = "Invalid continuation token"
```
spetrescu84, line 43: This is not needed, it's not possible to recover from this, will be treated as generalError

MSAL/src/native_auth/network/errors/MSALNativeAuthErrorMessage.swift
```diff
@@ -40,7 +40,6 @@ enum MSALNativeAuthErrorMessage {
     static let generalError = "General error"
     static let invalidCode = "Invalid code"
     static let delegateNotImplementedV2 = "Delegate %@ is not implemented"
```
spetrescu84, line 42: Should we remove `delegateNotImplementedV2` and reuse `delegateNotImplemented` ?

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -0,0 +1,264 @@
...
+            MSALLogMask.maskPII(serverError.message))
+    }
+
+    private func estsErrorCodes(from message: String?) -> [Int] {
+        guard let message = message else {
+            return []
+        }
+        var codes: [Int] = []
+        let scanner = Scanner(string: message)
+        let marker = "AADSTS"
```
spetrescu84, line 246: This is not a good way to check for errors, but it's only thing we have, we need to talk to API team to see if they can add error code.

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthHALResponse.swift
```diff
@@ -0,0 +1,111 @@
...
+@_implementationOnly import MSAL_Private
+
+/// Generic, server-driven HAL response used by all Native Auth V2 flows.
+///
+/// Every V2 HTTP outcome (200 success, the `401` from `authorize-challenge`,
+/// and `4xx` error bodies) is parsed into a single ``MSALNativeAuthHALResponse``. The
+/// V2 response validator then inspects `error`, `state` and `action` to decide how the
+/// flow should proceed. HAL parsing itself is delegated to the shared
+/// `HALResource` / `HALLink` Swift types.
+struct MSALNativeAuthHALResponse: MSALNativeAuthResponseCorrelatable {
```
spetrescu84: I've kept the response that works for all flows here as it's very difficult to see what gets added after

MSAL/src/native_auth/network/responses/v2/HALResource.swift
```diff
@@ -0,0 +1,117 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+/// Represents a HAL Link Object as defined by the JSON-HAL specification.
```
spetrescu84, line 27: This works for all - sspr/signup/signin

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthV2HALAction.swift
```diff
@@ -0,0 +1,40 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+/// The `action` a Native Auth V2 (HAL) interaction response instructs the SDK to perform next.
+///
+/// The validator maps the raw `action` string carried by ``MSALNativeAuthHALResponse`` onto one of
+/// these cases to decide the next step of the flow.
+enum MSALNativeAuthV2HALAction: String {
```
spetrescu84: I've kept all actions as it's difficult to see the new ones being added

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthV2HALResponseSerializer.swift
```diff
@@ -0,0 +1,157 @@
...
+
+@_implementationOnly import MSAL_Private
+
+/// Parses a raw HTTP response into a ``MSALNativeAuthHALResponse``.
+///
+/// V2 responses are HAL+JSON and every HTTP outcome carries a meaningful body, so this
+/// serializer never throws on a non-200 status - it captures the status code and lets the
+/// V2 validator decide. HAL `_links` / `_embedded` extraction is delegated to the shared
+/// `HALResource`.
+final class MSALNativeAuthV2HALResponseSerializer: NSObject, MSIDResponseSerialization {
```
spetrescu84, line 33: I've kept the serialiser fully as it's hard to see the new things getting added

MSAL/src/native_auth/network/v2/MSALNativeAuthV2HrefURLResolver.swift
```diff
@@ -0,0 +1,139 @@
...
+import Foundation
+
+/// Builds request URLs for the Native Auth V2 flows.
+///
+/// V2 is server-driven, so most steps follow `_links` hrefs returned by the server. Those
+/// hrefs may be absolute, or relative/templated (e.g. `{tenant}/api/v0.1/auth/...`). This
+/// resolver normalises a server href against the configured authority host, and also builds
+/// URLs for the fixed ``MSALNativeAuthV2Endpoint`` cases. The slice/data-center query
+/// parameter is appended consistently.
+struct MSALNativeAuthV2HrefURLResolver {
```
spetrescu84, line 36: This class currently works with how the urls are returned in the response from dc. We can change for production but it will then stop working for dc

MSAL/src/native_auth/network/v2/MSALNativeAuthV2LinkRelation.swift
```diff
@@ -0,0 +1,43 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+/// HAL `_links` relation names the SDK follows to advance a Native Auth V2 (server-driven) flow.
+///
+/// Every href the SDK resolves is keyed by one of these relations (top-level `_links` or an
+/// embedded method's `_links`).
+enum MSALNativeAuthV2LinkRelation: String {
```
spetrescu84: I've kept all possible links relations as it's hard to see the new ones getting added

MSAL/src/native_auth/network/MSALNativeAuthUrlRequestSerializer.swift
```diff
@@ -33,10 +33,16 @@ final class MSALNativeAuthUrlRequestSerializer: NSObject, MSIDRequestSerializati
...
     private let context: MSIDRequestContext
     private let encoding: MSALNativeAuthUrlRequestEncoding
+    private let body: [AnyHashable: Any]?
 
-    init(context: MSIDRequestContext, encoding: MSALNativeAuthUrlRequestEncoding) {
+    /// When non-nil, `body` is serialized as the HTTP body instead of the `parameters` argument passed
+    /// to `serialize(with:parameters:headers:)`. This lets callers supply a nested JSON body that does
+    /// not fit `MSIDHttpRequest.parameters` (`[String: String]`). When nil, the `parameters` argument
+    /// is used, preserving the default behavior.
+    init(context: MSIDRequestContext, encoding: MSALNativeAuthUrlRequestEncoding, body: [AnyHashable: Any]? = nil) {
```
spetrescu84, line 42: This one got changed because so far request bodies have had the MSIDHTTPRequest parameters format of [String: String]. In v2 they can be more complex such as: 
```
{ "attributes" : ["key" : "value"] }
```

MSAL/src/native_auth/telemetry/MSALNativeAuthOperationTypes.swift
```diff
@@ -83,3 +83,20 @@ enum MSALNativeAuthSignOutType: MSALNativeAuthOperationType {
     case signOutAction = 0
     case signOutForced = 1
 }
+
+enum MSALNativeAuthV2OperationType: MSALNativeAuthOperationType {
```
spetrescu84, line 87: Kept all operation types for v2 to show the full picture, otherwise numbers wouldn't make sense

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -0,0 +1,611 @@
...
+        }
+
+        guard case .pollInProgress(var pollToken, let pollHref) = updateResult else {
+            return interactionFailure(updateResult, event: event, context: context, scenario: continuation.flowScenario, newState: nil)
+        }
+
+        var completionToken: String?
+        for attempt in 0..<kNumberOfTimesToRetryPollCompletionCall {
+            if attempt > 0 {
+                try? await Task.sleep(nanoseconds: pollIntervalNanoseconds)
```
antrix1989: do we run this on the bg thread?

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -0,0 +1,611 @@
...
+                context: context
+            )
+        }
+
+        guard case .pollInProgress(var pollToken, let pollHref) = updateResult else {
+            return interactionFailure(updateResult, event: event, context: context, scenario: continuation.flowScenario, newState: nil)
+        }
+
+        var completionToken: String?
+        for attempt in 0..<kNumberOfTimesToRetryPollCompletionCall {
```
antrix1989: lets extract this logic (or reuse current [RetryExecutor](https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/cbad8087af87b1a5ed758cab1c923c096325acee/MSAL/test/integration/native_auth/end_to_end/otp_code_retriever/RetryExecutor.swift#L31) class)

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -0,0 +1,611 @@
...
+        } catch {
+            return .failure(error)
+        }
+    }
+
+    // MARK: - Result mapping
+
+    // Maps a validated interaction response onto a controller response, building the next
+    // required state or, on a terminal response, running the authorize-challenge → token completion.
+    private func mapInteraction(
```
antrix1989: Lets extract this logic into separate class and cover with tests. Maybe this can be placed in MSALNativeAuthFlowControllerResponse or in a category of it

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowResult.swift
```diff
@@ -0,0 +1,54 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+/// Identifies which V2 flow a ``MSALNativeAuthFlowInternalState`` belongs to.
+extension MSALNativeAuthFlowScenario {
+
+    /// The server-driven flows the SDK follows when resolving `authorize-challenge` links.
+    static let authorizeChallengeFlows: [MSALNativeAuthFlowScenario] = [.signUp, .signIn, .passwordReset, .unknown]
```
antrix1989: if we add a new flow type, does it mean we will need to update manually here? This brakes [Open-Closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle)

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowResult.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+}
+
+/// Result produced by the unified V2 controller for a single step of a flow.
+enum MSALNativeAuthFlowResult {
+    case actionRequired(state: MSALNativeAuthState)
+    case completed(MSALNativeAuthUserAccountResult)
+    case error(error: MSALNativeAuthFlowError, newState: MSALNativeAuthFlowInternalState?)
+    case browserRequired(url: URL, newState: MSALNativeAuthFlowInternalState)
```
antrix1989: did we confirm that server provided us url for browser required? How client will handle such url?

MSAL/src/native_auth/controllers/MSALNativeAuthTokenCacher.swift
antrix1989: How this class is related to MSALNativeAuthCacheAccessor? Does it internally use it via cacheAccessor

MSAL/src/native_auth/network/responses/v2/validator/MSALNativeAuthV2ResponseValidator.swift
```diff
@@ -0,0 +1,264 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@_implementationOnly import MSAL_Private
+
+/// Maps a raw ``MSALNativeAuthHALResponse`` (or transport error) into a validated, controller-facing response.
+protocol MSALNativeAuthV2ResponseValidating {
```
antrix1989: I understand that we follow v1 api where we use these "validators", but I don't think this name reflects what we do here. In my opinion "validator" should only return BOOL -- valid/non valid, but if we return the response object/model, I would rather call it "parser".

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthHALResponse.swift
antrix1989: How will this handle cred management responses? Does it mean we will need to add more properties that are cred management specific to this class? I think we should split this into subclasses for each response type

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthV2HALAction.swift
antrix1989: I think we should separate them, otherwise we brake Open-Closed principle.

MSAL/src/native_auth/network/v2/MSALNativeAuthV2Endpoint.swift
```diff
@@ -0,0 +1,36 @@
...
+
+/// Fixed (non server-driven) endpoints used by the Native Auth V2 flows.
+///
+/// V2 is server-driven: most steps follow `_links` hrefs returned by the server. Only
+/// the authorization challenge / token exchange have fixed paths the client must know up-front.
+enum MSALNativeAuthV2Endpoint: String, CaseIterable {
+    /// Authorization challenge
+    case authorizeChallenge = "/oauth2/v2.0/authorize-challenge"
+    /// Token exchange
+    case token = "/oauth2/v2.0/token"
```
antrix1989: it is same as v1  (MSALNativeAuthEndpoint). Can we reuse it from there?

MSAL/src/native_auth/network/v2/MSALNativeAuthV2HrefURLResolver.swift
antrix1989: I assume if use Charles and append dc there, we don't need this, rigth?

MSAL/src/native_auth/network/v2/MSALNativeAuthV2LinkRelation.swift
antrix1989: I suggest to separate

MSAL/src/native_auth/network/v2/MSALNativeAuthV2RequestBody.swift
```diff
@@ -0,0 +1,47 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+/// Typed body of a Native Auth V2 (HAL) follow-up request. Each field maps to a
+/// ``MSALNativeAuthV2RequestBodyKey`` and is emitted only when set, so a caller states just the
+/// fields the operation carries
+struct MSALNativeAuthV2RequestBody {
```
antrix1989: it seems this class has multiple responsibilities. Can we split into sublcasess?

MSAL/src/native_auth/public/MSALNativeAuthPublicClientApplication.swift
```diff
@@ -152,6 +152,7 @@ public final class MSALNativeAuthPublicClientApplication: MSALPublicClientApplic
     /// - Parameters:
     ///   - parameters: Parameters used for the Sign Up flow.
     ///   - delegate: Delegate that receives callbacks for the Sign Up flow.
+    @available(*, deprecated, message: "Use signUpV2(parameters:delegate:) instead")
```
antrix1989: lets not deprecate v1 apis until we are GA'ed

MSAL/src/native_auth/controllers/MSALNativeAuthTokenCacher.swift
spetrescu84: Yeah, I've changed the v1 class to have a reference to this `self.tokenCacher = MSALNativeAuthTokenCacher(cacheAccessor: cacheAccessor)` and then this class below has `private let cacheAccessor: MSALNativeAuthCacheInterface` actually realised that the cacheAccessor is no longer needed in the base class so removed.

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthV2HALAction.swift
spetrescu84: Not sure what you mean, I mean I've kept the SignUp and SignIn ones also in

MSAL/src/native_auth/network/v2/MSALNativeAuthV2HrefURLResolver.swift
spetrescu84: This is because the links that come back contain the dc. But also because we need to make sliceConfig work for v2 also

MSAL/src/native_auth/network/v2/MSALNativeAuthV2LinkRelation.swift
spetrescu84: Just to clarify, here I'm talking about SingUp/SignIn ones. I can remove but then they will be added to the same class

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: This changed now, using RetryExecutor

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Done, moved RetryExecutor into MSAL Codebase and renamed to MSALNativeAuthRetryExecutor and now both e2e and main codebase use

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowResult.swift
spetrescu84: Yeah, saw the issue originally also, but thought we can't do CaseIterable because of the ObjC thing but it actually works so I've updated.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Made `mapInteraction` internal and added unit tests for it.
Didn't move as there's quite a few depdendecies here and it's used to calculate the next steps in flows also

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowResult.swift
spetrescu84: Actually good point this case was never treated, updated and now it's parsed from the server. The URL is not used. Didn't encounter during testing but was in API design doc

MSAL/src/native_auth/network/responses/v2/validator/MSALNativeAuthV2ResponseValidator.swift
spetrescu84: Done, it only parses, there's not too much to validate

MSAL/src/native_auth/network/v2/MSALNativeAuthV2Endpoint.swift
spetrescu84: Should we move other one also to the other enum?

MSAL/src/native_auth/controllers/MSALNativeAuthTokenCacher.swift
antrix1989: Why don't we extend MSALNativeAuthCacheAccessor and put this logic there? Why it needs to be a separate class?

MSAL/src/native_auth/network/v2/MSALNativeAuthV2HrefURLResolver.swift
```diff
@@ -0,0 +1,141 @@
...
+        components.path += endpoint.rawValue
+        return try applyingDataCenter(to: components)
+    }
+
+    /// Resolves a server-provided `_links` href into an absolute URL against the authority host.
+    func url(forHref href: String) throws -> URL {
+        let trimmed = href.trimmingCharacters(in: .whitespacesAndNewlines)
+
+        // Absolute href: use as-is (still append the data-center parameter).
+        if let absolute = URL(string: trimmed), let scheme = absolute.scheme, scheme.hasPrefix("http") {
```
mipetriu, line 65: Since v2 is server-driven and we POST the continuation token — and for `verify`/`updatePassword` the OTP and the new password — to whatever `_links` href comes back, what constrains the host we send those to? In the absolute-href branch we take the server URL as-is (`scheme.hasPrefix("http")`) with no check that its host matches `authorityURL`. Two things worth confirming: (1) should we validate the absolute href's host against the configured authority host before following it, so a tampered/misconfigured response can't redirect credentials to another host? and (2) `hasPrefix("http")` also accepts `http://` — should we require `https` here so we never downgrade a code/password to cleartext? I get that today DC returns hrefs on the same host and TLS protects transit, but for a credential-bearing flow the host allow-list feels like cheap defense-in-depth. (Note: the relative branch already reconstructs against the authority, so this is only the absolute path.)

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -0,0 +1,607 @@
...
+            try self.requestProvider.updatePassword(
+                href: updateHref,
+                newPassword: password,
+                continuationToken: continuation.continuationToken,
+                apiId: .telemetryApiIdV2ResetPasswordSubmit,
+                context: context
+            )
+        }
+
+        guard case .pollInProgress(var pollToken, let pollHref) = updateResult else {
```
mipetriu, line 213: This step guards only on `.pollInProgress` and sends everything else to `interactionFailure(..., newState: nil)`. But a weak-password rejection at `update` maps to `.invalidPassword` (`passwordTooWeak`), and here that would come back with `newState: nil` — so the user can't retry the password and has to restart the whole reset. `mapInteraction` deliberately keeps `invalidPassword`/`invalidCode` recoverable (and you have `test_mapInteraction_error_invalidPassword_isRecoverable` for it), but `submitNewPassword` bypasses `mapInteraction`, so that recoverability is lost on this path. Could we route the update result through the same recoverable-state handling (or at least pass `recoverableState: state` for `.invalidPassword`)? Relatedly, if `update` ever returns `.readyToComplete`/`.updateRequired` directly instead of a poll link, we'd also treat it as a generic failure — is polling always guaranteed by the API here?

MSAL/src/native_auth/network/v2/MSALNativeAuthV2LinkRelation.swift
spetrescu84: Done

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Updated this, and removed `mapInteraction`, now it's more clean

MSAL/src/native_auth/network/v2/MSALNativeAuthV2Endpoint.swift
spetrescu84: Moved both, removed fully

MSAL/src/native_auth/network/v2/MSALNativeAuthV2RequestBody.swift
spetrescu84: Updated to have one request parameters body class per operation

MSAL/src/native_auth/controllers/MSALNativeAuthTokenCacher.swift
spetrescu84: Done, moved to extension of MSALNativeAuthCacheAccessor

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthHALResponse.swift
spetrescu84: Split by response type

MSAL/src/native_auth/network/responses/v2/MSALNativeAuthV2HALAction.swift
djanardhan98: Curious to understand why we're using an enum like this here and why we're using the struct + extension in MSALNativeAuthV2LinkRelation? Is it because we expect that to change or extend in the future but not this?

spetrescu84: Done, forgot to push this

MSAL/src/native_auth/network/v2/MSALNativeAuthV2HrefURLResolver.swift
spetrescu84: Thanks for the feedback, we are trying to get the API team to return just the API endpoint, like `oauth2/v2.0/authorize-challenge` and then we would be able to use the v1 URL Resolver so this class would go away. Added a //TODO a the start of the class to change it once we clarify conversation

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Good catch re the unrecoverable state, update to first check if we have any error and fail based on type
The API contract does say that `update` must return `poll` after

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowContinuationState.swift
```diff
@@ -0,0 +1,69 @@
...
+/// Holds the opaque server `continuation_token` and the resolved `_links` hrefs the SDK must
+/// follow to advance the server-driven flow.
+class MSALNativeAuthFlowContinuationState {
+    let flowScenario: MSALNativeAuthFlowScenario
+    let continuationToken: String?
+    let links: [MSALNativeAuthV2LinkKey: URL]
+    let username: String?
+    /// Scopes (caller-requested merged with the default OIDC scopes) to request on the final
+    /// `/token` exchange. Threaded through every step.
+    let scopes: [String]
```
antrix1989: should we have a new callback (state) to request scopes just before we call /token enpoint? This will also allow us to keep same SSPR params in V2 as we had in V1

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowContinuationState.swift
```diff
@@ -0,0 +1,69 @@
...
+
+/// Internal continuation context carried by a ``MSALNativeAuthFlowInternalState``.
+///
+/// Holds the opaque server `continuation_token` and the resolved `_links` hrefs the SDK must
+/// follow to advance the server-driven flow.
+class MSALNativeAuthFlowContinuationState {
+    let flowScenario: MSALNativeAuthFlowScenario
+    let continuationToken: String?
+    let links: [MSALNativeAuthV2LinkKey: URL]
+    let username: String?
```
antrix1989: when do we need it? Also in what flows it is needed?

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
antrix1989: yeah, I asked them to provide suberrors

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowContinuationState.swift
spetrescu84: Removed, shouldn't be in this flow

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowContinuationState.swift
spetrescu84: Indeed, done. Also fixed correlation id to be consistent in the flow

antrix1989 (approved):

PR #3067, opened 2026-07-31
[patch][engg]: Adopt UIScene lifecycle in 3 iOS host app targets for iOS 27

mipetriu:
Migrate the unit-test-host, MSAL Test Automation (iOS), and MSAL Test App (iOS) targets to the UIScene lifecycle for iOS 27 SDK readiness. Each target gets its own scene delegate class and scene manifest entry:

- unit-test-host: MSALUnitTestHostSceneDelegate (storyboard-driven).
- MSAL Test Automation (iOS): MSALAutoSceneDelegate, forwards scene:openURLContexts: to the app delegate so MSAL redirect completion works.
- MSAL Test App (iOS): MSALTestAppSceneDelegate owns the programmatic UITabBarController window (moved out of didFinishLaunchingWithOptions) and forwards scene:openURLContexts:.

Source-only migration; not built.

## PR Checklist (must be completed before review)

- [ ] All tests pass locally
- [ ] PR size is <= 500 LOC per PR Size Check policy
- [ ] PR is independently mergeable (no hidden dependencies)
- [ ] Appropriate reviewers are assigned
- [ ] PR reviewed by code owner (required if Copilot-generated)
- [ ] SME or Senior IC assigned where required

## PR Title Format

**Required Format:** `[Keyword1] [Keyword2]: Description`

- **Keyword1:** `major`, `minor`, or `patch` (case-insensitive)
- **Keyword2:** `feature`, `bugfix`, `engg`, or `tests` (case-insensitive)

**Examples:**
- `[MAJOR] [Feature]: new API`
- `[minor] [bugfix]: fix crash`
- `[PATCH] [tests]: add coverage`

## Proposed changes

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/MSAL/test/app/ios/MSALTestAppDelegate.m
+++ b/MSAL/test/app/ios/MSALTestAppDelegate.m
@@ -28,16 +28,9 @@
 #import "MSALTestAppDelegate.h"
 #import "MSALTestAppSettings.h"
 
-#import "MSALTestAppAcquireTokenViewController.h"
-#import "MSALTestAppCacheViewController.h"
-#import "MSALTestAppLogViewController.h"
-#import "MSALTestAppSettingsViewController.h"
 #import "MSALPublicClientApplication.h"
 
 @implementation MSALTestAppDelegate
-{
-    UITabBarController* _tabBar;
-}
 
 /*
 - (BOOL)application:(UIApplication *)application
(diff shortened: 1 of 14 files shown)


MSAL/test/app/ios/MSALTestAppDelegate.m
```diff
@@ -52,43 +45,10 @@ - (BOOL)application:(UIApplication *)application
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+    // Window and root view controller are set up by MSALTestAppSceneDelegate in the UIScene lifecycle.
```
kaisong1990, line 48: The `window` property on the app delegate is dead now - nothing assigns it and UIKit doesn't populate it under the scene lifecycle, so it'll always be nil. Worth removing it from `MSALTestAppDelegate.h` (and the automation / unit-test-host app delegates) so nobody wires up to it later.

kaisong1990:
Scope question: `Samples/ios/SampleApp` and `Samples/ios/SampleAppiOS-Swift` are still delegate-only, no scene manifest. Same iOS 27 problem applies there and those are what customers copy from. Follow-up PR, or intentionally out of scope here?

mipetriu:
Good catch on the samples. You're right that `Samples/ios/SampleApp` (ObjC) and `SampleAppiOS-Swift` are still delegate-only with no scene manifest, so they carry the same iOS 27 crash risk.

I'm going to **handle them in a separate follow-up PR** rather than fold them into this one, for two reasons:

1. **They're customer-copy templates**, so they deserve their own focused, simulator-tested change rather than riding along on the host-app test-target migration here.
2. **The Swift sample needs real verification**: `showMainVC()`/`showLoginVC()` read `self.window?.rootViewController` and are called externally from `SampleLoginViewController.swift` and `SampleMainViewController.swift` via `UIApplication.shared.delegate as! AppDelegate`. Once the app-delegate window goes nil under scene lifecycle, those would silently no-op and login/logout would stop switching screens — so the migration has to repoint those readers to the scene's window and be validated on a simulator, not just compiled.

This PR stays scoped to the 3 host-app test targets (the actual iOS 27 launch-crash blockers for CI). I'll open the samples migration as a tracked follow-up. Thanks!

MSAL/test/app/ios/MSALTestAppDelegate.m
mipetriu: Done — removed the dead `window` property from all three app-delegate headers (`MSALTestAppDelegate.h`, `MSALAutoAppDelegate.h`, `unit-test-host/AppDelegate.h`) in 7f5faa65a56fb7843b6d5c640774dac7de0e9c38. Confirmed there were no readers of the app-delegate window (only the SceneDelegate's own `self.window` remains). Thanks!

josephpab (approved):

PR #3070, opened 2026-08-07
[patch][engg]: Migrate MSAL iOS sample apps to UIScene lifecycle

mipetriu:
## Summary

Migrates the two MSAL iOS sample apps (`SampleAppiOS` ObjC and `SampleAppiOS-Swift`) to the UIScene-based lifecycle, required for apps built with the iOS 27.0 SDK. Apps linked against the 27.0 SDK without UIScene adoption crash on launch.

## Changes

**Both apps**
- Added `UIApplicationSceneManifest` to `Info.plist` — `UIApplicationSupportsMultipleScenes = NO`, single `UIWindowSceneSessionRoleApplication` configuration named `Default Configuration`, and no `UISceneStoryboardFile` — the window is created in code.
  - Note: the Swift sample's view controllers still come from `Main.storyboard` via `AppDelegate.mainVC()` / `loginVC()`. Only the window is code-created there. The ObjC sample is fully code-built.
- Added a scene delegate that creates the window from the `UIWindowScene` and installs the root view controller.
- Moved MSAL redirect-URL handling from the app delegate's `application:openURL:options:` to `scene:openURLContexts:` on the scene delegate, and removed the now-dead app-delegate handlers.

**ObjC (`SampleAppiOS`)**
- New `SampleSceneDelegate.{h,m}`.
- `SampleAppDelegate` keeps `_rootController` / `_currentController` and `+/-setCurrentViewController:`; the scene delegate calls a new `-installRootIntoWindow:` to build the hierarchy. Existing callers reaching the app delegate via `[[UIApplication sharedApplication] delegate]` are unchanged.

**Swift (`SampleAppiOS-Swift`)**
- New `SceneDelegate.swift`.
- The scene delegate assigns the scene's window to `AppDelegate.window`, so the existing `showMainVC()` / `showLoginVC()` root-swapping continues to work unchanged.

## Verification

- Both apps: `** BUILD SUCCEEDED **` on `iPhone 17, OS=27.0` (Xcode 26.6).
- Both apps install and launch on the iOS 27.0 simulator and stay alive under the scene lifecycle.
- No changes outside `Samples/ios`. No submodule pointer changes.

### DIFF
--- a/Samples/ios/SampleApp/SampleAppiOS/Info.plist
+++ b/Samples/ios/SampleApp/SampleAppiOS/Info.plist
@@ -35,6 +35,23 @@
 	<true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
+	<key>UIApplicationSceneManifest</key>
+	<dict>
+		<key>UIApplicationSupportsMultipleScenes</key>
+		<false/>
+		<key>UISceneConfigurations</key>
+		<dict>
+			<key>UIWindowSceneSessionRoleApplication</key>
+			<array>
+				<dict>
+					<key>UISceneConfigurationName</key>
+					<string>Default Configuration</string>
+					<key>UISceneDelegateClassName</key>
+					<string>SampleSceneDelegate</string>
+				</dict>
+			</array>
+		</dict>
+	</dict>
 	<key>UIRequiredDeviceCapabilities</key>
 	<array>
 		<string>armv7</string>
--- a/Samples/ios/SampleApp/SampleAppiOS/SampleAppDelegate.m
+++ b/Samples/ios/SampleApp/SampleAppiOS/SampleAppDelegate.m
@@ -47,12 +47,17 @@
     // The MSAL Logger should be set as early as possible in the app launch sequence, before any MSAL
     // requests are made.
     [SampleMSALUtil setup];
-    
-    UIWindow* window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
+
+    // Window and root view controller are set up by SampleSceneDelegate in the UIScene lifecycle.
+    return YES;
+}
+
+- (void)installRootIntoWindow:(UIWindow *)window
+{
     self.window = window;
-    
+
     _rootController = [UIViewController new];
-    
+
     if ([[SampleMSALUtil sharedUtil] currentAccount:nil])
     {
         [self setCurrentViewController:[SampleMainViewController sharedViewController]];
@@ -61,12 +66,10 @@
     {
         [self setCurrentViewController:[SampleLoginViewController sharedViewController]];
     }
-    
+
     [window setRootViewController:_rootController];
     [window setBackgroundColor:[UIColor whiteColor]];
     [window makeKeyAndVisible];
-    
-    return YES;
 }
 
 - (BOOL)application:(UIApplication *)app
--- a/Samples/ios/SampleApp/SampleAppiOS/SampleAppDelegate.m
+++ b/Samples/ios/SampleApp/SampleAppiOS/SampleAppDelegate.m
@@ -25,8 +25,6 @@
 //
 //------------------------------------------------------------------------------
 
-#import <MSAL/MSAL.h>
-
 #import "SampleAppDelegate.h"
 #import "SampleMSALUtil.h"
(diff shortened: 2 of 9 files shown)


Samples/ios/SampleAppiOS-Swift/SampleAppiOS-Swift/SceneDelegate.swift
```diff
@@ -0,0 +1,69 @@
...
+
+        // Keep the app delegate's window reference pointing at the active scene window so the
+        // existing showMainVC()/showLoginVC() call sites continue to swap the root controller.
+        let appDelegate = UIApplication.shared.delegate as? AppDelegate
+        appDelegate?.window = window
+
+        let initialViewController: UIViewController
+        do {
+            try SampleMSALAuthentication.shared.currentAccount()
+            initialViewController = appDelegate?.mainVC() ?? UIViewController()
```
kaisong1990, line 49: If the cast on line 43 fails, `appDelegate?.window` is never set and both branches fall back to a blank `UIViewController` - the app launches to a white screen with nothing in the log. The ObjC sample avoids this by handing root installation to the app delegate. Prefer `guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }` over the optional-chain fallbacks here.

sidudiki (approved):

Samples/ios/SampleApp/SampleAppiOS/Info.plist
```diff
@@ -35,6 +35,23 @@
 	<true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
+	<key>UIApplicationSceneManifest</key>
```
jasoncoolmax, line 38: Just curious, if apps linked against the iOS 27 SDK crash on launch without UIScene adoption, should we cover the test app here too? `MSAL/test/app/ios/resources/Info.plist` has no `UIApplicationSceneManifest`, and `MSALTestAppDelegate.m` still creates the window from `[[UIScreen mainScreen] bounds]` and handles the redirect in `application:openURL:options:`. It is a build target in `build.py` (`iosTestApp`) and it is what we use for manual + automation runs, so I guess it would hit the same crash. The vision test app already has the manifest, so it looks like only the iOS one is left behind :)

Samples/ios/SampleApp/SampleAppiOS/SampleAppDelegate.m
```diff
@@ -47,12 +45,17 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
     // requests are made.
     [SampleMSALUtil setup];
-    
-    UIWindow* window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
+
+    // Window and root view controller are set up by SampleSceneDelegate in the UIScene lifecycle.
+    return YES;
+}
+
+- (void)installRootIntoWindow:(UIWindow *)window
```
antrix1989, line 53: Can we reuse the existing root hierarchy, or detach and reset `_currentController` before creating a new root? If the scene disconnects and reconnects while the process survives, the app delegate and singleton child controllers survive too. The second installation then takes the animated swap path with a child from the old root; with unchanged account state, it can even transition from a controller to itself. Please add a reconnect regression for both unchanged and changed account state.

Samples/ios/SampleAppiOS-Swift/SampleAppiOS-Swift/SceneDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+            initialViewController = appDelegate.mainVC()
+        } catch {
+            initialViewController = appDelegate.loginVC()
+        }
+
+        window.rootViewController = initialViewController
+        window.makeKeyAndVisible()
+
+        if !connectionOptions.urlContexts.isEmpty {
+            self.scene(scene, openURLContexts: connectionOptions.urlContexts)
```
antrix1989, line 62: Can we add unit tests for the redirect forwarding in both new scene delegates? There are no test changes for this migration, so build/launch checks do not verify cold `connectionOptions` versus warm URL delivery, preservation of `sourceApplication`, or handled versus unhandled responses. Please use an injected response-handler fake and keep the delegate logic real, without swizzling or initializing live MSAL/Keychain state.

PR #3072, opened 2026-08-10
[minor][feature]: Add caller-provided keys for AT PoP

antonioalwan:
## Summary

Adds the public MSAL API for caller-provided RSA keys in Access Token Proof-of-Possession flows. MSAL validates and retains the caller-owned key pair, derives `req_cnf` from its public key, signs the SHR with its private key, and marks Broker requests for selective rollout control.

## Key changes

- Add public `MSALExternalKeyPair` with a read-only RFC 7638 `keyId`.
- Add an external-key initializer to `MSALAuthenticationSchemePop`.
- Use the same external key manager for both token binding and SHR signing.
- Add `external_key_pop=1` to scheme parameters for Broker classification.
- Add deterministic validation and cryptographic SHR signature tests.

## Dependency

- CommonCore: AzureAD/microsoft-authentication-library-common-for-objc#1934

## How to validate

- Run `MSALExternalKeyPairTests`.
- Run `MSALAuthSchemeTests`.
- Verify the external SHR signature with the supplied RSA public key.

## Notes

The caller owns key creation, storage, rotation, and deletion. MSAL does not expose or persist private key material.

### DIFF
--- a/MSAL/src/MSALAuthenticationSchemePop.m
+++ b/MSAL/src/MSALAuthenticationSchemePop.m
@@ -33,6 +33,7 @@
 #import "MSIDAccessToken.h"
 #import "MSIDDefaultTokenCacheAccessor.h"
 #import "MSIDAssymetricKeyPair.h"
+#import "MSALExternalKeyPair+Internal.h"
 
 static NSString *keyDelimiter = @" ";
 
@@ -42,6 +43,8 @@ static NSString *keyDelimiter = @" ";
 @property (nonatomic) NSURL *requestUrl;
 @property (nonatomic) NSString *nonce;
 @property (nonatomic) NSDictionary *additionalParameters;
+@property (nonatomic) MSALExternalKeyPair *externalKeyPair;
+@property (nonatomic) MSIDDevicePopManager *externalPopManager;
 
 @end
 
@@ -52,6 +55,19 @@ static NSString *keyDelimiter = @" ";
                              nonce:(NSString *)nonce
               additionalParameters:(NSDictionary *)additionalParameters
 {
+    return [self initWithHttpMethod:httpMethod
+                        requestUrl:requestUrl
+                             nonce:nonce
+              additionalParameters:additionalParameters
+                   externalKeyPair:nil];
+}
+
+- (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
+                        requestUrl:(NSURL *)requestUrl
+                             nonce:(NSString *)nonce
+              additionalParameters:(NSDictionary *)additionalParameters
+                   externalKeyPair:(nullable MSALExternalKeyPair *)externalKeyPair
+{
     self = [super init];
     if (self)
     {
@@ -60,6 +76,11 @@ static NSString *keyDelimiter = @" ";
         _requestUrl = requestUrl;
         _nonce = nonce ? nonce : [[NSUUID new] UUIDString];
         _additionalParameters = additionalParameters ? additionalParameters : [NSDictionary new];
+        _externalKeyPair = externalKeyPair;
+        if (externalKeyPair)
+        {
+            _externalPopManager = [[MSIDDevicePopManager alloc] initWithExternalKeyPair:externalKeyPair.msidKeyPair];
+        }
     }
 
     return self;
@@ -79,6 +100,7 @@ static NSString *keyDelimiter = @" ";
 
 - (NSDictionary *)getSchemeParameters:(MSIDDevicePopManager *)popManager
 {
+    popManager = self.externalPopManager ?: popManager;
     NSMutableDictionary *schemeParams = [NSMutableDictionary new];
     NSString *requestConf = popManager.keyPair.jsonWebKey;
     if (requestConf)
@@ -104,6 +126,7 @@ static NSString *keyDelimiter = @" ";
 
 - (nullable NSString *)getClientAccessToken:(MSIDAccessToken *)accessToken popManager:(nullable MSIDDevicePopManager *)popManager error:(NSError **)error
 {
+    popManager = self.externalPopManager ?: popManager;
     NSString *signedAccessToken = [popManager createSignedAccessToken:accessToken.accessToken
                                                            httpMethod:MSALParameterStringForHttpMethod(self.httpMethod)
                                                            requestUrl:self.requestUrl.absoluteString
--- a/MSAL/src/MSALAuthenticationSchemePop.m
+++ b/MSAL/src/MSALAuthenticationSchemePop.m
@@ -43,8 +43,8 @@ static NSString *keyDelimiter = @" ";
 @property (nonatomic) NSURL *requestUrl;
 @property (nonatomic) NSString *nonce;
 @property (nonatomic) NSDictionary *additionalParameters;
-@property (nonatomic) MSALExternalKeyPair *externalKeyPair;
-@property (nonatomic) MSIDDevicePopManager *externalPopManager;
+@property (nonatomic, nullable) MSALExternalKeyPair *externalKeyPair;
+@property (nonatomic, nullable) MSIDDevicePopManager *externalPopManager;
 
 @end
 
@@ -52,8 +52,8 @@ static NSString *keyDelimiter = @" ";
 
 - (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
                         requestUrl:(NSURL *)requestUrl
-                             nonce:(NSString *)nonce
-              additionalParameters:(NSDictionary *)additionalParameters
+                             nonce:(nullable NSString *)nonce
+              additionalParameters:(nullable NSDictionary *)additionalParameters
 {
     return [self initWithHttpMethod:httpMethod
                         requestUrl:requestUrl
@@ -64,8 +64,8 @@ static NSString *keyDelimiter = @" ";
 
 - (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
                         requestUrl:(NSURL *)requestUrl
-                             nonce:(NSString *)nonce
-              additionalParameters:(NSDictionary *)additionalParameters
+                             nonce:(nullable NSString *)nonce
+              additionalParameters:(nullable NSDictionary *)additionalParameters
                    externalKeyPair:(nullable MSALExternalKeyPair *)externalKeyPair
 {
     self = [super init];
@@ -100,9 +100,9 @@ static NSString *keyDelimiter = @" ";
 
 - (NSDictionary *)getSchemeParameters:(MSIDDevicePopManager *)popManager
 {
-    popManager = self.externalPopManager ?: popManager;
+    MSIDDevicePopManager *manager = self.externalPopManager ?: popManager;
     NSMutableDictionary *schemeParams = [NSMutableDictionary new];
-    NSString *requestConf = popManager.keyPair.jsonWebKey;
+    NSString *requestConf = manager.keyPair.jsonWebKey;
     if (requestConf)
     {
         [schemeParams setObject:MSALParameterStringForAuthScheme(self.scheme) forKey:MSID_OAUTH2_TOKEN_TYPE];
@@ -130,12 +130,12 @@ static NSString *keyDelimiter = @" ";
 
 - (nullable NSString *)getClientAccessToken:(MSIDAccessToken *)accessToken popManager:(nullable MSIDDevicePopManager *)popManager error:(NSError **)error
 {
-    popManager = self.externalPopManager ?: popManager;
-    NSString *signedAccessToken = [popManager createSignedAccessToken:accessToken.accessToken
-                                                           httpMethod:MSALParameterStringForHttpMethod(self.httpMethod)
-                                                           requestUrl:self.requestUrl.absoluteString
-                                                                nonce:self.nonce
-                                                                error:error];
+    MSIDDevicePopManager *manager = self.externalPopManager ?: popManager;
+    NSString *signedAccessToken = [manager createSignedAccessToken:accessToken.accessToken
+                                                        httpMethod:MSALParameterStringForHttpMethod(self.httpMethod)
+                                                        requestUrl:self.requestUrl.absoluteString
+                                                             nonce:self.nonce
+                                                             error:error];
     
     if (!signedAccessToken)
     {
--- a/MSAL/src/MSALAuthenticationSchemePop.m
+++ b/MSAL/src/MSALAuthenticationSchemePop.m
@@ -62,11 +62,11 @@ static NSString *keyDelimiter = @" ";
                    externalKeyPair:nil];
 }
 
-- (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
-                        requestUrl:(NSURL *)requestUrl
-                             nonce:(nullable NSString *)nonce
-              additionalParameters:(nullable NSDictionary *)additionalParameters
-                   externalKeyPair:(nullable MSALExternalKeyPair *)externalKeyPair
+- (nullable instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
+                                  requestUrl:(NSURL *)requestUrl
+                                       nonce:(nullable NSString *)nonce
+                        additionalParameters:(nullable NSDictionary *)additionalParameters
+                             externalKeyPair:(nullable MSALExternalKeyPair *)externalKeyPair
 {
     self = [super init];
     if (self)
@@ -76,10 +76,19 @@ static NSString *keyDelimiter = @" ";
         _requestUrl = requestUrl;
         _nonce = nonce ? nonce : [[NSUUID new] UUIDString];
         _additionalParameters = additionalParameters ? additionalParameters : [NSDictionary new];
-        _externalKeyPair = externalKeyPair;
         if (externalKeyPair)
         {
-            _externalPopManager = [[MSIDDevicePopManager alloc] initWithExternalKeyPair:externalKeyPair.msidKeyPair];
+            NSError *externalManagerError = nil;
+            _externalPopManager = [[MSIDDevicePopManager alloc] initWithExternalKeyPair:externalKeyPair.msidKeyPair
+                                                                                context:nil
+                                                                                  error:&externalManagerError];
+            if (!_externalPopManager)
+            {
+                MSID_LOG_WITH_CTX(MSIDLogLevelError, nil, @"Failed to initialize external AT PoP manager: %@", externalManagerError);
+                return nil;
+            }
+
+            _externalKeyPair = externalKeyPair;
         }
     }
 
@@ -107,7 +116,7 @@ static NSString *keyDelimiter = @" ";
     {
         [schemeParams setObject:MSALParameterStringForAuthScheme(self.scheme) forKey:MSID_OAUTH2_TOKEN_TYPE];
         [schemeParams setObject:requestConf forKey:MSID_OAUTH2_REQUEST_CONFIRMATION];
-        if (self.externalKeyPair)
+        if (manager == self.externalPopManager)
         {
             [schemeParams setObject:@"1" forKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
         }
--- a/MSAL/src/MSALAuthenticationSchemePop.m
+++ b/MSAL/src/MSALAuthenticationSchemePop.m
@@ -107,6 +107,10 @@ static NSString *keyDelimiter = @" ";
     {
         [schemeParams setObject:MSALParameterStringForAuthScheme(self.scheme) forKey:MSID_OAUTH2_TOKEN_TYPE];
         [schemeParams setObject:requestConf forKey:MSID_OAUTH2_REQUEST_CONFIRMATION];
+        if (self.externalKeyPair)
+        {
+            [schemeParams setObject:@"1" forKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
+        }
     }
     else
     {
--- a/MSAL/src/MSALExternalKeyPair.m
+++ b/MSAL/src/MSALExternalKeyPair.m
@@ -5,6 +5,24 @@
 //
 // This code is licensed under the MIT License.
 //
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
 //------------------------------------------------------------------------------
 
 #import "MSALExternalKeyPair+Internal.h"
@@ -14,11 +32,40 @@
 
 NSString *const MSALExternalKeyPairFailureReasonKey = @"MSALExternalKeyPairFailureReasonKey";
 
+static MSALExternalKeyPairFailureReason MSALFailureReasonFromMSIDReason(MSIDExternalKeyPairValidationFailureReason reason)
+{
+    switch (reason)
+    {
+        case MSIDExternalKeyPairValidationFailureReasonUnsupportedKeyType:
+            return MSALExternalKeyPairFailureReasonUnsupportedKeyType;
+
+        case MSIDExternalKeyPairValidationFailureReasonKeySizeTooSmall:
+            return MSALExternalKeyPairFailureReasonKeySizeTooSmall;
+
+        case MSIDExternalKeyPairValidationFailureReasonInvalidKeyClass:
+            return MSALExternalKeyPairFailureReasonInvalidKeyClass;
+
+        case MSIDExternalKeyPairValidationFailureReasonNotSigningCapable:
+            return MSALExternalKeyPairFailureReasonNotSigningCapable;
+
+        case MSIDExternalKeyPairValidationFailureReasonKeyPairMismatch:
+            return MSALExternalKeyPairFailureReasonKeyPairMismatch;
+
+        case MSIDExternalKeyPairValidationFailureReasonPublicKeySerializationFailed:
+            return MSALExternalKeyPairFailureReasonPublicKeySerializationFailed;
+
+        case MSIDExternalKeyPairValidationFailureReasonNone:
+        case MSIDExternalKeyPairValidationFailureReasonInvalidKeyHandle:
+        default:
+            return MSALExternalKeyPairFailureReasonInvalidKeyHandle;
+    }
+}
+
 @implementation MSALExternalKeyPair
 
-- (instancetype)initWithPrivateKey:(SecKeyRef)privateKey
-                         publicKey:(SecKeyRef)publicKey
-                             error:(NSError *__autoreleasing *)error
+- (nullable instancetype)initWithPrivateKey:(SecKeyRef)privateKey
+                                  publicKey:(SecKeyRef)publicKey
+                                      error:(NSError * _Nullable __autoreleasing * _Nullable)error
 {
     MSIDExternalKeyPairValidationFailureReason validationReason = MSIDExternalKeyPairValidationFailureReasonNone;
     NSError *validationError = nil;
@@ -36,7 +83,7 @@ NSString *const MSALExternalKeyPairFailureReasonKey = @"MSALExternalKeyPairFailu
                                          code:MSALErrorInvalidExternalKeyPair
                                      userInfo:@{
                                          MSALErrorDescriptionKey : description,
-                                         MSALExternalKeyPairFailureReasonKey : @([self msalFailureReasonFromMSIDReason:validationReason])
+                                         MSALExternalKeyPairFailureReasonKey : @(MSALFailureReasonFromMSIDReason(validationReason))
                                      }];
         }
 
@@ -71,35 +118,6 @@ NSString *const MSALExternalKeyPairFailureReasonKey = @"MSALExternalKeyPairFailu
 - (NSString *)keyId
 {
     return self.msidKeyPair.kid;
-}
-
-- (MSALExternalKeyPairFailureReason)msalFailureReasonFromMSIDReason:(MSIDExternalKeyPairValidationFailureReason)reason
-{
-    switch (reason)
-    {
-        case MSIDExternalKeyPairValidationFailureReasonUnsupportedKeyType:
-            return MSALExternalKeyPairFailureReasonUnsupportedKeyType;
-
-        case MSIDExternalKeyPairValidationFailureReasonKeySizeTooSmall:
-            return MSALExternalKeyPairFailureReasonKeySizeTooSmall;
-
-        case MSIDExternalKeyPairValidationFailureReasonInvalidKeyClass:
-            return MSALExternalKeyPairFailureReasonInvalidKeyClass;
-
-        case MSIDExternalKeyPairValidationFailureReasonNotSigningCapable:
-            return MSALExternalKeyPairFailureReasonNotSigningCapable;
-
-        case MSIDExternalKeyPairValidationFailureReasonKeyPairMismatch:
-            return MSALExternalKeyPairFailureReasonKeyPairMismatch;
-
-        case MSIDExternalKeyPairValidationFailureReasonPublicKeySerializationFailed:
-            return MSALExternalKeyPairFailureReasonPublicKeySerializationFailed;
-
-        case MSIDExternalKeyPairValidationFailureReasonNone:
-        case MSIDExternalKeyPairValidationFailureReasonInvalidKeyHandle:
-        default:
-            return MSALExternalKeyPairFailureReasonInvalidKeyHandle;
-    }
 }
 
 @end
--- a/MSAL/src/MSALExternalKeyPair.m
+++ b/MSAL/src/MSALExternalKeyPair.m
@@ -1,0 +1,104 @@
+//------------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+//------------------------------------------------------------------------------
+
+#import "MSALExternalKeyPair+Internal.h"
+#import "MSALError.h"
+#import "MSIDAssymetricKeyPair.h"
+#import "MSIDKeyOperationUtil.h"
+
+NSString *const MSALExternalKeyPairFailureReasonKey = @"MSALExternalKeyPairFailureReasonKey";
+
+@implementation MSALExternalKeyPair
+
+- (instancetype)initWithPrivateKey:(SecKeyRef)privateKey
+                         publicKey:(SecKeyRef)publicKey
+                             error:(NSError *__autoreleasing *)error
+{
+    MSIDExternalKeyPairValidationFailureReason validationReason = MSIDExternalKeyPairValidationFailureReasonNone;
+    NSError *validationError = nil;
+    BOOL valid = [[MSIDKeyOperationUtil sharedInstance] validateExternalRSAKeyPair:privateKey
+                                                                         publicKey:publicKey
+                                                                     failureReason:&validationReason
+                                                                             error:&validationError];
+    if (!valid)
+    {
+        if (error)
+        {
+            NSString *description = validationError.localizedDescription ?: @"Invalid external AT PoP key pair.";
+            *error = [NSError errorWithDomain:MSALErrorDomain
+                                         code:MSALErrorInvalidExternalKeyPair
+                                     userInfo:@{
+                                         MSALErrorDescriptionKey : description,
+                                         MSALExternalKeyPairFailureReasonKey : @([self msalFailureReasonFromMSIDReason:validationReason])
+                                     }];
+        }
+
+        return nil;
+    }
+
+    self = [super init];
+    if (self)
+    {
+        _msidKeyPair = [[MSIDAssymetricKeyPair alloc] initWithPrivateKey:privateKey
+                                                               publicKey:publicKey
+                                                          privateKeyDict:[NSDictionary new]];
+        if (!_msidKeyPair || !_msidKeyPair.kid)
+        {
+            if (error)
+            {
+                *error = [NSError errorWithDomain:MSALErrorDomain
+                                             code:MSALErrorInvalidExternalKeyPair
+                                         userInfo:@{
+                                             MSALErrorDescriptionKey : @"Unable to derive a public key thumbprint.",
+                                             MSALExternalKeyPairFailureReasonKey : @(MSALExternalKeyPairFailureReasonPublicKeySerializationFailed)
+                                         }];
+            }
+
+            return nil;
+        }
+    }
+
+    return self;
+}
+
+- (NSString *)keyId
+{
+    return self.msidKeyPair.kid;
+}
+
+- (MSALExternalKeyPairFailureReason)msalFailureReasonFromMSIDReason:(MSIDExternalKeyPairValidationFailureReason)reason
+{
+    switch (reason)
+    {
+        case MSIDExternalKeyPairValidationFailureReasonUnsupportedKeyType:
+            return MSALExternalKeyPairFailureReasonUnsupportedKeyType;
+
+        case MSIDExternalKeyPairValidationFailureReasonKeySizeTooSmall:
+            return MSALExternalKeyPairFailureReasonKeySizeTooSmall;
+
+        case MSIDExternalKeyPairValidationFailureReasonInvalidKeyClass:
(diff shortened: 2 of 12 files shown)


MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,104 @@
+//------------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
```
kaisong1990, line 6: The license header on the three new files is truncated - it stops after the `This code is licensed under the MIT License.` line. Every other source file in this repo carries the full MIT block including the permission and warranty paragraphs. Please copy the header from an existing file (e.g. MSALAuthenticationSchemePop.h).

MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,104 @@
...
+    if (!valid)
+    {
+        if (error)
+        {
+            NSString *description = validationError.localizedDescription ?: @"Invalid external AT PoP key pair.";
+            *error = [NSError errorWithDomain:MSALErrorDomain
+                                         code:MSALErrorInvalidExternalKeyPair
+                                     userInfo:@{
+                                         MSALErrorDescriptionKey : description,
+                                         MSALExternalKeyPairFailureReasonKey : @([self msalFailureReasonFromMSIDReason:validationReason])
```
kaisong1990, line 38: `[self msalFailureReasonFromMSIDReason:]` is called before `self = [super init]` on line 45, so we're messaging a raw alloc'd object. It happens to work today because the method touches no ivars, but it's fragile. Make it a class method or a static C function.

MSAL/src/MSALAuthenticationSchemePop.m
```diff
@@ -79,12 +100,17 @@ - (MSIDAuthenticationScheme *)createMSIDAuthenticationSchemeWithParams:(nullable
 
 - (NSDictionary *)getSchemeParameters:(MSIDDevicePopManager *)popManager
 {
+    popManager = self.externalPopManager ?: popManager;
```
kaisong1990, line 103: Reassigning the incoming `popManager` parameter is easy to miss when reading this and the same line is repeated in getClientAccessToken:. Use a local (`MSIDDevicePopManager *manager = self.externalPopManager ?: popManager;`) so the caller-supplied argument stays intact.

MSAL/src/MSALAuthenticationSchemePop.m
```diff
@@ -42,6 +43,8 @@ @interface MSALAuthenticationSchemePop()
 @property (nonatomic) NSURL *requestUrl;
 @property (nonatomic) NSString *nonce;
 @property (nonatomic) NSDictionary *additionalParameters;
+@property (nonatomic) MSALExternalKeyPair *externalKeyPair;
+@property (nonatomic) MSIDDevicePopManager *externalPopManager;
```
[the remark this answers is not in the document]
antonioalwan: Fixed in 14df482ae: the class-extension properties and nullable initializer parameters now match the public contract.

MSAL/test/unit/MSALExternalKeyPairTests.m
```diff
@@ -0,0 +1,157 @@
...
+{
+    NSDictionary *attributes = @{
+        (id)kSecAttrKeyType : (__bridge id)keyType,
+        (id)kSecAttrKeySizeInBits : size
+    };
+    CFErrorRef error = NULL;
+    SecKeyRef privateKey = SecKeyCreateRandomKey((__bridge CFDictionaryRef)attributes, &error);
+    XCTAssertNotEqual(privateKey, NULL);
+    XCTAssertEqual(error, NULL);
+    return privateKey;
```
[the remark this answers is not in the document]
antonioalwan: Fixed in 14df482ae: the test helper now releases the CFErrorRef when key creation returns one.

MSAL/test/unit/MSALAuthSchemeTests.m
```diff
@@ -165,6 +168,55 @@ - (void)test_getAccessToken_fromBearerAuthScheme_ValidInput_shouldReturnAT
...
+    NSData *signedData = [[NSString stringWithFormat:@"%@.%@", segments[0], segments[1]] dataUsingEncoding:NSUTF8StringEncoding];
+    NSData *signature = [NSData msidDataFromBase64UrlEncodedString:segments[2]];
+    CFErrorRef verificationError = NULL;
+    BOOL signatureValid = SecKeyVerifySignature(publicKey,
+                                                kSecKeyAlgorithmRSASignatureMessagePKCS1v15SHA256,
+                                                (__bridge CFDataRef)signedData,
+                                                (__bridge CFDataRef)signature,
+                                                &verificationError);
+    XCTAssertTrue(signatureValid);
+    XCTAssertEqual(verificationError, NULL);
```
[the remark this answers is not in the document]
antonioalwan: Fixed in 14df482ae: both the signature-verification error and random-key creation error are released when non-NULL.

MSAL/src/MSALExternalKeyPair.m
antonioalwan: Fixed in 14df482ae: the complete repository-standard MIT header is now present on the new files.

MSAL/src/MSALExternalKeyPair.m
antonioalwan: Fixed in 14df482ae: failure-reason mapping is now a static function, so no instance message is sent before super initialization.

MSAL/src/MSALAuthenticationSchemePop.m
antonioalwan: Fixed in 14df482ae: both paths now use a local manager variable instead of reassigning the incoming popManager parameter.

antonioalwan:
@copilot resolve the merge conflicts in this pull request

MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,123 @@
...
+
+        case MSIDExternalKeyPairValidationFailureReasonKeyPairMismatch:
+            return MSALExternalKeyPairFailureReasonKeyPairMismatch;
+
+        case MSIDExternalKeyPairValidationFailureReasonPublicKeySerializationFailed:
+            return MSALExternalKeyPairFailureReasonPublicKeySerializationFailed;
+
+        case MSIDExternalKeyPairValidationFailureReasonNone:
+        case MSIDExternalKeyPairValidationFailureReasonInvalidKeyHandle:
+        default:
```
kaisong1990: `default:` alongside the enumerated cases turns off -Wswitch here, so when IdentityCore adds a new `MSIDExternalKeyPairValidationFailureReason` we'll silently map it to InvalidKeyHandle instead of getting a build warning. Drop `default:` and keep the two explicit cases.

MSAL/src/public/MSALExternalKeyPair.h
```diff
@@ -0,0 +1,69 @@
...
+    MSALExternalKeyPairFailureReasonInvalidKeyHandle = 1,
+    MSALExternalKeyPairFailureReasonUnsupportedKeyType,
+    MSALExternalKeyPairFailureReasonKeySizeTooSmall,
+    MSALExternalKeyPairFailureReasonInvalidKeyClass,
+    MSALExternalKeyPairFailureReasonNotSigningCapable,
+    MSALExternalKeyPairFailureReasonKeyPairMismatch,
+    MSALExternalKeyPairFailureReasonPublicKeySerializationFailed
+};
+
+extern NSString *const MSALExternalKeyPairFailureReasonKey;
```
kaisong1990: Every other public error userInfo key (MSALOAuthErrorKey, MSALInternalErrorCodeKey, etc.) is declared in MSALError.h as `extern NSString *`. Move this one there and match the declaration style so callers have one place to look.

MSAL/src/public/MSALAuthenticationSchemePop.h
```diff
@@ -27,17 +27,26 @@
...
 - (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
                         requestUrl:(NSURL *)requestUrl
                              nonce:(nullable NSString *)nonce
               additionalParameters:(nullable NSDictionary *)additionalParameters;
 
+- (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
+                        requestUrl:(NSURL *)requestUrl
+                             nonce:(nullable NSString *)nonce
+              additionalParameters:(nullable NSDictionary *)additionalParameters
+                   externalKeyPair:(nullable MSALExternalKeyPair *)externalKeyPair;
```
kaisong1990: Now that the 4-arg init just forwards to this one, mark this `NS_DESIGNATED_INITIALIZER`.

MSAL/test/unit/MSALExternalKeyPairTests.m
```diff
@@ -0,0 +1,180 @@
...
+
+- (SecKeyRef)createPrivateKeyWithType:(CFStringRef)keyType size:(NSNumber *)size
+{
+    NSDictionary *attributes = @{
+        (id)kSecAttrKeyType : (__bridge id)keyType,
+        (id)kSecAttrKeySizeInBits : size
+    };
+    CFErrorRef error = NULL;
+    SecKeyRef privateKey = SecKeyCreateRandomKey((__bridge CFDictionaryRef)attributes, &error);
+    XCTAssertNotEqual(privateKey, NULL);
```
kaisong1990, line 204: XCTAssert doesn't abort the test, so if key generation fails this returns NULL and every caller then hits `CFRelease(privateKey)` on NULL and crashes the whole test run. Either `XCTAssertNotEqual` -> early return with the caller nil-checking, or set `self.continueAfterFailure = NO` in setUp.

MSAL/src/MSALAuthenticationSchemePop.m
```diff
@@ -60,6 +76,11 @@ - (instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
         _requestUrl = requestUrl;
         _nonce = nonce ? nonce : [[NSUUID new] UUIDString];
         _additionalParameters = additionalParameters ? additionalParameters : [NSDictionary new];
+        _externalKeyPair = externalKeyPair;
+        if (externalKeyPair)
+        {
+            _externalPopManager = [[MSIDDevicePopManager alloc] initWithExternalKeyPair:externalKeyPair.msidKeyPair];
```
josephpab: `initWithExternalKeyPair:` returns nil when its argument is nil, so `_externalPopManager` can end up nil while `_externalKeyPair` is set. Today `MSALExternalKeyPair`'s init guarantees `msidKeyPair` is non-nil so we never actually hit it, but nothing in this file enforces that — it's an implicit contract spanning two classes.

The failure mode is bad if it ever breaks: every downstream `self.externalPopManager ?: popManager` silently falls back to the device key while `external_key_pop=1` still goes on the wire (see below). We'd send a device `req_cnf`, tell broker it's an external key, and sign with the device key, with no error surfaced anywhere.

This initializer returns a non-null `instancetype` with no error out-param, so there's no way to report the failure. Either assert/log loudly here, or reconsider the signature so a bad key pair can't produce a half-configured scheme object.

MSAL/src/MSALAuthenticationSchemePop.m
```diff
@@ -79,12 +100,17 @@ - (MSIDAuthenticationScheme *)createMSIDAuthenticationSchemeWithParams:(nullable
...
-    NSString *requestConf = popManager.keyPair.jsonWebKey;
+    NSString *requestConf = manager.keyPair.jsonWebKey;
     if (requestConf)
     {
         [schemeParams setObject:MSALParameterStringForAuthScheme(self.scheme) forKey:MSID_OAUTH2_TOKEN_TYPE];
         [schemeParams setObject:requestConf forKey:MSID_OAUTH2_REQUEST_CONFIRMATION];
+        if (self.externalKeyPair)
+        {
+            [schemeParams setObject:@"1" forKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
+        }
```
josephpab: This gates on `self.externalKeyPair`, but `requestConf` above came from `manager`, which is `self.externalPopManager ?: popManager`. Those two can disagree (see my comment on the initializer), and when they do we stamp `external_key_pop=1` onto a `req_cnf` derived from the device key.

Gate on the manager that actually produced the confirmation:

```objc
if (manager == self.externalPopManager)
```

Then the flag and the key can't diverge — which matters, because broker uses this flag for rollout classification.

MSAL/src/MSALAuthenticationSchemePop.m
```diff
@@ -104,11 +130,12 @@ - (NSDictionary *)getSchemeParameters:(MSIDDevicePopManager *)popManager
...
-                                                           httpMethod:MSALParameterStringForHttpMethod(self.httpMethod)
-                                                           requestUrl:self.requestUrl.absoluteString
-                                                                nonce:self.nonce
-                                                                error:error];
+    MSIDDevicePopManager *manager = self.externalPopManager ?: popManager;
+    NSString *signedAccessToken = [manager createSignedAccessToken:accessToken.accessToken
+                                                        httpMethod:MSALParameterStringForHttpMethod(self.httpMethod)
+                                                        requestUrl:self.requestUrl.absoluteString
+                                                             nonce:self.nonce
+                                                             error:error];
```
josephpab, line 147: Worth thinking through the broker compatibility story here.

`MSIDAuthenticationSchemePop.accessToken` sets the AT's `kid` from the *requested* `req_cnf`, not from anything the server or broker echoes back. So MSAL never verifies that the token it received is actually bound to the key it's about to sign with.

For device-key PoP that was safe — MSAL and broker share the same device key, so there was nothing to diverge. With caller-supplied keys that's no longer true: a broker build that doesn't understand `external_key_pop` can bind the token to its own device key, and we'll still sign the SHR with the caller's private key. The app then gets an opaque 401 from the resource with nothing in MSAL logs pointing at the cause.

`external_key_pop=1` is a request-side hint only, and I don't see any minimum-broker-version gating on it in `MSIDBrokerTokenRequest` — it's added unconditionally. Is broker guaranteed to hard-fail on an unknown-but-present `external_key_pop`, or is there a version floor we should enforce before sending it? If neither, comparing `accessToken.kid` against `manager.keyPair.kid` here before signing would turn a silent resource-side failure into an actionable client error.

MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,123 @@
...
+            return MSALExternalKeyPairFailureReasonInvalidKeyHandle;
```
josephpab: Two problems in this block.

**`default:` defeats exhaustiveness checking.** Every enum case is already listed above, so the `default:` only serves to suppress `-Wswitch`. MSAL and CommonCore version independently, so when a new `MSIDExternalKeyPairValidationFailureReason` is added upstream it will silently map to `InvalidKeyHandle` and nobody finds out until a customer reports a misleading error. Drop `default:` and move the fallback `return` below the switch — then a new upstream value becomes a build break, which is what you want across a submodule boundary.

**`...ReasonNone` → `InvalidKeyHandle` is misleading.** `None` means validation failed but didn't set a reason — a contract bug between MSAL and CommonCore, not a bad key handle. Reporting it as an invalid handle sends integrators debugging the wrong thing. This wants a distinct value (see my suggestion about adding an `Unknown = 0` case in the header).

MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,123 @@
...
+    {
+        if (error)
+        {
+            NSString *description = validationError.localizedDescription ?: @"Invalid external AT PoP key pair.";
+            *error = [NSError errorWithDomain:MSALErrorDomain
+                                         code:MSALErrorInvalidExternalKeyPair
+                                     userInfo:@{
+                                         MSALErrorDescriptionKey : description,
+                                         MSALExternalKeyPairFailureReasonKey : @(MSALFailureReasonFromMSIDReason(validationReason))
+                                     }];
```
josephpab: `validationError` is reduced to its `localizedDescription` and then discarded. For a security-sensitive validation path that drops a lot of diagnostic signal — the underlying MSID error code and its `userInfo` are gone by the time the caller sees this, and there's no correlation ID either since `context:nil` is passed on line 75.

Please attach the original as `NSUnderlyingErrorKey` so integrators and our own logs can see what CommonCore actually rejected.

MSAL/src/public/MSALExternalKeyPair.h
```diff
@@ -0,0 +1,69 @@
...
+
+ MSAL retains the key handles for the lifetime of this object. The caller remains
+ responsible for creating, storing, rotating, and deleting the underlying keys.
+ MSAL never exports or persists the private key material.
+ */
+@interface MSALExternalKeyPair : NSObject
+
+- (nullable instancetype)initWithPrivateKey:(SecKeyRef)privateKey
+                                  publicKey:(SecKeyRef)publicKey
+                                      error:(NSError * _Nullable __autoreleasing * _Nullable)error NS_DESIGNATED_INITIALIZER;
```
josephpab: These parameters sit inside `NS_ASSUME_NONNULL_BEGIN`, and clang does audit CF pointer types — `SecKeyRef privateKey` / `publicKey` are implicitly `_Nonnull` here. I verified that passing `NULL` directly to a signature shaped like this produces `-Wnonnull`.

That contradicts everything else in the PR:

- CommonCore declares `validateExternalRSAKeyPair:(SecKeyRef _Nullable)privateKey publicKey:(SecKeyRef _Nullable)publicKey` — explicitly nullable
- the implementation deliberately handles NULL and maps it to `MSALExternalKeyPairFailureReasonInvalidKeyHandle`
- `MSALExternalKeyPairTests` has two tests for exactly that case, and they only compile warning-free because they route `NULL` through a local variable first

The Swift consequence is the real issue: an audited non-null `SecKeyRef` projects as a non-optional `SecKey`, so a Swift caller cannot reach `InvalidKeyHandle` at all. Either mark both params `_Nullable` to match the actual contract, or make them genuinely non-null and delete the NULL tests. Right now the header and the behavior disagree.

MSAL/src/public/MSALExternalKeyPair.h
```diff
@@ -0,0 +1,69 @@
...
+//------------------------------------------------------------------------------
+
+#import <Foundation/Foundation.h>
+#import <Security/Security.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NS_ENUM(NSInteger, MSALExternalKeyPairFailureReason)
+{
+    MSALExternalKeyPairFailureReasonInvalidKeyHandle = 1,
```
josephpab: This enum starts at 1 with no zero case, but callers read it as `[error.userInfo[MSALExternalKeyPairFailureReasonKey] integerValue]`, which returns `0` whenever the key is absent — and `0` isn't a valid `MSALExternalKeyPairFailureReason`. Swift has the same gap: `MSALExternalKeyPairFailureReason(rawValue: 0)` is `nil`.

Adding an explicit `MSALExternalKeyPairFailureReasonUnknown = 0` makes "reason absent" a defined value instead of an out-of-range one, and gives you something honest to map `MSIDExternalKeyPairValidationFailureReasonNone` onto instead of `InvalidKeyHandle`.

Since this enum is public API, it's much cheaper to add the case now than after it ships.

MSAL/src/public/MSALAuthenticationSchemePop.h
josephpab: New public initializer on a security-sensitive API with no doc comment. The existing 4-arg one has none either, but this one carries obligations that aren't discoverable from the signature:

- passing nil `externalKeyPair` silently falls back to the device key rather than failing
- the caller owns key creation, storage, rotation, and deletion
- **the same key has to be supplied on subsequent silent calls**, or the cached PoP token won't match on `kid` and you get a cache miss plus a fresh network request with no obvious explanation

That last one is the easiest for an integrator to get wrong and the hardest to diagnose. It should be documented on the API, not just in the PR description.

Also worth marking this `NS_DESIGNATED_INITIALIZER` now that the 4-arg version delegates to it.

MSAL/test/unit/MSALExternalKeyPairTests.m
```diff
@@ -0,0 +1,180 @@
...
+
+@end
+
+@implementation MSALExternalKeyPairTests
+
+- (void)testInitWithNilPrivateKey_ShouldReturnInvalidKeyHandle
+{
+    SecKeyRef privateKey = [self createPrivateKeyWithType:kSecAttrKeyTypeRSA size:@2048];
+    SecKeyRef publicKey = SecKeyCopyPublicKey(privateKey);
+    SecKeyRef nilPrivateKey = NULL;
```
josephpab: Assigning `NULL` to a local first is what lets this compile without `-Wnonnull` — the header declares these parameters non-null (see my comment on `MSALExternalKeyPair.h`). Passing `NULL` inline would warn.

Flagging it because it means this test asserts behavior the public contract says is impossible, and the indirection hides that. Once the nullability is resolved one way or the other, this either becomes a legitimate direct `NULL` test or it goes away.

MSAL/test/unit/MSALExternalKeyPairTests.m
```diff
@@ -0,0 +1,180 @@
...
+                                                                         publicKey:publicKey
+                                                                             error:&error];
+
+    XCTAssertNotNil(keyPair);
+    XCTAssertNotNil(keyPair.keyId);
+    XCTAssertNil(error);
+
+    CFRelease(publicKey);
+    CFRelease(privateKey);
+    XCTAssertNotNil(keyPair.keyId);
```
josephpab: This assertion doesn't test what it appears to test. `keyId` was already materialized during `init` (the thumbprint check on line 99 of `MSALExternalKeyPair.m` reads `msidKeyPair.kid`), and `MSIDAssymetricKeyPair` caches `_kid`. So after these `CFRelease` calls this just re-reads a cached `NSString` — it would pass even if the `CFRetain`s were missing entirely.

If the intent is to prove `MSIDAssymetricKeyPair` takes its own reference on the `SecKeyRef`s (and the public header's "MSAL retains the key handles for the lifetime of this object" claim depends on exactly that), exercise something that actually touches the key material after release — e.g. sign through `MSIDDevicePopManager` and verify the output.

MSAL/test/unit/MSALAuthSchemeTests.m
```diff
@@ -165,6 +168,59 @@ - (void)test_getAccessToken_fromBearerAuthScheme_ValidInput_shouldReturnAT
     XCTAssertTrue([signedAccessToken isEqualToString:msidAccessToken.accessToken]);
 }
 
+- (void)testExternalKeyPair_getSchemeParametersAndSignedToken_ShouldUseExternalKey
+{
+    SecKeyRef privateKey = [self createRSA2048PrivateKey];
+    SecKeyRef publicKey = SecKeyCopyPublicKey(privateKey);
```
josephpab, line 181: `SecKeyCopyPublicKey` can return NULL. It's fed straight into the initializer here and into `SecKeyVerifySignature` on line 208. Add `XCTAssertNotEqual(publicKey, NULL);` so a key-derivation failure reports as one, rather than surfacing as a confusing signature-verification failure 35 lines later.

MSAL/test/unit/MSALAuthSchemeTests.m
```diff
@@ -165,6 +168,59 @@ - (void)test_getAccessToken_fromBearerAuthScheme_ValidInput_shouldReturnAT
...
+                                                                                           requestUrl:requestUrl
+                                                                                                nonce:@"external-key-nonce"
+                                                                                 additionalParameters:nil
+                                                                                      externalKeyPair:externalKeyPair];
+    MSIDDevicePopManager *defaultManager = [MSALDevicePopManagerUtil test_initWithValidCacheConfig];
+    NSDictionary *schemeParameters = [authScheme getSchemeParameters:defaultManager];
+    NSString *requestConfirmation = schemeParameters[MSID_OAUTH2_REQUEST_CONFIRMATION];
+    NSString *decodedConfirmation = [requestConfirmation msidBase64UrlDecode];
+    XCTAssertTrue([decodedConfirmation containsString:externalKeyPair.keyId]);
+    XCTAssertEqualObjects(schemeParameters[MSID_OAUTH2_EXTERNAL_KEY_POP], @"1");
```
josephpab, line 201: Good positive assertion, but there's no test that `external_key_pop` is *absent* when the 4-arg initializer is used.

That flag is the rollout gate for broker. If it ever leaks onto regular device-key PoP requests we'd misclassify every existing PoP caller — exactly the kind of regression a cheap negative test catches and code review doesn't.

Please add the negative case against a scheme built with the non-external initializer, and assert there that `req_cnf` matches the device pop manager's `keyPair.jsonWebKey`. That second assertion would also catch the silent-fallback problem I flagged in `MSALAuthenticationSchemePop.m`.

MSAL/test/unit/MSALAuthSchemeTests.m
```diff
@@ -165,6 +168,59 @@ - (void)test_getAccessToken_fromBearerAuthScheme_ValidInput_shouldReturnAT
...
+    NSArray<NSString *> *segments = [signedAccessToken componentsSeparatedByString:@"."];
+    XCTAssertEqual(segments.count, 3);
+    NSString *decodedHeader = [segments[0] msidBase64UrlDecode];
+    XCTAssertTrue([decodedHeader containsString:externalKeyPair.keyId]);
+
+    NSData *signedData = [[NSString stringWithFormat:@"%@.%@", segments[0], segments[1]] dataUsingEncoding:NSUTF8StringEncoding];
+    NSData *signature = [NSData msidDataFromBase64UrlEncodedString:segments[2]];
+    CFErrorRef verificationError = NULL;
+    BOOL signatureValid = SecKeyVerifySignature(publicKey,
+                                                kSecKeyAlgorithmRSASignatureMessagePKCS1v15SHA256,
```
josephpab, line 223: The verification hardcodes `kSecKeyAlgorithmRSASignatureMessagePKCS1v15SHA256`, but nothing asserts the JWT header actually advertises `RS256`. If the declared `alg` and the real signing algorithm ever drift apart, this test still passes and the resource server is the one that finds out.

Line 203 only substring-matches the kid, which won't catch it either. Parse the decoded header as JSON and assert `alg` and `typ` explicitly — that also makes the kid check exact rather than a substring match that could pass on a coincidental prefix.

MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,123 @@
...
+
+        return nil;
+    }
+
+    self = [super init];
+    if (self)
+    {
+        _msidKeyPair = [[MSIDAssymetricKeyPair alloc] initWithPrivateKey:privateKey
+                                                               publicKey:publicKey
+                                                          privateKeyDict:[NSDictionary new]];
```
ameyapat, line 109: Why is an empty dict being passed here?

MSAL/src/MSALExternalKeyPair.m
antonioalwan: Addressed in 95c659560. I removed `default:`, made every CommonCore validation reason explicit, and retained a fallback return after the switch so a future upstream enum addition produces the expected `-Wswitch` diagnostic.

MSAL/src/public/MSALExternalKeyPair.h
antonioalwan: Addressed in 95c659560. `MSALExternalKeyPairFailureReasonKey` is now declared with the other public userInfo keys in `MSALError.h` and defined in `MSALError.m`.

MSAL/src/public/MSALAuthenticationSchemePop.h
antonioalwan: Addressed in 95c659560. The external-key initializer is now the designated initializer, while the 4-argument initializer remains the device-key convenience path.

MSAL/test/unit/MSALExternalKeyPairTests.m
antonioalwan: Addressed in 95c659560. Both affected test classes now set `continueAfterFailure = NO`, so failed key-generation assertions stop the test before callers can use or release a NULL handle.

MSAL/src/MSALAuthenticationSchemePop.m
antonioalwan: Addressed in 95c659560. Scheme initialization now uses CommonCore’s validating manager initializer and fails closed with an error log if the external manager cannot be created. The public initializer is nullable, and regression coverage verifies that an invalid internal pair cannot produce a half-configured scheme.

MSAL/src/MSALAuthenticationSchemePop.m
antonioalwan: Addressed in 95c659560. `external_key_pop=1` is now emitted only when the manager that actually produced `req_cnf` is `self.externalPopManager`, preventing marker and key-source divergence.

MSAL/src/MSALAuthenticationSchemePop.m
antonioalwan: I traced this through Broker and do not believe a compatibility guard or `kid` comparison is needed. Legacy Broker already copies the caller-provided `req_cnf` into `MSIDAuthenticationSchemePop`, and `MSIDBrokerTokenRequest` forwards that same value to ESTS; ignoring the newer `external_key_pop` marker does not substitute the device key. Also, `accessToken.kid` is populated from the originally requested `req_cnf`, so comparing it with the manager key here would always confirm the client request rather than validate server binding.

MSAL/src/MSALExternalKeyPair.m
antonioalwan: Addressed in 95c659560. The switch is exhaustive with no `default:`. CommonCore `None` now maps to the new public `Unknown = 0` value rather than `InvalidKeyHandle`, and `PublicKeyDerivationFailed` has its own explicit public mapping.

MSAL/src/MSALExternalKeyPair.m
antonioalwan: Addressed in 95c659560. The MSAL validation error now preserves the original CommonCore error under `NSUnderlyingErrorKey`, with test coverage confirming the underlying diagnostic remains available.

MSAL/src/public/MSALExternalKeyPair.h
antonioalwan: Addressed in 95c659560. Both `SecKeyRef` parameters are explicitly nullable in the public header and implementation, matching the failable initializer behavior and Swift optional projection.

MSAL/src/public/MSALExternalKeyPair.h
antonioalwan: Addressed in 95c659560. The public enum now defines `MSALExternalKeyPairFailureReasonUnknown = 0`, and missing reason metadata therefore has a valid Objective-C and Swift representation.

MSAL/src/public/MSALAuthenticationSchemePop.h
antonioalwan: Addressed in 95c659560. The initializer is documented and designated. The public documentation now explains nil device-key fallback, caller key-lifecycle ownership, and the requirement to supply the same external pair on later silent requests for cache matching.

MSAL/test/unit/MSALExternalKeyPairTests.m
antonioalwan: Addressed in 95c659560. After correcting the public nullability contract, the tests now pass `NULL` directly rather than hiding it in local variables.

MSAL/test/unit/MSALExternalKeyPairTests.m
antonioalwan: Addressed in 95c659560. The retention test now releases the caller-owned references, signs through `MSIDDevicePopManager`, and verifies the resulting JWT with the retained public-key handle, exercising both retained key handles rather than rereading cached `keyId` text.

MSAL/test/unit/MSALAuthSchemeTests.m
antonioalwan: Addressed in 95c659560. The test now asserts that `SecKeyCopyPublicKey` succeeded immediately, and `continueAfterFailure = NO` prevents a later misleading verification failure.

MSAL/test/unit/MSALAuthSchemeTests.m
antonioalwan: Addressed in 95c659560. A new device-key test verifies that `external_key_pop` is absent and that `req_cnf` exactly matches the supplied device manager’s JWK.

MSAL/test/unit/MSALAuthSchemeTests.m
antonioalwan: Addressed in 95c659560. The test now parses the JWT header as JSON and asserts exact `alg = RS256`, `typ = JWT`, and `kid` values before cryptographically verifying the signature.

MSAL/src/MSALExternalKeyPair.m
antonioalwan: This empty dictionary is intentional because caller-owned keys have no MSAL keychain lookup metadata, while the internal `MSIDAssymetricKeyPair` initializer requires a nonnull metadata dictionary. I added a clarifying comment in 95c659560.

MSAL/test/unit/MSALExternalKeyPairTests.m
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+- (void)testFailureReason_whenMissing_shouldReturnUnknown
+{
+    NSError *error = [NSError errorWithDomain:MSALErrorDomain
+                                         code:MSALErrorInvalidExternalKeyPair
+                                     userInfo:nil];
+
+    XCTAssertEqual([error.userInfo[MSALExternalKeyPairFailureReasonKey] integerValue], MSALExternalKeyPairFailureReasonUnknown);
```
kaisong1990: This test constructs its own NSError and never calls into MSAL. `error.userInfo[MSALExternalKeyPairFailureReasonKey]` is nil, `[nil integerValue]` is 0, and `MSALExternalKeyPairFailureReasonUnknown` is also 0 - so the assertion holds no matter what the production code does. It can't fail and gives no coverage.

If the intent is to cover the `MSIDExternalKeyPairValidationFailureReasonNone` -> `Unknown` mapping, drive it through `initWithPrivateKey:publicKey:error:` with an input that fails validation without setting a reason, and assert on the error MSAL produces.

antonioalwan: Agreed. This test only exercised Objective-C nil messaging and could not detect a production regression. I removed it in the latest commit. CommonCore currently assigns an explicit failure reason on every validation failure path, so there is no public key input that can exercise the defensive `None -> Unknown` mapping without introducing a test-only production seam.

MSAL/src/public/MSALAuthenticationSchemePop.h
```diff
@@ -27,17 +27,33 @@
...
+
+ Pass nil to use MSAL's device PoP key. The caller is responsible for creating, storing,
+ rotating, and deleting an external key pair. Supply the same external key pair on later
+ silent requests so cached PoP tokens can be matched by their key identifier.
+ */
+- (nullable instancetype)initWithHttpMethod:(MSALHttpMethod)httpMethod
+                                  requestUrl:(NSURL *)requestUrl
+                                       nonce:(nullable NSString *)nonce
+                        additionalParameters:(nullable NSDictionary *)additionalParameters
+                             externalKeyPair:(nullable MSALExternalKeyPair *)externalKeyPair NS_DESIGNATED_INITIALIZER;
```
jasoncoolmax, line 55: This initializer can return nil when the external pop manager fails to init, but we only `MSID_LOG` the reason in the `.m`, so the caller just gets nil with nothing to go on. `MSALExternalKeyPair`'s init already surfaces `MSALErrorInvalidExternalKeyPair` + `MSALExternalKeyPairFailureReasonKey`, will it be better to add an `error:` param here too so it is consistent? `testExternalKeyPair_whenInternalPairIsMissing_shouldFailInitialization` shows the nil path is reachable, and caller-owned keys can go bad after `MSALExternalKeyPair` was created (deleted / invalidated), so this is likely to be hit in the field :)

MSAL/src/MSALExternalKeyPair.m
```diff
@@ -0,0 +1,134 @@
...
+                                  publicKey:(SecKeyRef _Nullable)publicKey
+                                      error:(NSError * _Nullable __autoreleasing * _Nullable)error
+{
+    MSIDExternalKeyPairValidationFailureReason validationReason = MSIDExternalKeyPairValidationFailureReasonNone;
+    NSError *validationError = nil;
+    BOOL valid = [[MSIDKeyOperationUtil sharedInstance] validateExternalRSAKeyPair:privateKey
+                                                                         publicKey:publicKey
+                                                                     failureReason:&validationReason
+                                                                           context:nil
+                                                                             error:&validationError];
```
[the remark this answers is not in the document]
antonioalwan: This is a false positive: the submodule pointer read here is `dev`'s, not this PR's.

| | |
|---|---|
| `MSAL/IdentityCore` gitlink on this PR head (`5cb6536`) | `6eb0fef5cd6945f5b5074de5b9d9b6f94ab6c09c` |
| `MSAL/IdentityCore` gitlink on `dev` | `9f257218627d2a0061d82300a654914b09cfff7e` |
| CommonCore AzureAD/microsoft-authentication-library-common-for-objc#1934 | merged 2026-08-24T20:07Z, merge commit `6eb0fef5` |

The pointer on this branch is exactly the #1934 merge commit, and `MSIDKeyOperationUtil.h` at `6eb0fef5` declares `validateExternalRSAKeyPair:publicKey:failureReason:context:error:`. The pointer was advanced roughly 40 minutes before this comment was posted.

Verified by building this branch at its current submodule pointer:

- `MSAL (iOS Framework)` — BUILD SUCCEEDED
- `MSAL (Mac Framework)` — BUILD SUCCEEDED

Resolving.

MSAL/src/public/MSALExternalKeyPair.h
```diff
@@ -0,0 +1,69 @@
...
+    MSALExternalKeyPairFailureReasonPublicKeyDerivationFailed
+};
+
+/**
+ A caller-owned RSA key pair supplied to MSAL for Access Token Proof-of-Possession.
+
+ MSAL retains the key handles for the lifetime of this object. The caller remains
+ responsible for creating, storing, rotating, and deleting the underlying keys.
+ MSAL never exports or persists the private key material.
+ */
```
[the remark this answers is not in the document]
antonioalwan: Addressed. Added HeaderDoc to `MSALExternalKeyPair.h`:

**`MSALExternalKeyPairFailureReason`** now has a typedef-level block explaining that the value is read from `MSALExternalKeyPairFailureReasonKey` in the `userInfo` of an `MSALErrorInvalidExternalKeyPair` error, plus a per-case comment on all nine cases (`Unknown`, `InvalidKeyHandle`, `UnsupportedKeyType`, `KeySizeTooSmall`, `InvalidKeyClass`, `NotSigningCapable`, `KeyPairMismatch`, `PublicKeySerializationFailed`, `PublicKeyDerivationFailed`).

**`initWithPrivateKey:publicKey:error:`** now documents:
- the seven validation requirements (non-NULL readable handles, RSA only with EC/Secure Enclave explicitly rejected, one private plus one public, >= 2048 bits and equal size, RS256 sign/verify capability, exportable public key for the RFC 7638 thumbprint, derived public key matches the supplied one);
- that validation is non-interactive, so it never signs a challenge and never triggers a biometric or user-presence prompt;
- that it fails closed: if the public key cannot be derived from the private key the pair is rejected rather than accepted on weaker evidence;
- `@param`/`@return`, including the error contract (`MSALErrorDomain` / `MSALErrorInvalidExternalKeyPair`, with `MSALExternalKeyPairFailureReasonKey` and `NSUnderlyingErrorKey` in `userInfo`);
- ownership: MSAL retains the handles for the object's lifetime, the caller owns creation/storage/rotation/deletion, and MSAL never exports or persists private key material.

Verified with `MSAL (iOS Framework)` and `MSAL (Mac Framework)`, both **BUILD SUCCEEDED** with no `-Wdocumentation` warnings.

MSAL/src/public/MSALAuthenticationSchemePop.h
antonioalwan: The only way this initializer returns nil is if  [[MSIDDevicePopManager alloc] initWithExternalKeyPair:externalKeyPair.msidKeyPair ...]  fails, and that only fails when  msidKeyPair  is nil. That state is not reachable through the public API.  msidKeyPair  is declared in  MSALExternalKeyPair+Internal.h , so it is not visible to app code, and it is assigned in exactly one place:  MSALExternalKeyPair.m:107 , inside that class's own initializer. Nothing else ever writes it, so its value is fixed the moment the object exists.

That same initializer checks the result on the next line ( if (!_msidKeyPair || !_msidKeyPair.kid) ) and returns nil with  MSALErrorInvalidExternalKeyPair  plus  MSALExternalKeyPairFailureReasonKey  when it did not come out valid. The two outcomes are therefore a non-nil  MSALExternalKeyPair  with a valid pair, or nil with an error already in hand. There is no third state where the object exists but the pair is missing.

MSAL/test/unit/MSALExternalKeyPairTests.m
```diff
@@ -0,0 +1,214 @@
...
+    NSError *error = nil;
+
+    MSALExternalKeyPair *keyPair = [[MSALExternalKeyPair alloc] initWithPrivateKey:NULL
+                                                                         publicKey:publicKey
+                                                                             error:&error];
+
+    XCTAssertNil(keyPair);
+    XCTAssertEqual(error.code, MSALErrorInvalidExternalKeyPair);
+    XCTAssertEqual([error.userInfo[MSALExternalKeyPairFailureReasonKey] integerValue], MSALExternalKeyPairFailureReasonInvalidKeyHandle);
+    XCTAssertNotNil(error.userInfo[NSUnderlyingErrorKey]);
```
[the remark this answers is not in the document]
antonioalwan: This one is a false positive, the test passes as written.

The `underlyingError:` argument passed inside `failExternalKeyPairValidation` is indeed nil for the NULL-handle case, but that is not what ends up under `NSUnderlyingErrorKey`. The validator still builds a full error and assigns it to `*error` (`MSIDKeyOperationUtil.m:341-353`), unconditionally, whenever the caller passes an error pointer. `MSALExternalKeyPair` then stores that whole `MSIDErrorDomain` / `MSIDErrorInternal` error as the underlying error (`MSALExternalKeyPair.m:91-94`), so the key is populated.

Confirmed by running `MSALExternalKeyPairTests`: all 7 tests pass, including `testInitWithNilPrivateKey_ShouldReturnInvalidKeyHandle`.

antrix1989 (approved):

PR #3076, opened 2026-08-18
[patch] [bugfix]: Send claims in Native Auth V2 sign-in

antrix1989:
## Summary

Propagate `MSALNativeAuthSignInParameters.claimsRequest` through Native Auth V2 sign-in continuation states and include it as the `claims` form field in the final `/token` request.

The propagation covers immediate completion and resumed OTP, password, MFA, JIT registration, and resend paths while preserving compatibility for existing request-provider conformers.

## How to validate

- Start a V2 Native Auth sign-in with a claims request.
- Complete the applicable sign-in challenge.
- Verify the final `/token` request contains the serialized `claims` field.
- Verify sign-in without a claims request omits the field.

## Notes

Builds and tests were not run per request.

### DIFF
--- a/MSAL/src/native_auth/network/v2/MSALNativeAuthV2RequestProvider.swift
+++ b/MSAL/src/native_auth/network/v2/MSALNativeAuthV2RequestProvider.swift
@@ -129,6 +129,26 @@ protocol MSALNativeAuthV2RequestProviding {
                apiId: MSALNativeAuthTelemetryApiId,
                context: MSALNativeAuthRequestContext
     ) throws -> MSIDHttpRequest
+
+    /// Token exchange with an optional claims request.
+    func token(code: String,
+               scopes: [String],
+               claimsRequestJson: String?,
+               apiId: MSALNativeAuthTelemetryApiId,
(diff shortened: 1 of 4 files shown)


MSAL/src/native_auth/network/v2/MSALNativeAuthV2RequestProvider.swift
```diff
@@ -129,6 +129,26 @@ protocol MSALNativeAuthV2RequestProviding {
...
+
+extension MSALNativeAuthV2RequestProviding {
+
+    func token(code: String,
+               scopes: [String],
+               claimsRequestJson: String?,
+               apiId: MSALNativeAuthTelemetryApiId,
+               context: MSALNativeAuthRequestContext
+    ) throws -> MSIDHttpRequest {
+        return try token(code: code, scopes: scopes, apiId: apiId, context: context)
```
sidudiki, line 150: This default silently drops claims for conformers that don't implement the new overload; the current V2 test mock still uses the old overload, so this PR's claims propagation isn't exercised. Please update the mock to capture claims and add direct plus continuation/retry assertions.

PR #3079, opened 2026-08-19
Native Auth V2: server-driven (HAL) sign-in flow

spetrescu84:
## Summary

Adds the Native Auth **V2 (server-driven / HAL-based) sign-in flow**, mirroring the already-merged V2 SSPR flow. All changes are **purely additive** on top of dev's class-based architecture — no existing behavior is removed. Sign-in and password reset share a single `MSALNativeAuthFlowController` and the same HAL interaction parser.

### 1. Sign in (password-first)
1. `signIn(parameters:)` → **authorize-challenge start** (expects `401` + continuation token + `sign_in` link).
2. `signInStart` → parser returns `.challengeRequired(continuationToken, methods)`.
3. **Per-flow method validation (surfaced in the controller):** exactly **one** first-factor method of type **`password`** is required; anything else → `.generalError`.
4. Fire that method's `/challenge` → `handleSignInChallengeResult`:
   - `.verificationRequired` (password): if a password was supplied in the parameters it is **auto-submitted**; otherwise the developer receives `MSALNativeAuthPasswordRequiredState`. The password is passed as a parameter, never retained in state-machine objects (privacy).
5. `submitPassword` → `/verify` → `handleSignInSubmitPasswordResult`:
   - `.readyToComplete` → **token completion** (see below).
   - `.mfaRequired(methods)` → `MSALNativeAuthMFARequiredState` (developer must explicitly select a method — the SDK does **not** auto-send a code).
   - `.challengeRequired(methods)` → fire the first method's challenge.

### 2. MFA (second factor)
- `MSALNativeAuthMFARequiredState.selectAuthMethod` → `handleMFASelectAuthMethodResult` → fires the chosen method's `/challenge` → `MSALNativeAuthMFAVerificationRequiredState`.
- `submitChallenge` (guarded to `flowScenario == .signIn`; other scenarios → not-implemented) → `/verify` with the `otp` → `handleSignInSubmitChallengeResult` → token completion.

### 3. Password reset / SSPR (code-first, email only)
1. `resetPassword(parameters:)` → **authorize-challenge start** (expects `401` + continuation token + `reset_password` link).
2. `resetPasswordStart` → `.challengeRequired(continuationToken, methods)`.
3. **Per-flow method validation (surfaced in the controller):** exactly **one** first-factor code-based method of type **`email`** is required; anything else → `.generalError`.
4. Fire `/challenge` → `handlePasswordResetChallengeResult`:
   - `.verificationRequired` (email-only guard) → `MSALNativeAuthCodeRequiredState`.
5. `submitCode` → `/verify` → `handleSubmitCodeResult` → `.updateRequired` → `MSALNativeAuthNewPasswordRequiredState`.
6. `submitNewPassword` → poll to terminal → sign-in-after-reset token completion.
7. `resendCode` → resend `/challenge` → `handleResendCodeResult`.

### 4. Token completion (shared)
`completeSignIn`/`completeWithToken`: **authorize-challenge continue** → `authorization_code` → `POST /token` (`grant_type=authorization_code`, threading `scopes` + `claims`) → cache tokens → build `MSALNativeAuthUserAccountResult`. Token-request failures currently map to `.generalError`.

## HAL parsing model

- `MSALNativeAuthV2InteractionParsedResponse` cases: `.challengeRequired(continuationToken, methods:)`, `.mfaRequired(continuationToken, methods:)`, `.verificationRequired`, `.updateRequired`, `.pollInProgress`, `.readyToComplete`, `.browserRequired`, `.error`.
- `.challengeRequired` and `.mfaRequired` now carry **all** parsed methods (`[MSALNativeAuthV2ChallengeMethod]`); the parser no longer validates method count/type. **Per-flow method-type validation is surfaced at each controller consumption site**, so upcoming changes to the accepted method set stay explicit and local.
- `ChallengeMethodChannelType` enum (`none` / `email` / `password`) with `isEmailType` / `isPasswordType` helpers; `MSALNativeAuthV2ChallengeMethod { id, channelType, hint, challengeHref }`.
- `authenticationFactor` (`singleFactor` / `multiFactor`) is threaded through the HAL model → serializer → parser to distinguish first-factor vs MFA.

## Error mapping (grounded in captured traffic)

`flowError(from:)` requires **both** the outer `code` and the `innerError.code` to match before mapping to a specific error; it never infers a specific error from the outer code alone. Anything unrecognized (including an invalid/expired continuation token) falls through to `.generalError`.

| Outer `code` | inner / message | Mapped error |
|---|---|---|
| `invalidGrant` | `invalidOneTimeCode` (AADSTS50184) | `.invalidCode` |
| `invalidRequest` | `passwordTooWeak` (AADSTS120002) | `.invalidPassword` |
| `invalidGrant` | `invalidUserNameOrPassword` (AADSTS50126) | `.invalidCredentials` |
| `invalidRequest` | message contains `AADSTS50034` | `.userNotFound` |
| anything else | — | `.generalError` |

## Testing

- `MSALNativeAuthFlowControllerSignInTests` — sign-in incl. password auto-submit, MFA, and the single-`password`-method guard.
- `MSALNativeAuthFlowControllerTests` — SSPR incl. the single-`email`-method guard.
- `MSALNativeAuthV2ResponseParserTests` — `.challengeRequired` returning all methods + inner-code-based error mapping.

## Validation

- SwiftLint clean on `MSAL/src/native_auth`.
- iOS + macOS framework builds succeed.

## Known gaps / follow-ups

- **MFA at token exchange (`AADSTS50076` / `basic_action`)** on the `authorization_code` `/token` response is **not** yet handled — it currently maps to `.generalError`. This is intended to be handled **only** on the `authorization_code` path (not on `refresh_token` / silent refresh).

> Draft — opened for review of the V2 sign-in approach and error mapping.


MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +488,272 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+        return await completeWithToken(
+            flowContinuationState: flowContinuationState,
+            continuationToken: continuationToken,
+            scopes: flowContinuationState.scopes,
+            claimsRequestJson: flowContinuationState.claimsRequestJson,
+            step: step
+        )
+    }
+
+    // MARK: - Password Reset Result mapping
```
spetrescu84, line 743: I've moved these after SignIn to preserve order in the controller

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +488,272 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
 
     /// Maps the challenge response from the reset-password start sequence.
-    func handleChallengeResult(
+    func handlePasswordResetChallengeResult(
```
spetrescu84, line 746: updated name to clarify flow

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +488,272 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
         _ result: MSALNativeAuthV2InteractionParsedResponse,
         flowContinuationState: MSALNativeAuthFlowContinuationState,
         step: MSALNativeAuthFlowStepContext
     ) async -> MSALNativeAuthFlowControllerResponse {
         switch result {
-        case .codeRequired(let token, let verifyHref, let resendHref, let sentTo, let channelType, let codeLength):
+        case .verificationRequired(let token, let verifyHref, let resendHref, let sentTo, let channelType, let codeLength):
```
spetrescu84, line 752: This was actually incorrect, as it can be any type of verification. For SignIn it's password, for SSPR it's code

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -650,6 +1001,51 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
         return resolvedLinks
     }
 
+    // MARK: - Shared step helpers
+
+    private func performAuthorizeChallengeStart(
```
spetrescu84, line 1016: This was moved from above, it's same method as before

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -650,6 +1001,51 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+        apiId: MSALNativeAuthTelemetryApiId,
+        context: MSALNativeAuthRequestContext
+    ) async -> MSALNativeAuthV2AuthorizeChallengeParsedResponse {
+        let result: Result<MSALNativeAuthHALResponse, Error> = await send(context: context) {
+            try self.requestProvider.authorizeChallengeStart(apiId: apiId, context: context)
+        }
+        return responseParser.parseAuthorizeChallenge(context: context, result, flowScenario: flowScenario)
+    }
+
+    private func performAuthorizeChallengeContinue(
```
spetrescu84, line 1027: This was moved from above, it's same method as before

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -650,6 +1001,51 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+        apiId: MSALNativeAuthTelemetryApiId,
+        context: MSALNativeAuthRequestContext
+    ) async -> MSALNativeAuthV2AuthorizeChallengeParsedResponse {
+        let result: Result<MSALNativeAuthHALResponse, Error> = await send(context: context) {
+            try self.requestProvider.authorizeChallengeContinue(continuationToken: continuationToken, apiId: apiId, context: context)
+        }
+        return responseParser.parseAuthorizeChallenge(context: context, result, flowScenario: flowScenario)
+    }
+
+    private func performInteraction(
```
spetrescu84, line 1039: This was moved from above, it's same method as before

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -650,6 +1001,51 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+        context: MSALNativeAuthRequestContext,
+        requestBuilder: @escaping () throws -> MSIDHttpRequest
+    ) async -> MSALNativeAuthV2InteractionParsedResponse {
+        let result: Result<MSALNativeAuthHALResponse, Error> = await send(context: context, requestBuilder)
+        return responseParser.parseInteraction(context: context, result)
+    }
+
+    private func send(
```
spetrescu84, line 1047: This was moved from above, it's same method as before

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ParsedResponses.swift
```diff
@@ -47,15 +47,34 @@ enum MSALNativeAuthV2AuthorizeChallengeParsedResponse: Equatable {
...
+
 /// Parsed outcome of an SSPR interaction step (resetpassword start / challenge / verify / update / poll).
 ///
 /// A single enum represents every HAL interaction response; the parser selects the case
 /// from the HAL `state` / `action` pair.
 enum MSALNativeAuthV2InteractionParsedResponse: Equatable {
     /// `action == challenge`: a verification method is available; the SDK should auto-trigger the challenge.
     case challengeRequired(continuationToken: String, challengeHref: String, hint: String?)
-    /// `action == verify`: a one-time code is required from the user.
-    case codeRequired(continuationToken: String, verifyHref: String, resendHref: String?, sentTo: String, channelType: MSALNativeAuthChannelType, codeLength: Int)
```
spetrescu84, line 58: This was incorrect and renamed to verificationRequired as it can be code or password, it depends on method type

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -184,26 +223,22 @@ extension MSALNativeAuthV2ResponseParser {
     private func flowError(from serverError: MSALNativeAuthHALResponse.ServerError, context: MSIDRequestContext) -> MSALNativeAuthFlowError {
         let message = serverError.message
         let errorCodes = estsErrorCodes(from: message)
+        let code = serverError.code
         let innerErrorCode = serverError.innerErrorCode
         let type: MSALNativeAuthFlowError.ErrorType
 
-        if innerErrorCode == "invalidContinuationToken" {
```
spetrescu84, line 190: This was not the correct check. Flagged while looking at the JS PR. The error is captured below now in `if code == "invalidGrant" && innerErrorCode == "invalidOneTimeCode" {`
`If code == "invalidContinuationToken"` then it's actually a general error so we don't treat

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -105,13 +157,19 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
             )
         }
 
-        guard case .challengeRequired(let challengeContinuationToken, let challengeHref, _) = startResult else {
+        guard case .challengeRequired(let challengeContinuationToken, let methods) = startResult else {
             return interactionFailure(startResult, event: event, context: context, scenario: flowScenario, newState: nil)
         }
 
+        // Password reset is a code-first flow: exactly one first-factor code-based method (email) is expected.
```
spetrescu84: This was incorrect before as it wasn't checking properly. We will add in a separate pr to return the methods to the user first.

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ParsedResponses.swift
```diff
@@ -47,15 +47,49 @@ enum MSALNativeAuthV2AuthorizeChallengeParsedResponse: Equatable {
     }
 }
 
+enum ChallengeMethodChannelType: String {
```
spetrescu84: It looks like the channel challenge type is gone now and we reuse the `"type"` for both, but our top level MFA method still returns both, they will now be same, maybe we should return a new type for v2.

antrix1989 (approved):

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -162,6 +173,47 @@ final class MSALNativeAuthV2ResponseParser: MSALNativeAuthV2ResponseParsing {
...
+    private func parseChallengeMethods(
+        _ challengeResponse: MSALNativeAuthHALChallengeResponse
+    ) -> [MSALNativeAuthV2ChallengeMethod] {
+        return challengeResponse.methods.compactMap { method in
+            guard let id = method.id, let challengeHref = method.link(for: .challenge) else {
+                return nil
+            }
+            return MSALNativeAuthV2ChallengeMethod(
+                id: id,
+                channelType: ChallengeMethodChannelType(rawValue: method.type ?? "") ?? .none,
```
antrix1989: Severity: Medium – **Issue:** Any absent or unrecognized server method type is normalized to `.none`, but the method remains in the returned array as if it were structurally usable.

**Impact:** Downstream flow validation can only surface a generic error, and a newly introduced server channel is indistinguishable from a missing type. In mixed method arrays it can also affect the exact-count checks in the controller.

**Recommendation:** Fail/skip explicitly when `ChallengeMethodChannelType(rawValue:)` is nil, or preserve an `unknown(rawValue:)` representation and log the unsupported value so callers receive a diagnostic tied to the actual server response.

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ParsedResponses.swift
```diff
@@ -47,15 +48,49 @@ enum MSALNativeAuthV2AuthorizeChallengeParsedResponse: Equatable {
...
```
antrix1989: **Issue:** This new internal channel-type enum is the only V2 parser model without the `MSALNativeAuthV2` prefix, and `case none` is easy to confuse with `Optional.none` in optional contexts.

**Impact:** The inconsistent name makes this type harder to distinguish from the public `MSALNativeAuthChannelType`, while `.none` creates an avoidable Swift ambiguity as the model evolves.

**Recommendation:** Rename it to something like `MSALNativeAuthV2ChallengeMethodChannelType` and use `unknown` or `unspecified` for the fallback case.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +500,274 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+    ) async -> MSALNativeAuthFlowControllerResponse {
+        switch result {
+        case .readyToComplete(let token):
+            return await completeSignIn(flowContinuationState: flowContinuationState, continuationToken: token, step: step)
+        case .mfaRequired(let token, let methods):
+            let next = makeMFAContinuation(from: flowContinuationState, continuationToken: token, methods: methods)
+            return mfaRequiredResponse(flowContinuationState: next, methods: methods, step: step)
+        case .challengeRequired(let token, let methods):
+            guard let method = methods.first else {
+                return interactionFailure(result, event: step.event, context: step.context, scenario: flowContinuationState.flowScenario, newState: nil)
```
antrix1989: **Issue:** The changed native-auth sources do not lint clean with the repository's configured SwiftLint rules: this line exceeds 150 characters, line 575 also exceeds 150 characters and lacks a space after `//`, and `MSALNativeAuthV2ParsedResponses.swift:85` has trailing whitespace.

**Impact:** These additions introduce avoidable lint warnings in a directory that the repository guidance requires to be warning-free.

**Recommendation:** Wrap the long call and comment, add the comment space, and remove the trailing whitespace.

antrix1989 (commented):
**Additional telemetry concern**

**Issue:** `submitCode` and `resendCode` always use the reset-password telemetry IDs after the sign-in-specific branches were removed from the shared flow controller.

**Impact:** If either shared continuation is reached with a sign-in state, the operation is attributed to password reset, which makes sign-in funnel and failure telemetry inaccurate.

**Recommendation:** Either guard these methods to `.passwordReset` (as the MFA continuations already guard to `.signIn`) or select the telemetry API ID from `flowScenario`.

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -162,6 +173,47 @@ final class MSALNativeAuthV2ResponseParser: MSALNativeAuthV2ResponseParsing {
...
+                errorDescription: "Invalid interaction response: challenge action did not specify challengeContext"
+            ))
+        }
+    }
+
+    private func parseChallengeMethods(
+        _ challengeResponse: MSALNativeAuthHALChallengeResponse
+    ) -> [MSALNativeAuthV2ChallengeMethod] {
+        return challengeResponse.methods.compactMap { method in
+            guard let id = method.id, let challengeHref = method.link(for: .challenge) else {
```
sidudiki: The previous parser could fall back to the top-level `_links.challenge`; this now drops every method without an `id` or per-method challenge link, and an empty result becomes `generalError`. `id` is only needed for MFA method selection, so this can regress single-factor/reset-password responses that worked before. Please retain the top-level link fallback and require `id` only for the multi-factor path.

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -184,26 +236,22 @@ extension MSALNativeAuthV2ResponseParser {
...
+            // Wrong one-time code (AADSTS50184); the app can prompt the user for a new code.
+            type = .invalidCode
+        } else if code == "invalidRequest" && innerErrorCode == "passwordTooWeak" {
+            // New password fails complexity requirements (AADSTS120002).
             type = .invalidPassword
-        } else if innerErrorCode == "invalidUserNameOrPassword"
-                    || errorCodes.contains(MSALNativeAuthESTSApiErrorCodes.invalidCredentials.rawValue) {
-            // Wrong username/password at sign in (AADSTS50126): a recoverable credentials error,
-            // not an invalid one-time code.
+        } else if code == "invalidGrant" && innerErrorCode == "invalidUserNameOrPassword" {
```
sidudiki, line 268: This exact `code` + `innerErrorCode` mapping removes the existing ESTS-code fallbacks. A wrong password/code response with a different outer/suberror shape now becomes `generalError`; the controller only supplies a retry state for `invalidPassword`/`invalidCode`, so the user is forced to restart the flow. Please keep the exact matches first but retain the AADSTS/error-code fallbacks as a recovery safety net.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +500,274 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+                links: [.verify: verifyHref, .resend: resendHref]
+            )
+            // Sign in with password is a password first flow: the server must select the password method.
+            guard channelType.isPasswordType else {
+                let error = MSALNativeAuthFlowError(type: .generalError, errorDescription: MSALNativeAuthErrorMessage.generalError)
+                stopTelemetryEvent(step.event, context: step.context, error: error)
+                return response(.error(error: error, newState: nil), context: step.context, scenario: flowContinuationState.flowScenario)
+            }
+            if let password = password, !password.isEmpty {
+                stopTelemetryEvent(step.event, context: step.context)
```
sidudiki: This closes `SignInWithPasswordStart` as successful before password submission begins, then `submitPassword` starts a second event. A failed password sign-in therefore still records the start as success and emits two events for one public API call, skewing the launch funnel. Can we carry the existing step/event through the internal submit and stop it with the final outcome?

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +500,274 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+                ? MSALNativeAuthFlowError(
+                    type: .invalidPassword,
+                    errorDescription: error.errorDescription,
+                    errorCodes: error.errorCodes,
+                    correlationId: error.correlationId,
+                    errorUri: error.errorUri)
+                : error
+            stopTelemetryEvent(step.event, context: step.context, error: surfacedError)
+            return response(
+                .error(error: surfacedError, newState: surfacedError.type == .invalidPassword ? recoverableState : nil),
```
sidudiki: This creates a recoverable state for an invalid password, but `MSALNativeAuthFlowResponseDispatcher.dispatch` matches `.error(let error, _)` and discards it before calling `onFlowError`. The app therefore has no state with which to resubmit, so a mistyped password is terminal even though the controller tests assert `newState` exists. Please surface the state through the delegate contract (or re-dispatch the action-required state) and add a dispatcher-level retry test.

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -162,6 +173,47 @@ final class MSALNativeAuthV2ResponseParser: MSALNativeAuthV2ResponseParsing {
             return .error(MSALNativeAuthFlowError(type: .generalError, errorDescription: "Unexpected interaction response"))
         }
     }
+
+    private func parseChallengeResponse(
+        _ challengeResponse: MSALNativeAuthHALChallengeResponse,
+        continuationToken: String,
+        context: MSIDRequestContext
+    ) -> MSALNativeAuthV2InteractionParsedResponse {
+        if challengeResponse.authenticationFactor == "multiFactor" {
```
sidudiki, line 182: This makes `challengeContext.authenticationFactor` mandatory for every V2 challenge response. A missing or unrecognized value now becomes `generalError`, whereas the previous parser accepted the challenge — including for the already-shipped reset-password flow. Unless the service contract guarantees this field everywhere, can we default missing/unknown values to single-factor (or scope the strict check to sign-in) and add present/absent/unknown serializer tests?

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +500,274 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+            let next = makeSignInContinuation(from: flowContinuationState, continuationToken: token, links: [:])
+            let challengeResult = await performInteraction(context: step.context) {
+                try self.requestProvider.challenge(
+                    href: method.challengeHref,
+                    continuationToken: token,
+                    apiId: step.apiId,
+                    context: step.context
+                )
+            }
+            return await handleSignInChallengeResult(challengeResult, flowContinuationState: next, step: step)
```
sidudiki: This routes every post-password `challengeRequired` response back through `handleSignInChallengeResult`, whose verification branch accepts only password. A server-driven email/SMS follow-up therefore terminates with `generalError`, and this branch has no test. Please route code-based methods to a code-required state (or explicitly reject unsupported follow-ups here with a diagnostic) and cover the transition.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -77,7 +77,59 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+                context: context
+            )
+        }
+
+        guard case .challengeRequired(let challengeContinuationToken, let methods) = startResult else {
+            return interactionFailure(startResult, event: event, context: context, scenario: flowScenario, newState: nil)
+        }
+
+        // Sign in with password is a password-first flow: exactly one first-factor method of type `password` is expected.
+        guard methods.count == 1, let method = methods.first, method.channelType.isPasswordType else {
```
sidudiki: The parser explicitly supports multiple single-factor methods, but this guard rejects the whole response whenever a tenant offers more than one first factor — even if password is present. That makes password + email-OTP tenants fail with an opaque `generalError`. Can we select `methods.first(where: { $0.channelType.isPasswordType })` and fail only when no password method exists?

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Good catch. I've removed the `recoverableState` as the developer should use the previous state they had as the server doesn't return a new continuation token so the previous one has to be reused.
@antrix1989 to confirm, should we add a new state here?
`@MainActor func onFlowError(error: MSALNativeAuthFlowError, scenario: MSALNativeAuthFlowScenario, newState: MSALNativeAuthState?)` The developer would need to downcast and it would be the same as the previous one that they already have.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: I've updated the check because I agree there could be more returned but this doesn't depend on the tenant.
Each client Id is associated in EEID with a User Flow which is either Email + OTP or Email + Password, can never be both. So both cannot be returned for the same client id.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Good point. Added an internal method because it's being called from 2 spots. When the user starts signIn with username + password and when user submits password (after original signin only with email but server requires password)

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: Done

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
spetrescu84: Actually id is required for single factor also, specifically for SSPR, it could be user can select either Email or SMS or other methods to reset their password so this is boilerplate for that. We will be changing this in a future PR to return the methods to the user for this case also so they can select which one they want to use

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ParsedResponses.swift
spetrescu84: Renamed to MSALNativeAuthV2ChallengeMethodChannelType.
Removed none and now there's a guard in place to make sure it decodes

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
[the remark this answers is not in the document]
spetrescu84: this was changed and also it's by default lowercase

spetrescu84: Added this to the guard `let channelType = MSALNativeAuthV2ChallengeMethodChannelType(rawValue: method.type ?? ""),` so it has to be one of the 2 values now.

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
spetrescu84: We are relying on the `code` and `innnerErrorCode` because AADSTS error code might change at any point, while the others we expect never to change or in next V of the API

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
spetrescu84: We're expecting this always comes back from the API, if not it's a general error as we rely on this to decide the state and don't want to default to singleFactor as user might have already completed one

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
spetrescu84: We should not actually be treating `case .challengeRequired(let token, let methods):`, this was a mistake.
After validating the password, there can never be another first factor validation as the password was the first factor. It's either multi factor or general error.

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -77,7 +77,59 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+                context: context
+            )
+        }
+
+        guard case .challengeRequired(let challengeContinuationToken, let methods) = startResult else {
+            return interactionFailure(startResult, event: event, context: context, scenario: flowScenario, newState: nil)
+        }
+
+        // Sign in with password is a password-first flow: select the password method from the offered first-factor methods.
+        guard let method = methods.first(where: { $0.channelType.isPasswordType }) else {
```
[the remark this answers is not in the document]
spetrescu84: We are ok to select the first one of type password in case there are more

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -347,62 +507,257 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
...
+                    correlationId: error.correlationId,
+                    errorUri: error.errorUri)
+                : error
+            stopTelemetryEvent(step.event, context: step.context, error: surfacedError)
+            return response(
+                .error(error: surfacedError),
+                context: step.context, scenario: flowContinuationState.flowScenario
+            )
+        default:
+            return interactionFailure(result, event: step.event, context: step.context, scenario: flowContinuationState.flowScenario, newState: nil)
```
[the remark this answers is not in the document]
spetrescu84: `.challengeRequired`  is generated only for  authenticationFactor == singleFactor, which the server emits solely at  `signin/start`, by this point it was already validated the single factor so it should fail

MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift
```diff
@@ -424,15 +779,15 @@ final class MSALNativeAuthFlowController: MSALNativeAuthBaseController, MSALNati
         step: MSALNativeAuthFlowStepContext
     ) -> MSALNativeAuthFlowControllerResponse {
         switch result {
-        case .codeRequired(let token, let verifyHref, let resendHref, let sentTo, let channelType, let codeLength):
+        case .verificationRequired(let token, let verifyHref, let resendHref, let sentTo, let channelType, let codeLength):
             let next = makeContinuation(from: flowContinuationState, continuationToken: token, links: [.verify: verifyHref, .resend: resendHref])
             return codeRequiredResponse(flowContinuationState: next, sentTo: sentTo, channelType: channelType, codeLength: codeLength, step: step)
```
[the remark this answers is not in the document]
spetrescu84: Done

MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift
```diff
@@ -162,6 +173,49 @@ final class MSALNativeAuthV2ResponseParser: MSALNativeAuthV2ResponseParsing {
...
+    }
+
+    private func parseChallengeMethods(
+        _ challengeResponse: MSALNativeAuthHALChallengeResponse
+    ) -> [MSALNativeAuthV2ChallengeMethod] {
+        return challengeResponse.methods.compactMap { method in
+            guard let id = method.id,
+                    let channelType = MSALNativeAuthV2ChallengeMethodChannelType(rawValue: method.type ?? ""),
+                    let challengeHref = method.link(for: .challenge) else {
+                return nil
```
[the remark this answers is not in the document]
spetrescu84: Done

MSAL/test/unit/native_auth/network/v2/MSALNativeAuthV2ResponseParserTests.swift
```diff
@@ -268,18 +318,22 @@ final class MSALNativeAuthV2ResponseParserTests: XCTestCase {
         XCTAssertEqual(result, .error(MSALNativeAuthFlowError(type: .userNotFound)))
     }
 
-    func test_parseInteraction_invalidGrant_mapsToInvalidCode() {
+    func test_parseInteraction_invalidGrantWithoutInnerCode_mapsToGeneralError() {
         let serverError = MSALNativeAuthHALResponse.ServerError(code: "invalidGrant", message: "wrong code", innerErrorCode: nil, correlationId: nil)
         let response = makeResponse(error: serverError)
         let result = sut.parseInteraction(context: context, .success(response))
-        XCTAssertEqual(result, .error(MSALNativeAuthFlowError(type: .invalidCode)))
+        XCTAssertEqual(result, .error(MSALNativeAuthFlowError(type: .generalError)))
```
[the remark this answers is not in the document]
spetrescu84: Added

antrix1989 (approved):

Issue #3082, opened 2026-08-20
AcquireTokenSilent fails with AADSTS500210 for CIAM guest accounts — cached account is labelled with the home tenant, not the token's realm


daniyalalikausar52:
### Copilot Usage Confirmation

- [x] I have tried using GitHub Copilot to resolve this issue.

### MSAL Version

2.14.1

### Description

Description

acquireTokenSilent never succeeds for a guest account in a CIAM (External ID) tenant — i.e. a user whose home tenant is a different Entra tenant. Interactive sign-in works fine and returns a valid token; only silent renewal fails. Because the access token lives ~65 minutes, affected users are forced to sign in again roughly every hour.

Environment: MSAL for iOS 2.14.1 (SPM binary target), iOS 26.4, real device. Authority is MSALCIAMAuthority with a vanity domain: https://<subdomain>.ciamlogin.com/<ciam-tenant-id>.

The cached account is the problem: its only tenant profile carries the CIAM tenant's object id but is labelled with the home tenant id. MSAL then derives the refresh authority from that home tenant while keeping the CIAM host, and the server rejects the pairing.

Workarounds attempted, none of which work:

Setting MSALSilentTokenParameters.authority to the CIAM authority — MSAL substitutes the account's home tenant regardless.
accountForIdentifier: with <oid>.<ciam-tenant-id> (using either the home or CIAM object id) — no match.
accountsForParameters: with tenantProfileIdentifier set to the CIAM object id — returns the same home-tenant-labelled account.

There is no cached record for the CIAM tenant to target, and no public API to override the realm.

Cross-platform note: MSAL-Android has the same underlying defect ([AzureAD/microsoft-authentication-library-for-android#2172](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/2172), #2073). It surfaces earlier there — the local cache lookup misses and fails before any network call — and can be worked around by wrapping the account so getTenantId() returns the realm while getHomeAccountId() stays real. That is not portable to iOS: MSALAccount has no separate realm field, so correcting the tenant also changes the key used to locate the refresh token.

### Error Details

domain      = MSALErrorDomain
code        = -50002               (MSALErrorInteractionRequired)
oauthError  = server_error
description = "User interaction is required"
underlying  = MSALErrorDomain / -50000   (MSALErrorInternal)

Two things worth noting:

oauthError is server_error, not invalid_grant — the refresh token itself is valid; the request is malformed.
[...]

PR #3097, opened 2026-09-03
[patch] [bugfix]: Restore AI issue auto-reply by moving from GitHub Models to Copilot CLI

fidelianawar:
## Problem

The AI issue auto-reply workflow has been silently dead since **2026-07-30**, when [GitHub Models was retired](https://docs.github.com/en/copilot/how-tos/copilot-cli/use-copilot-cli-in-actions).

Every model in `MODELS_MODEL` now returns `HTTP 410 github_models_retirement_brownout`. I confirmed this is still true today:

```
POST https://models.github.ai/inference/chat/completions
→ 410 {"code":"github_models_retirement_brownout"}
```

The failure is invisible: the generate step warns, emits empty text, and exits 0, so `Post reply` skips on its `steps.ai.outputs.text != ''` guard and **the run goes green while nobody gets an answer.**

Concrete example — #3082 was filed on 2026-08-20 and has never received a reply. The workflow *did* run (`32371193675`, `32371193495`); the gate passed and the model call was attempted:

```
← Models API responded HTTP 410 (model=openai/gpt-5, attempt 1/3)
← Models API responded HTTP 410 (model=openai/gpt-4.1, attempt 1/3)
← Models API responded HTTP 410 (model=openai/gpt-4o, attempt 1/3)
← Models API responded HTTP 410 (model=openai/gpt-4o-mini, attempt 1/3)
##[warning]No model in the preference list returned 2xx.
```

## Fix

Move to **Copilot CLI**, the supported way to run AI in Actions. It authenticates with the built-in `GITHUB_TOKEN` — no PAT, no API key. That mattered here: this repo has **zero** Actions secrets, so any provider requiring a key (Azure AI Foundry, direct OpenAI) would have needed an admin to provision one first.

- `permissions`: swap the retired `models: read` for `copilot-requests: write`.
- Install `@github/copilot`; invoke with `--yolo` (required for non-interactive runs), plus `--no-ask-user` and `--disable-builtin-mcps` to keep the run autonomous and hermetic.
- **Pass the prompt on disk, not through `-p`.** The assembled prompt is ~90 KB of Objective-C, Markdown and shell metacharacters (backticks, `${...}`, quotes); passing that as an argv string is a quoting minefield. Copilot reads the file and writes the finished comment to `ai_reply.md`, so neither prompt nor reply transits the shell.
- **Randomise the `GITHUB_OUTPUT` heredoc delimiter.** The reply is model-generated, so a literal `EOF` line in a comment would have closed the heredoc early and truncated the posted comment.
- Add `timeout-minutes: 15` so a runaway agent can't hang the job.

The existing failure contract is preserved exactly — warn, emit empty text, exit 0, let `Post reply` skip. A flaky model call still must not fail the run or page anyone.

## Also: reference budget

The 8 KB reference cap existed *only* because GitHub Models capped input at 8000 tokens/request. It was truncating the curated grounding set from 46 KB down to 8 KB — on the #3082 run: `Reference material truncated: was 46358B, now 8000B`.

Copilot CLI has no comparable per-request cap, so the manifest is now sent whole. The caps remain as a runaway guard for forks that override them downward.

## Validation

- `yaml.safe_load` parses; step graph is `gate → Install Copilot CLI → ai → Post reply`.
- `bash -n` clean on the extracted generate step.
- Ran the prompt-assembly path locally against the new budgets:

| | before | after |
|---|---|---|
| reference material | 8,000 B (**truncated**) | 84,932 B (**untruncated**) |
| final prompt | — | 91,190 B |

  The assembled prompt now contains `MSALCIAMAuthority` and `MSALSilentTokenParameters` — both of which the old 8 KB cap cut off, and both directly relevant to #3082.
- Confirmed no stale references to `MODELS_MODEL` / `models.github.ai` / `GH_MODELS_MODEL` remain.

## Follow-up

Once this merges, #3082 can be triaged by adding the `target-ai-triage` label.

Worth noting for whoever picks up #3082: the reporter's analysis looks correct, and it appears to be a genuine library bug rather than misconfiguration — the CIAM factory builds its cache authority from `client_info.utid` (the **home** tenant) where the AAD factory uses the id_token realm. That realm flows into the account's tenant profile, which is why the profile carries the CIAM `oid` but the home `tid`. So `needs-review` looks more appropriate than a "send us your config" reply.

### DIFF
--- a/.github/ai-prompts/reference-files.txt
+++ b/.github/ai-prompts/reference-files.txt
@@ -2,15 +2,15 @@
 # One path per line, relative to the repo root. Blank lines and lines
 # starting with # are ignored. Missing files are skipped with a warning.
 #
-# BUDGET: GitHub Models caps some free-tier models (gpt-4o-mini) at
-# 8000 tokens per request. With system prompt + question + completion
-# overhead, reference content must fit in ~8 KB (~2000 tokens).
+# BUDGET: the workflow runs Copilot CLI, which has no small per-request
+# input cap, so ai-issues-auto-reply.yml raises the caps via the
+# *_OVERRIDE env vars and this manifest (~107 KB) is sent whole.
 #
 # Each .h file is stripped of its MIT license block and #import lines
 # before inlining. A per-file cap (MAX_PER_FILE_BYTES, default 3000)
-# and a total cap (MAX_REFERENCE_BYTES, default 8000) apply. Both can be
-# raised on plans with higher per-request limits via the corresponding
-# *_OVERRIDE env vars in the workflow.
+# and a total cap (MAX_REFERENCE_BYTES, default 8000) still apply as a
+# runaway guard. Both are raised via the corresponding *_OVERRIDE env
+# vars in the workflow.
 #
 # Files are included in the order listed below. Anything that doesn't
 # fit under the total cap is truncated from the tail, so the most
--- a/.github/workflows/ai-issues-auto-reply.yml
+++ b/.github/workflows/ai-issues-auto-reply.yml
@@ -25,7 +25,9 @@ on:
 permissions:
   issues: write
   contents: read
-  models: read
+  # Lets Copilot CLI authenticate with the built-in GITHUB_TOKEN instead of a
+  # PAT. Replaces the retired `models: read` scope.
+  copilot-requests: write
 
 env:
   # ---- Behavior toggles ----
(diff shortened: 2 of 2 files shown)


fidelianawar:
Auto-closed by GitHub when the head branch was renamed to `fidelianawar-fix-github-issues-actions-bot`. Continued in #3098 — same commit, no content changes.

